PR #2, opened 2015-02-27
Add search page

tombye:
Addition of a /search page, as part of the spike to add G6 filters.

The filters for this page were meant to be built on top of a static data source but this wasn't available so we just have the page and a sample data source to be built on top of later.


TheDoubleK:
npm install is dying in Travis...

PR #4, opened 2015-02-27
Update to run on port 5002

TheDoubleK:
For development we will want to run multiple apps, so they should each bind to a different port number.


tombye:
:+1:

PR #5, opened 2015-03-02
Use NPM for frontend builds

tombye:
Makes NPM the one tool for all front-end build tasks.

Some of this was previously done using python scripts in `./scripts` so these are removed.


README.md
```diff
@@ -92,29 +92,13 @@ To install the required Node modules, type:
...
-```
-node_modules/gulp/bin/gulp.js build:development
-```
-
-#### Compiling for production 
-
-```
-node_modules/gulp/bin/gulp.js build:production
-```
+Note: `npm run frontend-install` is run as a post-install task after you run `npm install`.
```
TheDoubleK: This line could be clearer - does it mean
1. `npm run frontend-install` is run **automatically** as a post-install task **when** you run `npm install`, or
2. `npm run frontend-install` **should be** run manually after you run `npm install`?

tombye: Number 1. I'll change it

TheDoubleK:
Run locally and all seems to work :+1:

PR #6, opened 2015-03-02
Use search api

TheDoubleK:
Alters the buyer app to hit the search API instead of using elasticsearch directly.

More red than green here - my favourite kind of pull request :)


app/models.py
```diff
@@ -12,110 +28,44 @@ def strip_services_wrapper(content):
...
-                        }
-                    }
-                    )
-    return res
+def search_with_filters(query, filters, start_from=0, size=10):
+    filter_string = ""
+    for k, v in filters.iteritems():
+        filter_string = filter_string + "&" + k + "=" + v.lower()
+    url = search_url + "/search?q=" + query + filter_string
+    response = requests.get(
```
robyoung: Query parameters can be passed directly to requests. This is safer than joining them ourselves because things like query escaping is done for us. (same point for line 43 as well)

run_buyer_frontend.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
robyoung: :+1: on having a simple `run_something.sh` script that adds sensible defaults. I wonder if we should call it the same in every app (something like `run_app.sh`). It would mean that we could more easily write tools to automate startup for development.

TheDoubleK: Hmm, yeah I was deliberately making them with different names (renaming from the original ``run.sh``), but it might be better to have the same in each app.  I'll sort it now.

app/models.py
robyoung: Bah, sorry, I forgot to add the link http://docs.python-requests.org/en/latest/user/quickstart/?highlight=query#passing-parameters-in-urls

TheDoubleK:
Should be ready to go now I think.

PR #7, opened 2015-03-05
Use search api

TheDoubleK:


app/models.py
```diff
@@ -12,110 +28,45 @@ def strip_services_wrapper(content):
...
-                                },
-                                "filter": {
-                                    "bool": {"must": filters}
-                                }
-                            }
-                        }
-                    }
-                    )
-    return res
+def search_with_filters(query, filters):
```
quis: You could probably get rid of the `search_without_query` method if you set a default value for the `filter` argument on this method (eg `filters=[]`)

TheDoubleK:
Quite right @quis, sorted now to be much neater.

README.md
```diff
@@ -40,35 +35,28 @@ Install new Python dependencies with pip
...
+### Run the development server
 
-### Run the tests
+To run the Buyer Frontend App for local development you can use the convenient run 
+script, which sets the required environment variables to defaults if they have
+not already been set:
 
 ```
-./scripts/run_tests.sh
+./run_app.sh
```
allait: `./scripts/run_app.sh`?

app/models.py
```diff
@@ -1,9 +1,25 @@
...
+    raise Exception("DM_BUYER_FRONTEND_API_AUTH_TOKEN token is not set")
+if api_url is None:
+    print('API URL must be supplied in DM_API_URL')
+    raise Exception("DM_API_URL is not set")
+if search_access_token is None:
+    print('Token must be supplied in DM_BUYER_FRONTEND_SEARCH_API_AUTH_TOKEN')
+    raise Exception("DM_BUYER_FRONTEND_SEARCH_API_AUTH_TOKEN token is not set")
+if search_url is None:
+    print('Search API URL must be supplied in DM_SEARCH_API_URL')
+    raise Exception("DM_SEARCH_API_URL is not set")
```
allait: I'm not sure if it's worth doing later rather than right now, but instead of looking up the variables here we could add them to the app config. `app.create_app` already checks the environment for variables matching our config attributes, we could extend it to raise an exception when the value is `None` in the config and the matching environment variable doesn't exist.

We'll then have to pass the config or the required values to the `search_for_services`, but this will make it easier to test.

TheDoubleK: This is a good plan, I wasn't sure of the best place to do this so put it here.  We'll need to do the same for all frontend apps (which currently check in the same way as this).  Let's do that next in a separate set of pull requests.

README.md
TheDoubleK: Yep.

PR #11, opened 2015-03-10
Add authentication exceptions for API access

tombye:
Requests to the APIs currently return a 404 (via a `KeyError` being raised) if the response does not contain service data.

It would be useful if we could log some of the information about the response for debugging. This attempts to set a good starting point by adding an exception for authentication errors.


quis:
@tombye Fixed your Travis tests for you. LGTM now but @TheDoubleK should probably have a look because he's been closer to the development of this app.

tombye:
Thanks @quis!

TheDoubleK:
The logged error isn't really that helpful, as it only points to the line in the exceptions class, with no further stack trace: 
```--------------------------------------------------------------------------------
ERROR in exceptions [/Users/kevinkeenoy/gdsworkspace/digitalmarketplace-buyer-frontend/app/exceptions.py:14]:
An authentication error occurred
--------------------------------------------------------------------------------```

But certainly enough to go on for now.

PR #13, opened 2015-03-19
Add service page

tombye:
service-page-desktop

It also includes some custom styling to help the service attributes work better on smaller screens:

service-page-mobile-table-styling

This takes the formatting logic from the Grails app and puts it into presenters instead, meaning we can put the expectations for that logic into tests.

Also brings in the breadcrumb SCSS from the Digital Marketplace front-end toolkit.

Story URL: https://www.pivotaltracker.com/story/show/89291246


quis:
Heroic effort. Going to take a proper look at this later. One thing I notice is the table at the bottom of the screenshot is 2/3rds width—should be full width.

tombye:
Ah, good point. Once I've fixed the builds I'll make that amendment.

app/helpers/service_helpers.py
```diff
@@ -0,0 +1,4 @@
+def get_lot_name_from_acronym(blueprint, acronym):
+    """Return the full lot name for its acronym"""
+
+    return blueprint.config['LOTS'][acronym]
```
robyoung: This could throw a `KeyError` which isn't being handled at the call site. What's the intended behaviour if the acronym isn't found?

quis:
:+1:

app/helpers/service_helpers.py
tombye: The block in which it's called is in a `try` statement https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/add-service-page/app/main/views.py#L22, set up to catch KeyErrors.

robyoung: Completely missed that sorry.

app/presenters/service_presenters.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def format(self, value):
+        """Formats the value parameter based on its type"""
+        value_format = self.get_data_type(value)
+        if value_format is 'boolean':
+            if value:
+                return u'Yes'
+            else:
+                return u'No'
+        elif (value_format is 'list') and (value[0] == 'None'):
```
TheDoubleK: This line throws an error for the case of empty lists - e.g. `/service/5390159512076288`, `/service/6312248523882496`, `/service/6110806571745280`

TheDoubleK:
I'm not sure that this necessarily needs to be addressed in this pull-request as it's maybe a wider design question, but I'm not sure that using bulleted lists for "checkbox" questions looks quite right - see the screenshot here - as a user I'm going to think "Why do some UK's have bullet points and some don't?"
Maybe a list without bullets would be better - and then where suppliers had only picked one item from a list the answers would look the same as for other questions.

screen shot 2015-03-20 at 16 39 33

quis:
@TheDoubleK In the Grails app I didn't show a bullet if there was only one option—see: https://www.digitalmarketplace.service.gov.uk/service/5997878627532800

Also, are we missing the assurance part of these responses?

tombye:
Yeah, I think we can combine both these issues with one rule: Single item lists should display as a string. I mean instead of assuming the API will return a `["None"]` which is what were are doing now.

tombye:
Closing as working on the assurance part Chris highlighted

tombye:
OK, I've added support for assurances, plus a few fixes for the following I noticed in the meantime:

- styling issues on page
- lack of back-to-top link
- fix for a problem with attributes whose value is generated by a function

tombye:
There's another problem with how this code is handling the pricing caveats. Closing until repaired.

tombye:
The last two commits (aa1bf7e 77b730d) cover some problems with the presenter code making assumptions about what attributes should be available in the data that comes back from the content API.

I've checked all the attributes the presenter uses against the [schemas for each lot](https://github.com/alphagov/digitalmarketplace-api/tree/master/json_schemas) so all that are not listed as 'required' are tested before being used.

quis:
Looks excellent with a variety of services now. One nitpick: the service title shouldn't be more than 2/3 width.

PR #14, opened 2015-04-07
Elastic Beanstalk deployment

allait:
Adds .ebextensions packages config, build script and updates gulp-sass

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,2 @@
 [pep8]
-exclude = ./venv,./bower_components,./digital_marketplace_frontend_toolkit
+exclude = ./venv,./bower_components,./node_modules
--- a/.ebextensions/packages.config
+++ b/.ebextensions/packages.config
@@ -1,0 +1,3 @@
+packages:
+  yum:
+    libffi-devel: []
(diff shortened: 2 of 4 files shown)


setup.cfg
```diff
@@ -1,2 +1,2 @@
 [pep8]
-exclude = ./venv,./bower_components,./digital_marketplace_frontend_toolkit
```
robyoung: Why does this also remove `./digital_marketplace_frontend_toolkit`? It doesn't mention it in the commit comment.

TheDoubleK:
I think that the `digital_marketplace_frontend_toolkit` folder is no longer imported here at all.  It used to be brought in as part of the frontend build but is done in a different way now so we no longer need this exclusion from pep8.

quis:
For reference, it's imported here now: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/bower.json#L7

setup.cfg
allait: I've used the same file we have in other frontend apps. I couldn't find `digital_marketplace_frontend_toolkit` after running frontend build tasks, but as @quis mentioned it seems to be in `./bower_components/digital_marketplace_frontend_toolkit` now, which is ignored by `./bower_components`

PR #16, opened 2015-04-13
Add latest search patterns from toolkit

quis:
This commit bumps the version of the Digital Marketplace frontend toolkit, which updates the following patterns:
- Search result (now looks more like a GOV.UK finder)
- Multi-select (now renamed to 'option select' and using a more recent version of the GOV.UK component)

image


robyoung:
If I uncheck boxes and click filter again they remain checked.

quis:
@robyoung Is that something that this PR has broken, or something that's broken already?

robyoung:
I don't know, this is the first time I've started up this app. Do you want me to raise that as a separate bug?

quis:
I don't really know the history of the app… It might well be something that
@tombye is fixing with his forthcoming mega pull request.

quis:
Rebased on top of @tombye's branch, for review as one unit.

PR #17, opened 2015-04-14
Replace hard-coded filter content with YAML

quis:
This does not change which filters we show. It just moves the _definition_ of which filters we show:
- from being hard-coded in the app
- to a YAML manifest which the content loader helper can use to pull in the labels from the content repo

This means changing the keys used in the template to those used by the content repo.

This also fixes the bug that @robyoung noticed where filter state was persisting when it shouldn't.

_N.B. This pull request is against @tombye's branch, not master_

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,7 @@
 from flask import Flask
 from flask.ext.bootstrap import Bootstrap
 from config import config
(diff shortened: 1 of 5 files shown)


quis:
Closing until tests pass.

PR #20, opened 2015-04-16
Add homepage

quis:
This replaces "Hello world!"

It copies the current Digital Marketplace, with the following caveats:

1. No link to Crown Hosting—we don't have the page yet
2. No link to framework information pages (we don't have the pages yet, and maybe they will be published on GOV.UK by the time this app goes live)

And the following sub-caveat:

1. I don't know what the URL structure for search will be yet, so have assumed `/search?lot=saas`

### DIFF
--- a/app/assets/scss/_index-page.scss
+++ b/app/assets/scss/_index-page.scss
@@ -1,0 +1,32 @@
+// styles specific to index pages
+
+.index-page {
+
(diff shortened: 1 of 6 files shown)


TheDoubleK:
The lack of all punctuation disturbs me, but "it's in the guide", I know.

PR #28, opened 2015-04-30
`/_status` endpoint includes information about dependent APIs

pcraig3:
`/_status` endpoint knows about the (data) api and the search api.  

If all is well, response looks like this:
```json
{
  "api_status": {
    "db_version": "10_adding_supplier_to_user_table",
    "status": "ok",
    "version": null
  },
  "search_api_status": {
    "app_version": null,
    "db_status": {
      "message": [
        {
          "num_docs": 16423,
          "primary_size": "26mb"
        }
      ],
      "status_code": 200
    },
    "status": "ok"
  },
  "status": "ok",
  "version": null
}
```
If one of the APIs returns a 500-level response (maybe it's still responding, but no longer has a connection to its database), the status response looks more like this:
```json
{
  "api_status": {
    "message": "Error connecting to database",
    "status": "error",
    "version": null
  },
  "message": "Error connecting to the (Data) API.",
  "search_api_status": {
    "app_version": null,
    "db_status": {
      "message": [
        {
          "num_docs": 16423,
          "primary_size": "26mb"
        }
      ],
      "status_code": 200
    },
    "status": "ok"
  },
  "status": "error",
  "version": null
}
```
If our APIs are down:
```json
{
  "api_status": null,
  "message": "Error connecting to the (Data) API and the Search API.",
  "search_api_status": null,
  "status": "error",
  "version": null
}
```

#### Note : 

I'm referring to it as the (Data) API in the response message text in the case of an error, but haven't used the `data_` prefix anywhere else.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -34,19 +34,6 @@ def strip_services_wrapper(content):
     return content_json["services"]
 
 
-def get_api_status():
-    return requests.get(
-        api_url + '/_status',
-    )
-
-
-def get_search_api_status():
-    # Might be the wrong url - os.getenv('DM_SEARCH_API_URL') + "/search"
-    return requests.get(
-        search_api_url + '/_status',
-    )
-
-
 def get_service(service_id):
     url = api_url + "/services/" + service_id
     response = requests.get(
--- a/app/models.py
+++ b/app/models.py
@@ -34,7 +34,7 @@ def strip_services_wrapper(content):
     return content_json["services"]
 
 
-def get_data_api_status():
+def get_api_status():
     return requests.get(
         api_url + '/_status',
     )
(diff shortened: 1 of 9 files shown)


app/status/views.py
```diff
@@ -1,9 +1,50 @@
...
+
+    api_response = utils.return_response_from_api_status_call(
+        models.get_api_status
+    )
+
+    search_api_response = utils.return_response_from_api_status_call(
+        models.get_search_api_status
+    )
+
+    apis_wot_got_errors = []
```
TheDoubleK: Same again on the variable name I'm afraid. I'm not against frivolity, honest...

app/status/views.py
```diff
@@ -1,9 +1,50 @@
...
+
+    if api_response is None or api_response.status_code is not 200:
```
TheDoubleK: And same again for integer comparison - should always be == or !=

app/models.py
```diff
@@ -33,6 +34,19 @@ def strip_services_wrapper(content):
     return content_json["services"]
 
 
+def get_api_status():
```
allait: These can be replaced with `dmutils.apiclient` now

PR #31, opened 2015-05-05
Extract apis

robyoung:


robyoung:
Added commits to `add-static-filter-data`.

PR #32, opened 2015-05-05
Add static filter data

tombye:
This adds filtering to the search page based on fields shared between G4, G5 and G6 services.

The shared fields were determined by comparing a record of all the G4 and G5 services, formatted to match G6, with the fields defined in the [JSON schemas](https://github.com/alphagov/digitalmarketplace-api/tree/master/json_schemas) used by the content API.

The filters are derived from these fields with their in-page copy taken from the [SSP content repo](https://github.com/alphagov/digital-marketplace-ssp-content).

https://www.pivotaltracker.com/story/show/89414742

The following is left out of this story:

- categories (due to requiring changes to the SSP Content repo)
- uniform URLs for clearer analytics data

Depends on the following pull requests to be merged before it will work:

- https://github.com/alphagov/digitalmarketplace-utils/pull/12
- https://github.com/alphagov/digitalmarketplace-search-api/pull/22

The full page, filtered by the SaaS lot:

search-results


app/main/views.py
```diff
@@ -57,15 +57,28 @@ def get_service_by_id(service_id):
...
         'search_keywords': search_keywords,
-        'filter_groups': search_filters_obj.get_filter_groups(),
-        'services': search_results_json['services']
+        'filter_groups': search_filters_obj.filter_groups,
+        'services': response['services']
     })
     return render_template('search.html', **template_data)
+
+
+def _get_questions():
```
robyoung: This function is no longer called from anywhere.

app/helpers/questions.py
```diff
@@ -0,0 +1,77 @@
...
+                self.__get_dependent_lots__(question_content["dependsOnLots"])
+            ) if "dependsOnLots" in question_content else (
+                ["saas", "paas", "iaas", "scs"]
+            )
+
+            self._question_cache[question] = question_content
+
+        return self._question_cache[question]
+
+    def __populate_section__(self, section):
```
robyoung: This style of method name should be reserved for Python [special methods](https://docs.python.org/2/reference/datamodel.html#special-method-names).

app/__init__.py
```diff
@@ -1,26 +1,42 @@
...
+    application.config.from_object(configs[config_name])
+    configs[config_name].init_app(application)
+    config.init_app(application)
+    g6_questions = QuestionsLoader(
+        manifest="app/helpers/questions_manifest.yml",
+        questions_dir="bower_components/digital-marketplace-ssp-content/g6/"
+    )
+    filter_groups = SearchFilters.get_filter_groups_from_questions(
+        g6_questions.sections
+    )
```
robyoung: Could these two calls be wrapped up into a single call?

PR #33, opened 2015-05-07
Add categories to filters

tombye:
The service types weren't correct in the SSP content data so they could not be included as filters (the buyers app uses the SSP content data to render the search filters).

This was corrected in https://github.com/alphagov/digital-marketplace-ssp-content/pull/38 so this pull request brings in those changes and adds service types to the filters.

image

### DIFF
--- a/app/presenters/search_presenters.py
+++ b/app/presenters/search_presenters.py
@@ -68,13 +68,11 @@ class SearchFilters(object):
             # if the 1st question has options, they will become the
             # filters and it's details will define the group
             if question['type'] in questions_with_options:
-                filters = (
-                    filters +
+                filters += \
                     SearchFilters.get_filters_from_question_with_options(
                         question,
                         len(filters)
                     )
-                )
             else:
                 filters.append(
                     SearchFilters.get_filters_from_default_question(
--- a/app/presenters/search_presenters.py
+++ b/app/presenters/search_presenters.py
@@ -53,14 +53,14 @@ class SearchFilters(object):
         return filter_group
 
     @staticmethod
-    def get_filter_groups_from_questions(manifest=None, questions_dir=None):
+    def get_filter_groups_from_questions(manifest, questions_dir):
         filter_groups = []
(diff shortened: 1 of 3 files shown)


app/presenters/search_presenters.py
```diff
@@ -29,33 +30,58 @@ def get_filters_from_default_question(question):
...
+            filter_name = 'serviceTypes'
+        else:
+            filter_name = question['id']
+        for option in question['options']:
+            filter = {
                 'label': option['label'],
-                'name': question['id'],
-                'id': question['id'] + option['label'].lower(),
+                'name': filter_name,
+                'id': '%s-%s' % (filter_name, option['label'].lower()),
```
robyoung: `option['label']` can have spaces in it. We should replace them with something else.

app/presenters/search_presenters.py
```diff
@@ -29,33 +30,58 @@ def get_filters_from_default_question(question):
...
-            'text': SearchFilters.get_filters_from_default_question,
-            'pricing': SearchFilters.get_filters_from_default_question,
-            'options': SearchFilters.get_filters_from_question_with_options
-        }
         g6_questions = QuestionsLoader(
             manifest=manifest,
             questions_dir=questions_dir
         )
+
+        def add_filters_for_question(question=None, filters=None):
```
robyoung: Should these arguments be optional? It looks like they're required.

app/presenters/search_presenters.py
```diff
@@ -29,33 +30,58 @@ def get_filters_from_default_question(question):
...
+            # if the 1st question has options, they will become the
+            # filters and it's details will define the group
+            if question['type'] in questions_with_options:
+                filters = (
+                    filters +
+                    SearchFilters.get_filters_from_question_with_options(
+                        question,
+                        len(filters)
+                    )
+                )
```
robyoung:
```python
filters += SearchFilters. get_filters_from_question_with_options(
    question, len(filters))
```

PR #35, opened 2015-05-08
Add highlighting to search results

tombye:
Add highlighting to show where matches occurred in each search result.

https://www.pivotaltracker.com/story/show/94070032

highlighting

### DIFF
--- a/app/assets/scss/_search-page.scss
+++ b/app/assets/scss/_search-page.scss
@@ -1,3 +1,5 @@
+@import "_search-result.scss";
+
 .search-page-filters {
   .lot-filters {
     h3 {
@@ -27,10 +29,4 @@
     border: 1px solid $border-colour;
   }
 }
-
-.search-result-excerpt {
-  em {
-    font-style: normal;
-    background-color: $yellow-25;
-  }
 }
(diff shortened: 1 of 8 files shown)


app/assets/scss/_search-page.scss
```diff
@@ -27,3 +27,10 @@
     border: 1px solid $border-colour;
   }
 }
+
+.search-result-excerpt {
+  em {
```
quis: The toolkit [already has SASS for `.search-result-highlighted-text`](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/search-result.html).

app/presenters/search_presenters.py
```diff
@@ -195,3 +196,19 @@ def __set_filter_states(self):
...
                             filter['value'] in param_values
                         )
+
+
+class SearchResults(object):
+    """Provides access to the search results information"""
+
+    def _add_highlighting(self):
+        for index, service in enumerate(self.search_results):
+            if ('highlight' in service):
```
quis: This isn't Javascript :wink:

tombye: Good point, corrected in a47373c

app/assets/scss/_search-page.scss
tombye: Yeah, sorry I missed that. I've added some code to make elasticsearch produce HTML that will let us use it: https://github.com/alphagov/digitalmarketplace-search-api/pull/33.

I've added some new SCSS that will use them instead (once that pull request is merged).

PR #36, opened 2015-05-11
Fix highlighting SCSS & add styles for <em> tag

tombye:
A loose trailing bracket was breaking the Sass compilation and some extra SCSS needed adding to deal with the italic styling inherited from using the `<em>` tag.

### DIFF
--- a/app/assets/scss/_search-page.scss
+++ b/app/assets/scss/_search-page.scss
@@ -1,5 +1,9 @@
 @import "_search-result.scss";
 
+em.search-result-highlighted-text {
+  font-style: normal;
+}
+
 .search-page-filters {
   .lot-filters {
     h3 {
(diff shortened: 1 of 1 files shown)


quis:
This might be better in the toolkit? If you added `font-style: normal` to the declaration for `.search-result-highlighted-text` then it would still override the default styling for `<em>`. I don't think there's a use case where we *would* want the highlighted text to be italicised, in any app.

tombye:
This brings together a few things I'd be interested in adding to the pattern on the toolkit: mandating the use of `<em>`s to mark in-text matches and use of the selector given here to do it. That or list all descendant `<em>`s to have the styles.

I figured there was a discussion worth having about it but on a toolkit pull request with the implementation here used as an example. I'm happy to raise it separately once this is merged.

PR #40, opened 2015-05-13
Fixed issue with integration with utils

minglis:
Changes to utils meant the JSON proesses by the buyer app had changed. This led to key errors.

- Updated the Service presenter code to be aware of the root element
- Added lot to Service object
- Added framework name to results page

### DIFF
--- a/app/templates/_search_results.html
+++ b/app/templates/_search_results.html
@@ -2,21 +2,21 @@
 <div class="search-result">
     <h2 class="search-result-title">
         <a href="/service/{{ service.id }}">{{ service.serviceName }}</a>
-        </h1>
-        <p class="search-result-supplier">
-            {{ service.supplierName }}
-        </p>
+    </h2>
+    <p class="search-result-supplier">
+        {{ service.supplierName }}
+    </p>
 
-        <p class="search-result-excerpt">
-            {{ service.serviceSummary }}
-        </p>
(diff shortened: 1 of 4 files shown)


app/templates/_search_results.html
```diff
@@ -1,18 +1,22 @@
...
-    {{ service.serviceSummary }}
-  </p>
-  <ul aria-label="tags" class="search-result-metadata">
-    <li class="search-result-metadata-item">
-      {{ service.lot }}
-    </li>
-  </ul>
+    <h2 class="search-result-title">
+        <a href="/service/{{ service.id }}">{{ service.serviceName }}</a>
+        </h1>
```
robyoung: I know you didn't introduce this but the header tags don't match here.

app/templates/_search_results.html
```diff
@@ -1,18 +1,22 @@
...
+            {{ service.serviceSummary }}
+        </p>
+        <ul aria-label="tags" class="search-result-metadata">
+            <li class="search-result-metadata-item">
+                {{ service.lot }}
+            </li>
+            <li class="search-result-metadata-item">
+                {{ service.frameworkName }}
+            </li>
+        </ul>
```
robyoung: Why has the indentation on this file been changed? It makes spotting the actual change quite a bit harder.

PR #41, opened 2015-05-13
Changed url of service page

minglis:
- We refer to services everywhere (API and so on)
- Buyer app page is service (singluar)
- Updated to be services as per other facets of the DM.


robyoung:
There should be at least one test around this view.

quis:
You'll want to look at this nomenclature [in the admin app](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views.py#L53-L89) as well.

minglis:
Tests added ready to go.

minglis:
@quis URLs are interesting - we've plumped for services plural in all the APIs. Was bringing this into line. Note that the grails app is service singular.

We need to do a redirect for all grails URLs if we, as Ash wants, add framework into the urls:

/g-cloud/search
/g-cloud/services
/digital-specialists/search
etc

So we can change from service to services as part of that if we choose to do it.

quis:
@minglis So shouldn't this PR also do the redirect from `/service/…` to `/services/…`?

robyoung:
We have a [`@deprecated`](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/deprecation.py) decorator in the utils package for monitoring changes like this.

TheDoubleK:
Unless anyone has strong feelings about this I'm tempted to get this merged now - adding the redirects is part of this upcoming story: https://www.pivotaltracker.com/n/projects/997836

quis:
@TheDoubleK What story? That link just goes to the project in Pivotal. LGTM :+1: though if there is a followup to deal with redirects/deprecation.

TheDoubleK:
Sorry - https://www.pivotaltracker.com/story/show/93097694

PR #42, opened 2015-05-13
Changed Header

minglis:
- Buyer app has no login so removed
- Supplier app needs to be linked to so "login" link is now "suppliers"
- Removed "create account" as currently can't do this.


quis:
This partly addresses [this story](https://www.pivotaltracker.com/story/show/93978934). Might need to revisit the log in link because it's only a G-Cloud suppliers log in. Going to merge for the time being though as this represents progress.

PR #43, opened 2015-05-14
WIP: Change to use updated API response format …

robyoung:
The search API response has been updated to be more in line with the
JSON API standard. This change also makes the `links` object available
to the buyers app.

NOTE: This commit is not complete. It depends on https://github.com/alphagov/digitalmarketplace-utils/pull/34 before the pagination can be integrated.


robyoung:
Closing until utils is available.

PR #51, opened 2015-05-18
Add trailing slashes to URLs

quis:
This URL works:
https://preview.development.digitalmarketplace.service.gov.uk/buyers-guide

This URL doesn't work:
https://preview.development.digitalmarketplace.service.gov.uk/buyers-guide/

We need to support both because the existing app uses URLs with trailing
slashes, which users may have bookmarked.

That we standardise on having trailing slashes in our URLs. If we go
with this option then Flask handles the redirects for us.

---

The alternative would be something like the following, but it doesn't feel as clean:
``` Python
@app.before_request
def remove_trailing_slash():
    if request.path != '/' and request.path.endswith('/'):
        return redirect(request.path[:-1])
```

Bring thy reckons.

### DIFF
--- a/app/templates/index-g-cloud.html
+++ b/app/templates/index-g-cloud.html
@@ -16,7 +16,7 @@
     </h1>
   </header>
   <div class="grid-row">
-    <form action="/search" method="get" class="search-box column-two-thirds">
+    <form action="/search/" method="get" class="search-box column-two-thirds">
         <input type="text" class="text-box" name="q">
         <button type="submit" class="button-save">
           Show services
@@ -25,7 +25,7 @@
   </div>
   <ul class="browse-list">
     <li class="browse-list-item">
-      <a href="/search?q=&lot=saas" class="browse-list-item-link">
+      <a href="/search/?q=&lot=saas" class="browse-list-item-link">
           Software as a Service
       </a>
       <p class="browse-list-item-body">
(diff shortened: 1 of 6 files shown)


app/templates/index-g-cloud.html
```diff
@@ -25,7 +25,7 @@
   </div>
   <ul class="browse-list">
     <li class="browse-list-item">
-      <a href="/search?q=&lot=saas" class="browse-list-item-link">
+      <a href="/search/?q=&lot=saas" class="browse-list-item-link">
```
TheDoubleK: Not directly related to this PR, I know, but is the empty `q=` param required here?

TheDoubleK:
This seems the neatest way of doing things, although I'm not mad on the trailing slashes.

app/templates/index-g-cloud.html
quis: @tombye Will know the answer to that one. It definitely was at one point.

quis:
We've changed our minds. Going to default to slashless URLs. Will close this and open a separate PR.

app/templates/index-g-cloud.html
tombye: It is at the moment, I've added this as a bug: https://www.pivotaltracker.com/story/show/94931176

PR #53, opened 2015-05-19
Redirect URLs with query parameters, add tests

quis:
This pull request adds the improvements to trailing slash redirection from the admin and supplier apps.

It also preserves query parameters when doing the redirection, in case anyone has a URL like `/search/?q=cloud` bookmarked.


minglis:
I'm assuming this is the only way to do this?

PR #54, opened 2015-05-20
G cloud links

minglis:
Delivers: https://www.pivotaltracker.com/story/show/93097694

- Story to prefix search and services pages with the framework name
- As currently we only have a single framework this is hardcoded to g-cloud
- Added redirects to the existing urls to the new ones
- Decorated these old routes with the deprecated annotation
- Added tests for new and old endpoints

- Fixed bug where empty keywords caused an error.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -216,10 +216,7 @@ def redirect_search():
 
 @main.route('/g-cloud/search')
 def search():
-    print "HERE"
     search_keywords = get_keywords_from_request(request)
-    print " KEY {}".format(search_keywords)
-    print " dONE"
     search_filters_obj = SearchFilters(blueprint=main, request=request)
     response = search_api_client.search_services(
         **dict([a for a in request.args.lists()]))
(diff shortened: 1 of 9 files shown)


app/main/views.py
```diff
@@ -185,6 +196,12 @@ def get_service_by_id(service_id):
 
 
 @main.route('/search')
+@deprecated(dies_at=datetime(2016, 1, 1))
+def redirect_search():
+    return redirect(url_for(".search", **request.args), code=301)
```
pcraig3: Cool, does this pass along all arguments included in the original request?  Does it do GET and POST arguments?

I was looking for something like this the other day to use with the supplier app.

allait: @pcraig3 `.args` only contains GET parameters. There's also `request.form` for form POST params and `request.values` with the combined GET and POST parameters. But you probably only want GET for redirects.

allait:
Can't be merged since it's behind master now, but other than that LGTM :+1:
It's great that we're transitioning links to `url_for` in templates!

PR #55, opened 2015-05-20
PROTOTYPE: Make search asynchronous

quis:
This is prototype quality code. It is not intended to be merged, but you might find it interesting.

It modifies the search endpoint to return JSON, and adds some Javascript to the front end to load the search results dynamically every time the user interacts with the filters or search box.

ajax


minglis:
Closed this as not for merging. Runs on heroku on: http://dm-auto-reload.herokuapp.com/

PR #56, opened 2015-05-20
Show Contact Information on Service Pages

pcraig3:
Decided to just do two calls to the API on service pages: one to get service, and one to get supplier (which includes contact information.)

Added a supplier fixture as well for the purposes of the test. 

> [Story on Pivotal](https://www.pivotaltracker.com/story/show/91385166)


allait:
:+1:

PR #57, opened 2015-05-20
Error pages

minglis:
Added error handlers and 404/500 pages to buyer app

- Error page content from Roz


TheDoubleK:
Tested with Debug mode off - looks good.

PR #58, opened 2015-05-21
Display documents on G5 Service pages.  

pcraig3:
### Problem
[G6 Service Pages show documents](https://preview.development.digitalmarketplace.service.gov.uk/services/5061011874775040).
[G5 Service Pages don't](https://preview.development.digitalmarketplace.service.gov.uk/services/5-G4-1046-001).

### Solution

Now they do.

> [Story in Pivotal](https://www.pivotaltracker.com/story/show/94960826)


pcraig3:
***

As an addendum, a G6 service's document is returned with the following keys:
```json
"termsAndConditionsDocument": "APM0112 Terms and Conditions - Direct Services v.1.2.pdf",
"termsAndConditionsDocumentURL": "https://assets.digitalmarketplace.service.gov.uk/documents/{supplier_id}/{service_id}-terms-and-conditions.pdf",
```

While a G5 service's document looks like this:
```json
"termsAndConditionsDocumentURL": "http://assets-production.govstore.service.gov.uk/G5/1046/{service_id}/QD2/APM0112 Terms and Conditions - Direct Services v.1.2.pdf",
```

The reason that G5 documents weren't being show was that document keys without the `URL` suffix were being looked for.  
G6 documents seem to be rewriting the URL a more generalized one composed of the a service's {supplier_id}/{service_id}-{name-of-document}.pdf, while G5 services don't do this.

Do we care about this?  Does it matter that G5 and G6 services have different ways of linking to their documents?

TheDoubleK:
Some (many) G5 servcies also have "Additional documents" as well as the four G6 document types.  See e.g. https://www.digitalmarketplace.service.gov.uk/service/5-g5-0175-002.

It would be good to link to these additional documents too.

TheDoubleK:
The document URLs point to where the documents are - we can't easily change these.  The G5 servcies are in Govstore but G6 docs are in S3.  The G5 SSP didn't rename any documents, or have a sensible system of folders, hence the ugly URLs.  For G6 we changed all uploaded document names to the standard pattern, but kept the original document name just in case  (e.g. in the  "termsAndConditionsDocument" field).

I don't think these original document names should be being used at all though?  They certainly shouldn't be being shown in the frontend AFAIK. If the frontend is using them anywhere then we should probably change it to only look for the "xxDocumentURL" keys, which are the same for both G5 and G6 services. (With the addition of "additionalDocumentURLs" for G5).

TheDoubleK:
For additional documents it looks like the Grails app uses the existing filename (minus the extension) as link text.  Seems like a reasonable approach?

pcraig3:
Grails App Documents == Buyers App Documents

#### [Live Grails app](https://www.digitalmarketplace.service.gov.uk/service/5-g4-1046-001): 

screen shot 2015-05-21 at 17 07 44

#### [Buyer App](http://localhost:5002/services/5-G4-1046-001) : 
(links to localhost :/ )

screen shot 2015-05-21 at 17 07 54

TheDoubleK:
The ordering of documents is different in the screenshots here, and looking at our display code could be different for different services.  I think we should decide on an ordering of documents and enforce that on service pages, rather than just returning them in the order they happen to appear in the JSON.  The order from the Grails app looks like it's probably right to me, but should run it by some design folks to check:
1. Service definition
2. T&Cs
3. Pricing
4. SFIA rate card [If it exists]
5. Additional documents [G5 only]

TheDoubleK:
Checked with Cath and ordering of docs will go in as a new story.

pcraig3:
Obviously not a useful consideration, but I liked the flow from smaller to larger.

screen shot 2015-05-22 at 12 19 36

PR #63, opened 2015-05-28
Redirect /service/{lowercased_id} routes.

pcraig3:
Redirect urls that would have been (and are currently being) generated by the Grails app. 
Remove Martin's `/service/{id}` route because nobody is relying on it.

##### Note: 
Both the Admin and the Supplier Apps let you view a live service by pointing to their live digitalmarketplace urls with hardcoded strings (they're building a `/service/{id}` url), so these don't have to be fixed now, but they will have to be eventually.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/94569554)

The story on pivotal says we want to enforce a casing on our IDs.
Actually though, after talking it over with @allait, none of our apps are generating or using lowercased IDs.  They're all being generated by the grails app, so once it's retired we won't have the problem of mixed cased IDs.
_However_, we should redirect urls that the grails app was using.  

And that is how this pull request relates to the story on pivotal.


tombye:
The search api client has changed which is breaking the search page. If you rebase you should get the new api client.

tombye:
LGMT :+1:

PR #65, opened 2015-05-28
Add summary to search results

tombye:
Add a paragraph at the top of the search results to summarise the search performed.

https://www.pivotaltracker.com/story/show/93972040

search_summary

As explained in the story description, this should match the look and feel of the one on gov.uk:

image

**Some detail on how it works:**

It presents the following information from the search:

- keywords
- lot
- search filters, organised into the groups they appear in in-page

This is formed into a sentence using the information in the [search_summary_manifest YAML](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/add-summary-to-search-results/app/helpers/search_summary_manifest.yml) file


TheDoubleK:
It would be nice to have an ellipsis at the end of search summaries (especially where sentences are cut off part way through).

TheDoubleK:
Exact match searches in quotes don't seem to generate snippets with the matched phrase shown - e.g. "cloud hosting", "threat protection".

Not sure why this is?

TheDoubleK:
As discussed - these things are not part of this story/pull-request so ignore comments.  I will add a story in Pivotal to look at improving the search snippet generation.

tests/unit/test_search_presenters.py
```diff
@@ -109,37 +112,800 @@ class TestSearchSummary(unittest.TestCase):
...
+        summary_fragment.filters = [u"FreeOption"]
+        self.rules['filtersPreposition'] = 'with'
+        self.assertEqual(summary_fragment.str(), u"categories with FreeOption")
+
+
+filter_groups = [
+    {
+        'depends_on_lots': ['scs', 'saas', 'iaas'],
+        'filters': [{
+            'id': 'serviceTypes-implementation',
```
minglis: Is all this stuff a duplicate from the YAML files? Can we import that instead? Save duplication, breaks if changes, and so on.

app/helpers/search_helpers.py
```diff
@@ -1,6 +1,18 @@
 from math import ceil
 
 
+def get_lot_label(lot_param_value):
```
minglis: Should this end up in utils at some point? Can see buyer/supplier apps utilising it.

tombye: Makes sense. Think it should be outside of this pull request though.

tests/unit/test_search_presenters.py
tombye: I've made that change in https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/67674bf08ba3d0d32ba7563e0ddbdc4e58446f91

app/presenters/search_presenters.py
```diff
@@ -205,17 +212,253 @@ def _add_highlighting(self):
...
+                self.filters_fragments.append(
+                    SummaryFragment(
+                        group_id=group_id,
+                        filters=filters,
+                        rules=group_rules)
+                )
+
+    def _set_initial_sentence(self, results_total, request_args):
+        template = u"{} found containing {} in {}"
+        keywords = u"{}&ldquo;{}&rdquo;{}".format(
```
quis: I don't like the omnipresent quotes here—makes it ambiguous whether you're doing a phrase search. There is a difference between these two searches (screenshots from the live marketplace):

image

image

TheDoubleK: Agreed.  Might also need something to change quotes into fancy quotes for phrase search strings, as they come through as plain quotes at the moment:
screen shot 2015-06-08 at 11 23 39

quis: Typographic quotes would be nice if you could do them. In the Grails app [I did them in Javascript](https://github.gds/gds/digitalmarketplace-frontend/blob/68d13a578fb5f37018c4aff84154cbd8e7f66167/grails-app/assets/javascripts/_curlyQuotes.js), but the same logic (or better) could work on the backend.

TheDoubleK: The summary for an empty search string looks better on the existing app - new:
screen shot 2015-06-08 at 11 36 05

Old:
screen shot 2015-06-08 at 11 38 49

TheDoubleK: * Ssh! Don't ask about the numbers being different.

tombye: I've removed mention of the keywords when they don't exist but kept the lot in:

image

Thoughts?

app/helpers/search_helpers.py
pcraig3: <sub>There's also [this function](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/helpers/service_helpers.py) that seems to do the same thing.</sub>

app/presenters/search_presenters.py
quis: I like the _found in "All categories"_. Makes it explicit.

TheDoubleK: Me too.

tombye:
@pcraig3 I merged the helper you found with the one I made so there's only one now: b2cfa34

TheDoubleK:
Nice work, Tom.

PR #67, opened 2015-05-29
Handle service ID not found

robyoung:
get_service returns None if the API returns a 404. This may change in
the future to raise an HTTPError (which is already handled) but this
change will just become dead code then.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -195,6 +195,8 @@ def redirect_service_page(service_id):
 def get_service_by_id(service_id):
     try:
         service = data_api_client.get_service(service_id)
+        if service is None:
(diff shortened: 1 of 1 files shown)


TheDoubleK:
:+1: Finally!  I thought this had been fixed several times, can't believe it's taken this long to sort it.

Do we care about the Python 3 failures?  They seem unrelated to this PR (and more to do with pagination), but might want to fix before merging?

robyoung:
I have fixed the Python 3 failure in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/69. I'll rebase this PR once the other one is merged.

TheDoubleK:
#69 (dudes!) is merged.

PR #70, opened 2015-05-29
Only published services should be displayed.

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/94942196

Currently 'enabled' and 'disabled' services are displayed by the buyer frontend.
This change makes the buyer app only display published services.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -197,6 +197,9 @@ def get_service_by_id(service_id):
         service = data_api_client.get_service(service_id)
         if service is None:
             abort(404, "Service ID '{}' can not be found".format(service_id))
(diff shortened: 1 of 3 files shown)


app/main/views.py
```diff
@@ -197,6 +197,9 @@ def get_service_by_id(service_id):
         service = data_api_client.get_service(service_id)
         if service is None:
             abort(404, "Service ID '{}' can not be found".format(service_id))
+        if service['services'].get('status') != 'published':
```
allait: Can we `or` this condition with the previous one since the outcome is the same?

TheDoubleK: Yep, should have done that.

PR #71, opened 2015-06-01
Fixed breadcrumbs.  Updated tests.

pcraig3:
* Added breadcrumbs to search page (which didn't have them before.) 
* Fixed breadcrumbs on service pages where they (a) didn't link back to search, or (b) include the 'cloud technology and support' page.
* Removed current page title as last breadcrumb I found them.  It's not how gov.uk ([random example page](https://www.gov.uk/apply-funding-community-project)) do it, so I figured we don't want to do it either.

Also rewrote the tests using lxml (recommended by @allait).  

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/95241386)

***

#### Before: Search Results
screen shot 2015-06-01 at 12 17 59

#### After: Search Results
screen shot 2015-06-01 at 12 18 28

***

#### Before: Service Page
screen shot 2015-06-01 at 12 19 14

#### After: Service Page
screen shot 2015-06-01 at 12 19 03

***

#### Before: Page Title in Breadcrumbs
screen shot 2015-06-01 at 12 20 38

#### After: Page Title removed from Breadcrumbs
screen shot 2015-06-01 at 12 20 25


pcraig3:
Ready to merge whenever.  Ralph has seen it.

tests/app/views/test_services.py
```diff
@@ -78,13 +116,15 @@ def _assert_service_page_url(self):
...
         res = self.client.get('/g-cloud/services/{}'.format(service_id))
         assert_equal(200, res.status_code)
-        assert_true(
-            "<h1>{}</h1>".format(service_title)
-            in res.get_data(as_text=True)
-        )
 
-        self._assert_contact_details(res)
-        self._assert_document_links(res)
+        document = html.fromstring(res.get_data(as_text=True))
```
TheDoubleK: Is it worth having this new dependency just for this one method?  It would be as easy to assert that there is `<h1>service_title</h1>` in the content, without building the HTML and tracing an xpath.

tests/app/views/test_services.py
```diff
@@ -78,13 +116,15 @@ def _assert_service_page_url(self):
...
+
+        page_title = document.xpath('//div[@id="wrapper"]//h1/text()')[0]
+        assert_equal("{}".format(service_title), page_title)
```
TheDoubleK: Why not just `assert_equal(service_title, page_title)`?

That passes fine with Python 2 - does Python 3 break it or something?

TheDoubleK:
I left a couple of comments about the tests, bit otherwise looks good :+1:

tests/app/views/test_services.py
allait: It's not the only method `lxml` is used for: there are tests for breadcrumbs and links that would be brittle with string matching.

TheDoubleK: Ah sorry, yes I missed that.

PR #72, opened 2015-06-05
Hide certifications header if there aren't any

tombye:
Some services in lot 4 have no certifications but still show that section of attributes. They should hide them if this is the case.

See https://www.pivotaltracker.com/story/show/95242202 for reference.

Note: some of the tests for attributes with empty lists referenced the 'vendorCertifications' field so broke when this change was made. This pull request includes fixes for them.

### DIFF
--- a/app/presenters/service_presenters.py
+++ b/app/presenters/service_presenters.py
@@ -53,19 +53,20 @@ class Service(object):
                 continue
             data_value = attribute.get_data_value()
             data_type = attribute.get_data_type(data_value)
-            if data_type is not False:
-                current_row = {
-                    'key': row['key'],
-                    'value': data_value,
-                    'type': data_type
-                }
-                if hasattr(attribute, 'assurance'):
-                    current_row = attribute.add_assurance_to_row(
-                        attribute.assurance,
-                        current_row
-                    )
+            if data_type is not False and \
+                    attribute.is_empty_certifications_field() is False:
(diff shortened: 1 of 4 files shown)


app/presenters/service_presenters.py
```diff
@@ -53,19 +53,20 @@ def _get_rows(self, group, service_data):
...
-                }
-                if hasattr(attribute, 'assurance'):
-                    current_row = attribute.add_assurance_to_row(
-                        attribute.assurance,
-                        current_row
-                    )
-
-                rows.append(current_row)
+            if data_type is not False and \
+                    attribute.is_empty_certifications_field() is False:
```
quis: I find the indentation here confusing. Something like this is clearer to me:
``` python
if (
    condition is true and
    other_condition is false
):
```

quis:
I wonder if this could be more generic. I _think_ that the generic case here is where you have
- a list entry or a series of checkboxes
- which are optional

This is data that is (mostly) held by the content repo at the moment—don't know if that integration is work that's worth doing now.

tombye:
I talked to Cath about whether there was a generic need here but she said after going through the services this was the only field where it was a good idea to hide the absence of the data.

This is why I ended up making it as being very specific to that case.

quis:
@tombye OK, my other comment notwithstanding `:+1:` from me.

PR #73, opened 2015-06-08
Incorporate Feature Flags.

pcraig3:
The upgraded `dmutils` sets up FeatureFlags (among others) in its `init_app` function, and includes the `get_version_label` and `get_flags` methods.

The `init_app` function also sets up `logging`, `config`, and `proxy_fix`, so we don't have to import them into our api anymore.


pcraig3:
Updated to require as-yet unreleased version of dmutils, so it's going to break until [this pull request](https://github.com/alphagov/digitalmarketplace-utils/pull/54) is merged.

PR #74, opened 2015-06-09
Add DM_HOST environment variable

tombye:
Allows you to run the app against your local IP over the network.

### DIFF
--- a/application.py
+++ b/application.py
@@ -6,7 +6,8 @@ from flask.ext.script import Manager, Server
 
 application = create_app(os.getenv('DM_ENVIRONMENT') or 'development')
 manager = Manager(application)
-manager.add_command("runserver", Server(port=5002))
+manager.add_command("runserver", Server(
+    os.getenv('DM_HOST') or '127.0.0.1', port=5002))
 
 if __name__ == '__main__':
     manager.run()
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ not already been set:
 ./scripts/run_app.sh
 ```
 
+If you want the app to be available on the network, you need to set the `DM_HOST` environment variable to '0.0.0.0':
+
+```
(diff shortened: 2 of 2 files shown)


application.py
```diff
@@ -6,7 +6,8 @@
 
 application = create_app(os.getenv('DM_ENVIRONMENT') or 'development')
 manager = Manager(application)
-manager.add_command("runserver", Server(port=5002))
+manager.add_command("runserver", Server(
+    os.getenv('DM_HOST') or '127.0.0.1', port=5002))
```
TheDoubleK: This works for me.  Are we 100% sure that '127.0.0.1' is the right value for when it's deployed to AWS?

TheDoubleK:
If we're not 100% sure and it isn't easy to find out then happy to merge to preview and check it still works!

application.py
tombye: Should the value for AWS be the default then? I based it on the `DM_ENVIRONMENT` variable which defaults to a development setting.

I guess I need to look into changing the AWS configuration before this can be merged yeah?

TheDoubleK: I don't know - it just seemed like specifically setting the address the server runs on might have an impact.  If AWS needs it to run on something other than `127.0.0.1` then it would need to be set as DM_HOST in the Cloudformation templates.  But as far as I know this might just work.  @allait probably knows more than me about this to say.  Or we could just suck it and see.

allait:
Why do we need this? This won't affect AWS since the apps are run by apache and application.py is just used to import the WSGI app.

If this is just for local apps then we don't need the environment variable: `python application.py runserver --help` will list the built-in flags, which include `-h HOST, --host HOST` and `-p PORT, --port PORT`

tombye:
Great so we don't need this. Closing...

PR #75, opened 2015-06-10
Make gulpfile generic

quis:
This copies the Gulpfile from the Suppliers app, which is now generic, and makes the necessary changes to the setup of this app so that it works.

It also bumps the toolkit version to make sure that the Gulpfile copies the templates.


tombye:
This is great and much clearer. I've added a few reckons which would be worth discussing.

tombye:
:+1:

PR #80, opened 2015-06-12
Deny third-party sites from embedding in an iframe

pcraig3:
Upgrading to new verison of dmutils which initiates each app with the `X-Frame-Options` header set to `DENY`.
Also added test to confirm.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Flask-Script==2.0.5
 requests==2.5.1
 inflection==0.2.1
 pyyaml==3.11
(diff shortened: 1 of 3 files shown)


quis:
I think it would be better if the test was its own test, rather than being tacked onto the analytics one. Just makes it clearer _why_ you're testing for that specific header.

pcraig3:
Right, how does that look?  Moved this test and the one in the supplier app into roughly the same place.

quis:
@pcraig3 No, yes, totally.

PR #85, opened 2015-06-30
Refactor search presenters

allait:
Pivotal story: https://www.pivotaltracker.com/story/show/95284020

### Major refactor of building search filters
- Refactor presenters to use new content builders
- Split search presenter class into several classes
- logic deferred to content builder on lots

### 	Add functions to clean/validate request to search_helpers
`build_search_query` will match request args with known filters
to remove any key-value pairs that don't match the rendered form.

It will also remove any unknown query parameters, and will only keep
`page`, `q` and `lot` if they are set.

Resulting query will only contain fields and values that could be set
through the UI and will group OR and AND filters:

* Questions of type "radios" result in OR filters - there's only one
  possible value for the service, so searching for more than one option
  returns a union of results for both options.
* Questions of type "checkbox" result in AND filters - each service can
  have multiple values for the field, so searching for more than one
  option returns only services that contain all matching values.

Since we haven't refactored SearchSummary presenter it won't use the new
cleaned request filters, so any incorrect filters/values set in the
query parameters will result in a 'None' string in search summary text.


quis:
What is `app/templates/macros/forms.html` being used for in 148db05ca22204a6da2b0bd0c816b40a9fe9742d? I don't see it being imported anywhere?

allait:
@quis I think it's a copy of the file from suppliers-frontend that got accidentally committed. Good spot, I'll rebase it out.

PR #86, opened 2015-07-01
[#98094170] Prettify G5 additional filenames

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/98094170

Rather than re-introduce spaces to the filenames for link text, I think it is better to standardise on "Additional document #1", etc. for consistency.

This PR also bumps the PEP8 line length limit to 120 characters.

Before: 
screen shot 2015-07-01 at 12 29 10

After:
screen shot 2015-07-01 at 13 29 54

### DIFF
--- a/app/presenters/service_presenters.py
+++ b/app/presenters/service_presenters.py
@@ -231,18 +231,18 @@ class Meta(object):
                 extension = self._get_document_extension(url)
                 documents.append({
                     'name':  names[index],
-                    'url': self._replace_whitespace(url, '%20'),
+                    'url': url,
                     'extension': extension
                 })
 
         # get additional documents, if they exist
         if 'additionalDocumentURLs' in service_data:
-            for document_url in service_data['additionalDocumentURLs']:
-                extension = self._get_document_extension(document_url)
-                name = self._get_document_name_without_extension(document_url)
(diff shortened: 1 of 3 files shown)


TheDoubleK:
I've also removed a couple of methods that are no longer required since the G4/G5 export script was fixed to make document links valid URIs.

robyoung:
Why do we not want to keep the name they give their files? Additional document #1 doesn't seem very helpful.

TheDoubleK:
We can do it either way, but some filenames are very long and break the style of the page. Some also have rubbish names with weird Unicode characters in them. @minglis thought it would be better to standardise names, as all the others are standardised - we don't preserve their filenames for service description or anything. But if there are strong feelings wet should preserve their filenames it's an easy fix.

TheDoubleK:
Although the long ones would wrap if we put the spaces back in and not be quite so bad.

TheDoubleK:
I've restored the spaces to filenames.  Also noticed that there is an issue with underscores - e.g. on the Grails app we have (https://www.digitalmarketplace.service.gov.uk/service/5-g4-0328-246):
screen shot 2015-07-02 at 10 22 38

So I've also replaced underscores in filenames with spaces, which results in:
screen shot 2015-07-02 at 10 23 02

PR #88, opened 2015-07-03
Supplier a to z

minglis:
Adds supplier A-Z pages

- adds a footer link to A-Z pages
- url is /g-cloud/suppliers for list and /g-cloud/supplier/{id} for supplier details
- only gets suppliers with an active service on a live framework

- implements; https://www.pivotaltracker.com/n/projects/997836/stories/94085326


.buildpacks
```diff
@@ -0,0 +1,2 @@
+https://github.com/heroku/heroku-buildpack-nodejs.git
```
allait: Probably shouldn't have been commited

Procfile
```diff
@@ -0,0 +1 @@
+web: python application.py runserver
```
allait: Probably shouldn't have been commited

app/templates/content/suppliers-guide-g-cloud.html
```diff
@@ -32,7 +32,7 @@
       <div class="govspeak">
   <h2 id="about-g-cloud">
   <span class="number">1. </span>Digital Marketplace and G-Cloud</h2>
-
+TEST
```
allait: ^^

application.py
```diff
@@ -4,9 +4,10 @@
 from app import create_app
 from flask.ext.script import Manager, Server
 
+port = int(os.environ.get('PORT', 5002))
 application = create_app(os.getenv('DM_ENVIRONMENT') or 'development')
 manager = Manager(application)
-manager.add_command("runserver", Server(port=5002))
+manager.add_command("runserver", Server(host='0.0.0.0', port=port))
```
allait: We can customize port with `-p` argument, I assume this is part of the accidental commit of heroku-related changes

package.json
```diff
@@ -20,6 +20,6 @@
     "frontend-build:development" : "./node_modules/gulp/bin/gulp.js build:development",
     "frontend-build:production" : "./node_modules/gulp/bin/gulp.js build:production",
     "frontend-build:watch" : "./node_modules/gulp/bin/gulp.js watch",
-    "postinstall": "npm run frontend-install"
+    "postinstall": "npm run frontend-install && npm run frontend-build:production"
```
allait: I'd keep the `package.json` files the same between the frontend apps, I don' think any of the others are running the build on install

scripts/run_app.sh
```diff
@@ -10,8 +10,8 @@ npm run frontend-build:production
 echo ""
 
 # Use default environment vars for localhost if not already set
-export DM_DATA_API_URL=${DM_DATA_API_URL:=http://localhost:5000}
-export DM_DATA_API_AUTH_TOKEN=${DM_DATA_API_AUTH_TOKEN:=myToken}
+export DM_DATA_API_URL=${DM_API_URL:=http://localhost:5000}
```
allait: `DM_DATA_API_URL` is the name of the new variable that all apps were switched to back in April. Are we going back to the old one or was this an accidental change?

src/digitalmarketplace-utils
```diff
@@ -0,0 +1 @@
+Subproject commit 3160906d58ac57b532b8cc8731278a142fdf8ccb
```
allait: Shouldn't be commited

package.json
quis: Agree. Any changes to these scripts or the gulpfile should be the same across all apps, otherwise it gets a bit painful.

allait:
`tmp` shouldn't be commited.

Is it possible to squash this removing merges and any files and commits that shouldn't have been part of the PR and joining the ones with the same/similar messages and changes?

package.json
minglis: all of these relate to my erroneous heroku commit.

Sorting

app/templates/content/suppliers-guide-g-cloud.html
minglis: Dunno what I was thinking with that, or even what I was testing 😳

quis:
Tweaked a few things based on a conversation with @ralph-hawkins, :+1: from us two now.

PR #99, opened 2015-07-17
Remove warning banners from supplier A–Z

quis:
Since we're not showing any of the supplier's services on these pages it's not possible for a buyer to use them to do a non compliant procurement.

Therefore we don't need to show the warning.

Discussed with Caroline this morning.

After
--
image

Before
--
image


minglis:
👍ship it

quis:
Awaiting confirmation from Caroline…

PR #100, opened 2015-07-17
Move contact details if no clients/description

quis:
It looks weird to have the contact details all the way over to the right of the page if the left hand column is empty. This commit removes the left hand column if a supplier has no clients or description.

After
--
image

Before
--
image


minglis:
👍OK. ship it.

PR #101, opened 2015-07-20
Use correct analytics properties for live/other

quis:
This commit makes the app select different Google analytics properties to track against depending on which domain the user is browsing.

This will ensure continuity of analytics when we switch the DNS.


minglis:
Checked with Ash, he seems happy.

👍

Aviator74:
This should hopefully work

PR #102, opened 2015-07-20
Fix cookie link on seen cookies message warning to use local cookies …

minglis:
…page

- seen cookie message pointed to GOV.UK
- changed to use /cookies


quis:
:+1:

PR #104, opened 2015-07-20
Fix search summary missing keywords containing unicode characters

allait:
`request.args.get` with `type` returns a default value (`''`) if a
ValueError is raised by the `type` callable. Since `str` raises an
exception on non-ASCII characters, `type=str` removes any keywords
with unicode characters from the search summary text.

Removing the `type` argument fixes the issue. The `request.args` is
unicode by default, so there's no need to add any type conversion.


minglis:
👍

quis:
Nice: http://www.fileformat.info/info/unicode/char/1f47e/index.htm

PR #106, opened 2015-07-21
Move search button underneath for mobile layouts

tombye:
Allows the search box to be as big as possible on smaller screens.

This also includes a shift from pixels to ems for the button width to make it relative to the font rendering which was causing an issue on IE8, Windows XP.

### Desktop/tablet

search_box_desktop

### Mobile

search_box_mobile

### DIFF
--- a/app/assets/scss/_index-page.scss
+++ b/app/assets/scss/_index-page.scss
@@ -21,6 +21,7 @@
 
     button {
       height: 48px;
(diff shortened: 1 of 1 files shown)


quis:
Can we have the button full-width for mobile please? Considering that the majority of users are right handed, a button all the way over to the left will be a stretch.

PR #107, opened 2015-07-21
Make service presenters more shareable

quis:
_This pull request is a precursor to:_
- _extracting the service presenters into utils_
- _sharing them between apps_
- _having them understand SSP concepts like an "incomplete" row_

Use summary table from frontend toolkit 
--
This bumps the toolkit version to (primarily) bring in: alphagov/digitalmarketplace-frontend-toolkit#107

It then
- return service attributes in the structure expected by the template
- return service attribute values as rendered HTML, ready for the template
- refactors the service presenters to decouple the Attribute presenter from the Service presenter
- scopes the CSS on the cookies page to not affect the tables on the service page

Use content loader
--

Previously the service presenter used a hard-coded mapping to know which service attributes to put on the page.

This commit changes the service presenter to use the content loader in dmutils instead.

It adds a separate manifest for service pages, which is derived from the previous hard-coded mapping (so should have no effect on which attributes are displayed)

Use 16px type in sidebar
--

This commit changes the sidebar where service meta information is displayed to have less variation in type sizes, no weird button-ish service ID, and restores the spacing and copy-paste-ability of the service ID.

Before | After
--------- | ------
screen-live|screen-local


tombye:
LGMT now :+1:

Great work here btw, these changes manage to simplify the code while preserving pretty much all of the interface. Will make a big difference.

PR #108, opened 2015-07-21
Support searching for '.' versions of the G5 service IDs

allait:
[#99026308](https://www.pivotaltracker.com/story/show/99026308)

When forming a request for the search-api buyer frontend will look
for G5 service IDs in the query string and try to replace dots with
'-'. This way, search-api always gets a request containing '-' IDs.

This allows us to support searches for '.' service IDs without making
any changes to the search-api index.

Only service IDs with all-'.' are suppoted, mixing '.' and '-' will
not trigger the replacement and probably won't return any results.

**Note:** I haven't found any other G5 service ID regular expressions in our apps so I'm not sure if all of them have the same `5-G<digit>-<4 digits>-<3 digits>` format, but I haven't seen any that didn't.

### DIFF
--- a/app/helpers/search_helpers.py
+++ b/app/helpers/search_helpers.py
@@ -1,3 +1,4 @@
+import re
 from math import ceil
 
 from werkzeug.datastructures import MultiDict
@@ -103,6 +104,16 @@ def group_request_filters(request_filters, content_loader):
     return filter_query
 
 
+def replace_g5_search_dots(keywords_query):
+    """Replaces '.' with '-' in G5 service IDs to support old ID search format."""
+
+    return re.sub(
+        r'5\.G(\d)\.(\d{4})\.(\d{3})',
+        r'5-G\1-\2-\3',
+        keywords_query
+    )
+
+
 def build_search_query(request, lot_filters, content_loader):
     """Match request args with known filters.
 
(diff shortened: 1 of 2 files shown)


quis:
Is it worth dumping all the G5 service IDs and `grep`ing (or similar) through them to see if any don't match your regex?

allait:
@quis that's a great idea. I've done this with the old DB dump: 
```bash

$ jq '.services[] | .id' ~/Downloads/listings_dump.json | head
"4-G2-0391-003"
"5-G5-1768-001"
"5-G5-0263-261"
"5-G4-0267-202"
"5-G4-0267-201"
"5-G4-0267-204"
"5-G5-1668-004"
"5-G5-1551-004"
"5-G5-1551-006"
"5-G5-1551-007"

$ jq '.services[] | .id' ~/Downloads/listings_dump.json | wc -l
16443

$ jq '.services[] | .id' ~/Downloads/listings_dump.json | grep '5-G\d-\d\{4\}-\d\{3\}' | wc -l
9200

$ jq '.services[] | .id' ~/Downloads/listings_dump.json | grep '4-G\d-\d\{4\}-\d\{3\}' | wc -l
7243
```

quis:
`9200 + 7243 = 16443` so :+1:

PR #110, opened 2015-07-23
Html audit

tombye:
Changes to improve the semantics of the Buyers' app pages.

https://www.pivotaltracker.com/story/show/98821678

I'd recommend reading this commit by commit as the changes are deliberately grouped in this way for clarity.

### DIFF
--- a/app/assets/scss/_error-pages.scss
+++ b/app/assets/scss/_error-pages.scss
@@ -1,0 +1,3 @@
+.error-page p {
+  @include core-19;
+}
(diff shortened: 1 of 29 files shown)


quis:
Can we also add `<title>` tags to any pages that are missing them. Cursory look suggests `404.html` and `500.html`.

tombye:
Added the page title back to the error pages in e39b0d1

quis:
:+1:  to

PR #111, opened 2015-07-23
Make URLs for assets cache-proof

quis:
This commit adds a query string to assets URLs which is generated from a hash of the file contents. When asset files are changed they will now be served from a different URL, which means they wont be loaded from browser cache.

This is similar to how GOV.UK template adds its version number as a querystring parameter for its assets.

Addresses [98936730](https://www.pivotaltracker.com/story/show/98936730)


tombye:
LGTM :+1:

PR #114, opened 2015-07-24
Refactor service presenters to use new new tables

quis:
Brings in: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/116

--

**This pull request changes the Attribute() class to be a mapping between the SSP content and the data returned from the API, and the field templates in the summary table pattern in the toolkit.** Pfewf.

--

Might be better explained with examples:

The method takes:
- question type from content, eg `checkboxes`, `percentage`
- value from the api, eg `["Gold", "Silver"]`, `99.99`

It returns an object which has:
- the field template that should be used to render the content, eg `list`, `text`
- the formatted value for that field, eg `["Gold", "Silver"]`, `99.99%`

It still takes care of assurance values and all that business.

It now also takes `label` as a parameter, which means one less method (`_get_row`) in the buyer app.

### DIFF
--- a/app/helpers/service_page_manifest.yml
+++ b/app/helpers/service_page_manifest.yml
@@ -88,6 +88,7 @@
     - dataStorageMediaDisposal
     - dataSecureEquipmentDisposal
     - dataRedundantEquipmentAccountsRevoked
+    - serviceAvailabilityPercentage
 -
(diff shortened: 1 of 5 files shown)


allait:
:+1:

PR #115, opened 2015-07-24
Fixes for broken functional tests

tombye:
Makes the lot in the search breadcrumb use the
label rather than the short form and corrects a
spelling error with the Digital Marketplace part.


robyoung:
:+1:

PR #117, opened 2015-07-29
Make the 4xx apache logs alarm more specific

robyoung:
Only fire the alarm if the rate of 400 - 403s goes over the limit. 404s happen quite frequently and are not indicative of anything going wrong.

See https://github.com/alphagov/digitalmarketplace-search-api/pull/50


allait:
:+1:

PR #118, opened 2015-07-29
updating links on supplier lists

minglis:
- should use 'other' as url param
- needs to ship AFTER https://github.com/alphagov/digitalmarketplace-api/pull/192


quis:
Can the link in the footer change from "Supplier A-Z" to "Supplier A–Z" while we're here?

:trollface:

minglis:
@quis 😡

PR #120, opened 2015-07-31
Let the toolkit handle the type of service attributes

quis:
This commit defers the handling of what type of value an attribute is to the toolkit, which is possible because the toolkit [now has macros for these fields](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/summary-table.html).

So all that the Attribute class now does is:
- provide the value and assurance to the template in a consistent format
- format the assurance value
- determines which answers are 'required' and marks them as such

We will be able to reuse this in the supplier app, partially fulfilling: https://www.pivotaltracker.com/story/show/98558406

### DIFF
--- a/app/presenters/service_presenters.py
+++ b/app/presenters/service_presenters.py
@@ -65,12 +65,21 @@ class Service(object):
 class Attribute(object):
     """Wrapper to handle accessing an attribute in service_data"""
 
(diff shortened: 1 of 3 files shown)


tombye:
:+1:

PR #121, opened 2015-07-31
Add cross domain tracking to analytics

tombye:
https://www.pivotaltracker.com/story/show/98185416

The above story specified the use of raw Google Analytics (GA) code but we use the [GOVUK analytics code](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/analytics.md), a wrapper intended to allow analytics tracking without being tying the project to a vendor's code.

The raw GA code included the following requirements to allow cross-domain tracking.

1. loading the `linker` plugin
2. defining the domain to link to
3. a change to the tracker creation, setting its `allowLinker` option to `true`

Points 1. and 2. were possible with the existing GOVUK analytics code but 3. required some changes to be made (see fd1fcc6). I'm investigating this change being pushed into the shared GOVUK analytics code but for now, this patches our version to add the required functionality.

I've added a JavaScript test for this change. This repo didn't have any JavaScript tests so f799c84 includes the addition of a browser-based Jasmine test suite so the test can run.

To run the test, open `spec/javascripts/support/LocalTestRunner.html` in a browser.


quis:
:+1:

PR #124, opened 2015-08-05
Fix cross-domain tracking

tombye:
The cross-domain tracking is currently setting 2 trackers which is producing duplicate pageviews.

https://www.pivotaltracker.com/story/show/100627412

As described by Persicope, our analytics contractor, the following calls to the `ga` method should be removed:

```
ga(‘create’, ‘UA-49258698-1’, ‘auto’, { ‘name’: ‘link’ });
ga(‘link.require’, ‘linker’)
ga(‘link.linker:autoLink’, [‘digitalservicesstore.service.gov.uk’])
ga(‘link.set’, ‘anonymizeIp’, true)
ga(‘link.send’, ‘pageview’)
```

...while these should be preserved:

```
ga(‘create’, ‘UA-49258698-1’, { ‘cookieDomain’: ‘.digitalmarketplace.service.gov.uk’, ‘allowLinker’: true });
ga(‘set’, ‘anonymizeIp’, true)
ga(‘send’, ‘pageview’)
ga(‘require’, ‘linker’)
ga(‘linker:autoLink’, [‘digitalservicesstore.service.gov.uk’])
```

The GOVUK.Analytics code sets an additional tracker for cross-domain tracking by default so this part of the code has had to be rewritten to allow the above requirements.

The tests can be run by navigating to `/spec/javascripts/support` and opening `LocalTestRunner.html` in a browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,3 +106,8 @@ npm install
 
 Note: `npm run frontend-install` is run automatically as a post-install task when you run `npm install`.
 
+## Frontend tests
+
+To run the JavaScript tests, navigate to `spec/javascripts/support/` and open `LocalTestRunner.html` in a browser.
+
+TODO: Add a Gulp task which is run as part of `./scripts/run_tests.sh`.
--- a/app/assets/javascripts/analytics/_dm-analytics.js
+++ b/app/assets/javascripts/analytics/_dm-analytics.js
@@ -1,18 +1,38 @@
 (function() {
   "use strict";
+  // Rewritten GOVUK.Analytics
+  var DMAnalytics = function(config) {
+    this.trackers = [];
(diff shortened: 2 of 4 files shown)


robyoung:
How do we run the jasmine tests?

tombye:
@robyoung I just added a note about that (at the bottom of the description above)!

robyoung:
Could we get these wired into `./scripts/run_tests.sh`.

tombye:
@robyoung yes I agree, and I did have that but it got derailed by a problem with the Node module used (see https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/7535023de16ff9042615af7d66f85498f24bfc0c).

robyoung:
:+1:

PR #125, opened 2015-08-05
Remove supplier guide link from preview

tombye:
We need to be able to view the sidebar module without it until further notice:

g-cloud-7-side-bar-pre-launch-without-supplier-guide

g-cloud-7-side-bar-post-launch-without-supplier-guide


quis:
:+1: If there is a sensible reason for this happening.

PR #132, opened 2015-08-12
Change to homepage copy

quis:
From CCS this morning

image

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -54,7 +54,7 @@
       </h2>
     {% endif %}
     <p>
-      Register your interest in becoming a G&#8209;Cloud&nbsp;7 supplier
+      Register your interest in becoming a G&#8209;Cloud&nbsp;7 supplier.
     </p>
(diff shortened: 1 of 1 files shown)


tombye:
Do we have to explain what 'BST' means?

tombye:
Talking with @superroz, writing it in full would cause more disruption. We could wrap it in an `<abbr>` tag  to put the information in there though.

quis:
@tombye Good idea. Added in d9fedfdf37b03b47241eff9fe8c1164254790926.

superroz:
Happy to change to British Summer Time (BST) if we think BST isn't well understood. Could we have a full stop after 'in becoming a G-Cloud 7 supplier'?

quis:
@superroz 53204de

tombye:
This is what it would look like if we used the full form:

image

Thoughts?

tombye:
Talking with @quis and @superroz the form as it currently is is most suitable.

:+1:

PR #137, opened 2015-08-18
Switch on G-Cloud 7 notice on staging

robyoung:
We need to test it before going live.


TheDoubleK:
Looks good.  Should the G_CLOUD_7_SUPPLIER_GUIDE be turned on in preview as well as dev?

TheDoubleK:
But if you want to keep this just to the staging change then go for it. :+1:

robyoung:
I don't think so, because it's not gone live yet. I'm not sure though. This is to get just the notice up on staging ready to switch it on tomorrow afternoon.

PR #143, opened 2015-08-26
Update suppliers-guide-g-cloud.html

superroz:
updated link text for creating an account and logging into an account. the previous link only took users to the login page.


quis:
:+1:

PR #145, opened 2015-08-28
Use DM-Request-ID header from the response

robyoung:
Use the DM-Request-ID header from the response rather than the request.
As this app is what generates the request ID (Nginx strips any that may
have been nefariously set from outside) there will be no DM-Request-ID
request header here.

http://httpd.apache.org/docs/2.2/mod/mod_log_config.html#formats

See: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/242

### DIFF
--- a/.ebextensions/apache_log.config
+++ b/.ebextensions/apache_log.config
@@ -4,8 +4,8 @@ files:
     owner: root
     group: root
     content: |
-      LogFormat "apache-access buyer-frontend \"%{DM-Request-ID}i\" %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" cloudwatchlogs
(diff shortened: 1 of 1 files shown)


.ebextensions/apache_log.config
```diff
@@ -4,8 +4,8 @@ files:
     owner: root
     group: root
     content: |
-      LogFormat "apache-access buyer-frontend \"%{DM-Request-ID}i\" %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" cloudwatchlogs
+      LogFormat "apache-access buyer-frontend \"%{DM-Request-ID}0\" %h %l %u %t \"%r\" %>s %b \"%{Referer}i\" \"%{User-Agent}i\"" cloudwatchlogs
```
allait: That's a zero

robyoung: :trollface:

allait:
:+1:

PR #148, opened 2015-09-01
Switch on G7 links

minglis:


quis:
Matches staging config, :+1:

PR #150, opened 2015-09-02
Remove supplier A–Z feature flag

quis:
Per https://www.pivotaltracker.com/story/show/102080450

> Some feature flags we will want to switch off again in future (e.g. G7 Submissions open), but others will never be turned off again and so should be removed from the codebase.


TheDoubleK:
:+1:

PR #154, opened 2015-09-04
Fix typo in CloudWatch application logs

robyoung:
I had entered the LogGroupName twice rather than adding a
JSONLogGroupName


allait:
:+1:

PR #157, opened 2015-09-08
[#102845416] Fix 500s for bad service ids

TheDoubleK:
Should fix this bug: https://www.pivotaltracker.com/story/show/102845416

Most of the fix is in utils - this just pulls utils in and fixes an error message.

Depends on this utils update going in first:
- [x] https://github.com/alphagov/digitalmarketplace-utils/pull/161

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -1,5 +1,7 @@
 # coding=utf-8
 
+from __future__ import unicode_literals
+
 from datetime import datetime
 from flask import abort, render_template, request, redirect, \
     url_for, current_app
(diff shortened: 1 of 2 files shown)


robyoung:
I wonder if importing `unicode_literals` from `__future__` would be more appropriate given that we're not really porting a legacy Python 2 API.

http://python-future.org/unicode_literals.html
https://docs.djangoproject.com/en/1.8/topics/python3/#unicode-literals

TheDoubleK:
That looks like a nice solution. Will do the same for utils too.

TheDoubleK:
Done, but needs the corresponding Utils PR to go in again now.

robyoung:
:+1:

PR #159, opened 2015-09-09
Allow the JavaScript tests to run in CI

tombye:
You can currently only run the JavaScript tests by loading the test runner in a browser.

This adds a Gulp task which uses PhantomJS to run the tests so it can be done from a command as well.

This is a copy of the same changes made to the supplier app:

https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/264

### DIFF
--- a/scripts/run_tests.sh
+++ b/scripts/run_tests.sh
@@ -36,4 +36,7 @@ npm run --silent frontend-build:production
 display_result $? 1 "Build of front end static assets"
 
 nosetests -v -s --with-doctest
-display_result $? 3 "Unit tests"
+display_result $? 3 "Python unit tests"
+
+npm test
+display_result $? 3 "JavaScript unit tests"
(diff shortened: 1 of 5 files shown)


scripts/run_tests.sh
```diff
@@ -36,4 +36,7 @@ npm run --silent frontend-build:production
 display_result $? 1 "Build of front end static assets"
 
 nosetests -v -s --with-doctest
-display_result $? 3 "Unit tests"
+display_result $? 3 "Python unit tests"
+
+npm test
+display_result $? 3 "JavaScript unit tests"
```
robyoung: If we're using different exit statuses we should probably use a different one for the JavaScript tests as well so that we can tell the difference between Python and JavaScript test failures.

robyoung:
Other than the exit status :+1:

PR #160, opened 2015-09-10
Update node modules

tombye:
Bumping some old Node modules and replacing the deprecated gulp-uglifyjs with gulp-uglify, as requested by its messaging.

See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/275 for comparable changes on the suppliers' app.


quis:
:+1:

PR #161, opened 2015-09-10
Fix parsing of document URLs with semicolons in them

TheDoubleK:
This will stop this page from giving a 500 error: https://www.digitalmarketplace.service.gov.uk/g-cloud/services/5-G5-0248-010

Before:
screen shot 2015-09-10 at 14 54 49

After:
screen shot 2015-09-10 at 14 54 58

Some G5 document URLs (well, at least one) have semicolons in them, which confuses `urlparse`, which splits off anything after a semicolon into params.

By replacing the semicolons with their URL-encoded escape then the methods changed here work properly.

As the methods are only used for finding the document extension and extracting a 'nice' document title, we don't need to worry overly about making the substitution here as the change will not be propagated to any displayed links, etc.

### DIFF
--- a/app/presenters/service_presenters.py
+++ b/app/presenters/service_presenters.py
@@ -180,12 +180,12 @@ class Meta(object):
         return caveats
 
     def _get_pretty_document_name_without_extension(self, document_url):
-        document_basename = os.path.basename(urlparse(document_url).path)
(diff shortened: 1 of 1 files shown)


quis:
:+1:

PR #162, opened 2015-09-16
Use bold for search highlights

quis:
Brings in alphagov/digitalmarketplace-frontend-toolkit#155

To match alphagov/frontend#879


TheDoubleK:
:+1:

PR #163, opened 2015-09-16
Update the README file so that tests can be run

pcraig3:
Yesterday working alongside Andrew, we noticed that the information in the README is a bit out of order.
Basically, you'd need to install `requirements_for_test.txt` and also make sure your npm dependencies are installed beforehand.


quis:
:+1:

PR #164, opened 2015-09-17
Fix yellow mark element

quis:
Brings in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/156

> Highlights were changed to use the `<mark>` element in alphagov/digitalmarketplace-search-api#54

> Some browsers, by default, style the `<mark>` element with a bright yellow background.

> This makes sure that doesn't happen.

aaeaaqaaaaaaaaikaaaajgzlmzc5ogrkltexntgtngjmzi1imgjklti1mzq0ytk0mznkzg

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "dependencies": {
     "jquery": "1.11.2",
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Yellow Mark should be our new logo.
:+1:

PR #167, opened 2015-09-23
Bring in corrected grid classes

tombye:
Brings in https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/161 which moves the grid class names used in this repo to the DM Toolkit.


quis:
:+1:

PR #169, opened 2015-09-23
Update summary manifest to match content change

quis:
The label for one of the filters was changed in https://github.com/alphagov/digitalmarketplace-frameworks/commit/881734dc27cc5430267512142c72390193dc13e0

This was causing `500`s in production because users were able to filter by something
that was not defined in the manifest.

This is an immediate fix. There is probably a longer-term solution that makes this more
robust


robyoung:
:+1: to this and :+1: to doing a longer term fix

PR #170, opened 2015-09-28
Added a check for a 404 from list suppliers call

minglis:
- could get a 5xx in the supplier app as the no suppliers response is a 404
- returns a 404 if invalid paginated query

Fix for pivotal:
https://www.pivotaltracker.com/story/show/104300146


TheDoubleK:
Looks good :+1:

PR #172, opened 2015-10-02
Update homepage for G7 submissions closed

TheDoubleK:
When the feature `FEATURE_FLAGS_G_CLOUD_7_IS_LIVE` is False, the home page will look like this:

screen shot 2015-10-02 at 09 12 17

Compare to the design here:
http://dm-prototype.herokuapp.com/states/homepage?g7.framework=inReview


tombye:
I've added a fix for a problem with the HTML (see 8ffcefc) but other than that :+1:

PR #173, opened 2015-10-05
Do not use priceString for displayed price

robyoung:
Now that we're moving to G7 the price should be built from the price parts like it is elsewhere.


allait:
:+1:

PR #174, opened 2015-10-05
Fix page size environment variable

robyoung:
The page size needs to be an integer but when passed in with an environment variable it is always a string. This converts it to an integer at the point of use.

### DIFF
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -188,7 +188,7 @@ def search():
 
     pagination_config = pagination(
         search_results_obj.total,
-        current_app.config["DM_SEARCH_PAGE_SIZE"],
+        int(current_app.config["DM_SEARCH_PAGE_SIZE"]),
         get_page_from_request(request)
     )
 


TheDoubleK:
Should there be a try-catch round the cast to int? Or we want it to break right away if the environment variable is wrong?

Has search pagination been broken up until now then?

allait:
I think it should be converted when it's assigned in the `config.py`: if the only type that makes sense is an integer we should convert to it as early as possible. Since this moves the check to app startup not catching the exception should be fine.

robyoung:
@TheDoubleK Pagination hasn't been broken up until now because we haven't set it with environment variables. We use the configuration value set in `config.py`

@allait Yes, ideally we would do the conversion as close to the place that it's defined as possible. However, this is a bit difficult. The code that takes config options from the environment is in `dmutils` and doesn't have any idea what type a value is supposed to be, so far we have only had string configuration options.

robyoung:
After discussion in real life this is going to be handled over [here](https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/config.py#L9)

robyoung:
Fixed with https://github.com/alphagov/digitalmarketplace-utils/pull/180

PR #175, opened 2015-10-06
Turn off G7-Live feature flag in prod

TheDoubleK:
To deploy at 3pm - immediately after the API database migration.


robyoung:
:+1:

PR #176, opened 2015-10-06
Replace the spacing below <p>'s in homepage notice

tombye:
[This line](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/scss/_index-page.scss#L48), included to set the spacing of the homepage G7 notice paragraphs, was cancelling out the spacing between paragraphs in the post-G7-closing notice.

This replaces the spacing to that in the original pattern.

### Before

image

### After

image

### DIFF
--- a/app/assets/scss/_index-page.scss
+++ b/app/assets/scss/_index-page.scss
@@ -75,4 +75,9 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


TheDoubleK:
:+1:

PR #177, opened 2015-10-09
Content update after zendesk feedback

superroz:
Update to 'choosing the right call-off contract' content. Old content implied the G6 services were in some way more up-to-date than G5 services.


tombye:
:+1:

PR #179, opened 2015-10-15
Add API error handling similar to supplier app

robyoung:
Currently 404s returned by the API are resulting in 500s on the buyer app for invalid supplier requests. This change takes the model used in the supplier frontend to automatically handle uncaught APIErrors. It is not identical to the supplier app version because the default template data is generated in a different way.


pcraig3:
:+1:

PR #181, opened 2015-10-21
Lowercase lot filters

allait:
Related to alphagov/digitalmarketplace-frameworks#99 and bower.json needs to be updated with a new SHA once frameworks PR is merged.

### Show full lot names in search results for services with lowercase lot slugs

New API returns service documents with lowercase lot slugs, which means
that once these service documents are indexed search API responses will
contain lowercase slugs in lot fields. Since lotName fields aren't
indexed yet we switch the comparison to be case-insensitive for search
result lot name display.

### Use lowercase lot slugs for finding category filters
Framework files should support both lowercase and mixed-case lots,
so using the new lot format for all content builder filter calls.

(alphagov/digitalmarketplace-frameworks#99)

### Rename digital-marketplace-ssp-content to digitalmarketplace-frameworks


tombye:
:+1:

PR #182, opened 2015-10-22
Watch for changes to content

quis:
Implements https://github.com/alphagov/digitalmarketplace-utils/pull/193

Adds https://pypi.python.org/pypi/watchdog as a requirement to make watching for file changes use system events, which is faster

### DIFF
--- a/application.py
+++ b/application.py
@@ -2,12 +2,10 @@
 
 import os
 from app import create_app
-from flask.ext.script import Manager
 from dmutils import init_manager
 
 application = create_app(os.getenv('DM_ENVIRONMENT') or 'development')
-manager = Manager(application)
-init_manager(manager, 5002, ['./app/content/frameworks'])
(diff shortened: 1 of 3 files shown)


quis:
# Have flask init init the manager
Depends on and implements alphagov/digitalmarketplace-utils#195

quis:
# Bring in 11.1.1 of utils

To allow us to use the version of `init_manager` which returns a `Manager` instance.

robyoung:
Once tests pass :+1:

quis:
@tombye Sorry. Tired. Fixed in 23584d2acb98f2602a48f0ef075a5e3b079bc57e.

tombye:
@quis I did exactly the same thing, the first go at https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/7db91ecea684598b5264f285a85110f20aa03a00, just had a bit more time to correct it :)

PR #183, opened 2015-10-27
[#105005304] DM homepage message contingent on framework status

pcraig3:
This story changes the message on the Digital Marketplace homepage by looking for the message associated with the first framework it finds with the highest-priority status.  The list of statuses we care about is kept [in the route for the homepage](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/pc_manage_mainpage_messages/app/main/views.py#L44). 
So, for example:
- If `pending` was the status we were most interested in, then [the message for `g-cloud-7`](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-7/messages/homepage-sidebar.yml#L1) would be returned.  [Same as exists currently](https://www.digitalmarketplace.service.gov.uk/).

- If we were looking for a `coming` framework, [the message for `digital-outcomes-and-specialists`](https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists/messages/homepage-sidebar.yml#L1) would be returned.  [Same as Ralph's prototype](http://dm-prototype.herokuapp.com/states/homepage?dos.framework=coming).

- If we were looking for `[ "open", "coming", "pending" ]` frameworks (and there were no open frameworks), then the `coming` message for `digital-outcomes-and-specialists` would be returned.

Also, I removed all of the feature flags as all of them were being used to control how content appeared on the homepage.

##### To keep in mind 

- If two frameworks are in the same status of equal priority (ie, if two are `pending` and none are `open`), then the one returned first in the API response will be selected
- If a framework with a valid status has no corresponding `.yml` file in its `messages/` directory, the app returns a 500.  Since it's on the Digital Marketplace homepage, it's probably best to avoid this where possible.

***
:trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet: :tada: :trumpet:

screen shot 2015-10-27 at 14 55 06

### DIFF
--- a/app/helpers/shared_helpers.py
+++ b/app/helpers/shared_helpers.py
@@ -8,3 +8,12 @@ def get_label_for_lot_param(lot_param):
     }
     if lot_param in lots:
         return lots[lot_param]
+
+
+def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_order_of_preference):
+    for status in statuses_in_order_of_preference:
+        for framework in frameworks:
+            if framework.get('status') == status:
+                return framework
+
+    return None
--- a/app/helpers/shared_helpers.py
+++ b/app/helpers/shared_helpers.py
@@ -15,5 +15,3 @@ def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_o
         for framework in frameworks:
             if framework.get('status') == status:
                 return framework
-
-    return None
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -33,7 +33,21 @@ from .. import search_api_client, data_api_client, content_loader
 @main.route('/')
 def index():
     template_data = get_template_data(main, {})
-    return render_template('index.html', **template_data)
+
+    framework_slug = 'digital-outcomes-and-specialists'
+    framework_slug = 'g-cloud-7'
+    # get framework status in a better way
+    framework_status = 'pending'
+    block = 'homepage-sidebar'
+
+    content_loader.load_messages(framework_slug, [block])
+    temporary_message = content_loader.get_message(framework_slug, block, framework_status)
+
+    return render_template(
+        'index.html',
+        temporary_message=temporary_message,
+        **template_data
+    )
 
 
 @main.route('/g-cloud')
--- a/app/main/views.py
+++ b/app/main/views.py
@@ -10,6 +10,7 @@ from dmutils.deprecation import deprecated
 from dmutils.formats import get_label_for_lot_param
 from dmutils.apiclient import HTTPError
 from dmutils.formats import LOTS
+from dmutils.apiclient import APIError
 
 from . import main
 from ..presenters.search_presenters import (
@@ -19,6 +20,7 @@ from ..presenters.search_presenters import (
 from ..presenters.search_results import SearchResults
 from ..presenters.search_summary import SearchSummary
 from ..presenters.service_presenters import Service
+from ..helpers.shared_helpers import get_one_framework_by_status_in_order_of_preference
 from ..helpers.search_helpers import (
     get_keywords_from_request, get_template_data, pagination,
     get_page_from_request, query_args_for_pagination,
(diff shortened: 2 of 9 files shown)


app/main/views.py
```diff
@@ -33,7 +35,37 @@
...
+
+        content_loader.load_messages(framework.get('slug'), ['homepage-sidebar'])
+        temporary_message = content_loader.get_message(
+            framework.get('slug'),
+            'homepage-sidebar',
+            framework.get('status')
+        )
+
+    # if no framework is found (should never happen), ditch the message and load the page
+    except (APIError, AttributeError):
```
robyoung: Where can `AttributeError` come from here?

app/templates/index.html
```diff
@@ -39,79 +38,20 @@
...
-          <li><a href="/suppliers">Log in to existing supplier account</a></li>
-          <li><a href="/suppliers/create">Create new supplier account</a></li>
-        </ul>
-      </nav>
-    </aside>
-    {% else %}
-      <div class="g-cloud-7-notice column-one-third">
+    <div class="framework-message column-one-third">
+      {% if temporary_message %}
+        {# define all variables used by any of the expected templates #}
```
robyoung: I'm not sure this comment is necessary and I'm a bit nervous about comments directly in templates (as apposed to in macro templates) as these files tend to get chopped around a bit and therefore comments can easily be end up separated from the thing they're supposed to be commenting.

config.py
```diff
@@ -72,34 +69,24 @@ def init_app(app):
...
 
 class Live(Config):
     DEBUG = False
-    FEATURE_FLAGS_G_CLOUD_7_NOTICE = enabled_since('2015-08-20')
-    FEATURE_FLAGS_G_CLOUD_7_IS_LIVE = False
 
 
 class Staging(Live):
-    FEATURE_FLAGS_G_CLOUD_7_NOTICE = enabled_since('2015-08-18')
-    FEATURE_FLAGS_G_CLOUD_7_IS_LIVE = False
```
robyoung: :tada:

tests/app/views/test_application.py
```diff
@@ -20,3 +24,125 @@ def test_should_use_local_cookie_page_on_cookie_message(self):
...
+
+        return {
+            'frameworks': _frameworks
+        }
+
+    @staticmethod
+    def _assert_message_container_is_empty(response_data):
+        empty_message_container = '<div class="framework-message column-one-third"></div>'
+        assert_in(
+            BaseApplicationTest._strip_whitespace(empty_message_container),
```
robyoung: Why not just `self`? Or better yet, having `_strip_whitespace` as a function in it's own right (although that would be a bit of a bigger change).

app/main/views.py
pcraig3: If `data_api_client.find_frameworks()` returns 
```json
{ 
  "frameworks": {}
}
```
then `framework.get('slug')` will throw an error.
Don't think it's possible to happen in a real-life scenario.

tests/app/views/test_application.py
pcraig3: Yeah, `self` works.  

Incidentally, the [buyer](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/tests/helpers.py#L72), [supplier](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/dcfd39653357aca908492f4fe93cabd6065c333d/tests/app/helpers.py#L77), and [admin](https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/614cc0655b100d30dff11b80e512baf23799c650/tests/app/helpers.py#L98) frontend apps all have their own versions of whitespace stripping methods that are all named slightly different things... :/

app/main/views.py
robyoung: Could you just check that `framework` is not null? That feels more explicit than catching the `AttributeError`.

app/helpers/shared_helpers.py
```diff
@@ -8,3 +8,12 @@ def get_label_for_lot_param(lot_param):
...
         return lots[lot_param]
+
+
+def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_order_of_preference):
+    for status in statuses_in_order_of_preference:
+        for framework in frameworks:
+            if framework.get('status') == status:
+                return framework
+
+    return None
```
robyoung: This is redundant. Functions return `None` by default.

pcraig3: Good to know :+1:

tests/app/views/test_application.py
robyoung: Yay! I've noticed that this is pretty endemic in the buyer frontend tests. Maybe it's not something that should be tackled in this PR.

robyoung:
:+1:

PR #185, opened 2015-11-04
Add error handling for 410 responses

TheDoubleK:
The API will soon return 410 ("Gone") responses for requests for services on an expired framework.

This shows users a nice error page if they request details of services on an expired framework.

See: https://www.pivotaltracker.com/story/show/107382498

This is related to, but not dependent on: https://github.com/alphagov/digitalmarketplace-api/pull/286

Looks like:
screen shot 2015-11-04 at 12 17 47

### DIFF
--- a/app/templates/errors/410.html
+++ b/app/templates/errors/410.html
@@ -1,0 +1,20 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Service no longer available - Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+<div class="error-page">
+  <header class="page-heading-smaller">
+    <h1>Service no longer available</h1>
+  </header>
+
+  <p>
+      The requested service was on a framework that is no longer live and so it is no longer available through the Digital Marketplace.
+  </p>
+  <p>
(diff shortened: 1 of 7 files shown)


quis:
Column should be ⅔<sup>rds</sup> wide so the line length isn’t too long. You should check the wording with @superroz as well.

app/templates/errors/410.html
```diff
@@ -0,0 +1,23 @@
...
+{% block main_content %}
+
+<div class="error-page">
+    <div class="index-page grid-row">
+        <div class="column-two-thirds">
+            <header class="page-heading-smaller">
+                <h1>Service no longer available</h1>
+            </header>
+            <p>
+                The requested service was on a framework that is no longer live and so it is no longer available through the Digital Marketplace.
```
robyoung: This message is specifically about gone services. I'm not sure it should be in a generic 410 template. Could we handle the 410 explicitly in the service show view if we need a specific service show message?

TheDoubleK: I did think about that.  The first message I wrote was a generic "this isn't here any more", but then decided a service-specific message might be nicer as that's the only thing we have that makes 410s.  But yeah in that case it probably shouldn't be a generic error page.

TheDoubleK:
* Columns are narrower
* Better message from Roz
* "Service gone" message is now only for service page 410s
* Generic message for other 410s

screen shot 2015-11-05 at 14 11 34

superroz:
Can we change 'contact us at' to 'email'?

TheDoubleK:
Yep of course, sorry I missed that.

quis:
:+1:

PR #186, opened 2015-11-09
Handle framework status in frontend

robyoung:
The framework status is now being passed back in the service response. 404ing and 410ing based on framework status can now be done in the buyer frontend which allows the API to start returning services for frameworks that are not live.

## Depends on
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/290


quis:
:+1:

PR #196, opened 2015-12-18
Fix filters accessibility issue

tombye:
Story: https://www.pivotaltracker.com/story/show/105291876

image

### TLDR

If an option-select (which we use as the UI for our filters, shown above) was in the collapsed state you would assume that the hidden options would be removed from the tabbing order of the document. The design instead worked by keeping the options in the tabbing order and opening the parent option-select if they were tabbed to.

This brings in changes from the latest option-select govuk component which fixes this.


quis:
:+1:

PR #200, opened 2016-01-07
Update dmutils and add dmapiclient package dependency

allait:
### Update dmutils and add dmapiclient package dependency
Fixes imports to use the new dmapiclient package instead of
dmutils.apiclient and dmutils.audit.

### Remove unused app.models module and dependencies
app.models seems to have been replaced by search API client and
is not imported anywhere

### Update gulp-sass to work on node v4.2
We're using gulp-sass 2.0.4 on supplier app and it seems to work
with both old and new node versions.

Also sets npm and travis node versions to v4.2.

### Switch from nostests to py.test runner
Using py.test allows us to send coverage reports similar to other
apps.

### Enable coveralls coverage reports


robyoung:
:+1::+1::+1: Nice cleanup!

PR #201, opened 2016-01-07
Remove unused heading

tombye:
This heading should have been removed now G-Cloud 7 is closed (meaning there are only existing
suppliers).

The existing code was also breaking due to a malformed `<h3>` tag:

image

Note: requires review by @superroz


TheDoubleK:
New suppliers can still sign up though... although applications only open for DOS at the moment.

TheDoubleK:
But still, the header is unnecessary I think :+1:

superroz:
Yep. The header isn't needed. Good to go. :+1:

PR #202, opened 2016-01-12
Unavailable service design

tombye:
This is dependant on https://github.com/alphagov/digitalmarketplace-api/pull/322 being merged.

Mark services that are unavailable to buy (by either their framework having expired or their not being published) as such on their page rather than returning a 404.

This ensures that, once they have been published, services will always have a live URL, as requested in this story:

https://www.pivotaltracker.com/story/show/110062348

The two states of the notification banner are as shown:

### Service no longer published

service-disabled-banner

### Service part of an expired framework

framework-expired-banner


tests/app/views/test_services.py
```diff
@@ -2,10 +2,29 @@
...
+
+    def __init__(self, document):
+        self.banner = document.xpath(
+            '//div[@id="wrapper"]//div[@class="banner-temporary-message-without-action"]'
+        )
+
+    def exists(self):
+        if not self.banner:
+            return False
+        return True
```
robyoung: This could just be ```return bool(self.banner)```

tests/app/views/test_services.py
```diff
@@ -157,34 +215,118 @@ def test_g6_service_page_url(self):
...
+        self.service = self._get_g6_service_fixture_data()
+        self._data_api_client.get_service.return_value = self.service
+        service_id = self.service['services']['id']
+        res = self.client.get('/g-cloud/services/{}'.format(service_id))
+        assert_equal(200, res.status_code)
+        document = html.fromstring(res.get_data(as_text=True))
+
+        banner = document.xpath(
+            '//div[@id="wrapper"]//div[@class="banner-temporary-message-without-action"]'
+        )
```
robyoung: Could `UnavailableBanner` be used here?

tombye:
@robyoung nice call with the `return bool(self.banner)`, didn't think of that. I've also made it a `@property` as it read better when fed to an assertion.

That and the extra use of the `UnavailableBanner` class are now added in ca3fd4e

robyoung:
Nice :+1:

PR #205, opened 2016-01-22
Update buyers-guide-g-cloud.html

superroz:
fixed broken 'Digital Marketplace overview' link


TheDoubleK:
:+1: 
I'm not sure why the coveralls test is unhappy... I think adding a character may have made the percentage of code covered drop by a tiny fraction of a percent, so it's unhappy.  Best bet might be to disable the check so we can merge it.

PR #209, opened 2016-01-29
Bump frontend toolkit

carolinegreen:
Bump frontend toolkit to fix bug where pressing enter on textbox causes synthetic click event on buttons in form.


pcraig3:
Seems :+1: to me

PR #210, opened 2016-01-29
Move login to buyer app

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/111536514

This copies login functionality from the supplier app, allowing login by users with any role.  It also adds routes for creating user accounts with `buyer` role.  

It also does a bit of refactoring - the single `views` file is split into logical sections, and a couple of routes with a deprecation date of 1st Jan 2016 have been removed.

This is ready for review, but not for merging yet until after the following things are done:
 - [x] Messaging OK'd by @superroz and/or @ralph-hawkins (e.g. `Invalid email domain` message) 
 - [x] Admin app bugfix to stop it relying on the deprecated `/service/<id>` route deleted here
 - [x] Supplier app changes to remove login and password reset pages
 - [x] Credentaials update to make sure supplier and buyer app have identical secret keys so that login can be shared between both apps
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/235
 - [x] https://github.com/alphagov/digitalmarketplace-apiclient/pull/8


app/main/views/crown_hosting.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
robyoung: :tada: high five for splitting these out!

app/templates/_proposition_header.html
```diff
@@ -4,7 +4,11 @@
     <nav id="proposition-menu">
       <a href="/" id="proposition-name">Digital Marketplace</a>
       <ul id="proposition-links">
-        <li><a class="home" href="/suppliers">Supplier login</a></li>
+        {% if current_user.email_address %}
+        <li><a class="home" href="{{ url_for('main.logout') }}">Log out</a></li>
+        {% else %}
+        <li><a class="home" href="{{ url_for('main.render_login') }}">Log in</a></li>
+        {% endif %}
         <li><a href="/suppliers/create">Create supplier account</a></li>
```
robyoung: Should this include a link for creating a buyer account?

app/main/views/login.py
```diff
@@ -0,0 +1,355 @@
...
+    else:
+        template_data = main.config['BASE_TEMPLATE_DATA']
+        return render_template("auth/reset-password.html",
+                               email_address=email_address,
+                               form=form,
+                               token=token,
+                               **template_data), 400
+
+
+@main.route('/create-account', methods=["GET"])
```
robyoung: I wonder if this makes more sense as /buyers/create so that it matches /admin/create and /suppliers/create

app/main/views/login.py
```diff
@@ -0,0 +1,355 @@
...
+            **template_data
+        ), 400
+
+
+@main.route('/create-user/<string:encoded_token>', methods=["GET"])
+def create_user(encoded_token):
+    form = CreateUserForm()
+    template_data = main.config['BASE_TEMPLATE_DATA']
+
+    token = decode_invitation_token(encoded_token, buyer=True)
```
robyoung: In the apiclient we have a supplier flag, here we have a buyer flag. I wonder if both of these would make more sense as a role argument.

app/__init__.py
```diff
@@ -53,4 +74,14 @@ def remove_trailing_slash():
             else:
                 return redirect(request.path[:-1], code=301)
 
+    @application.before_request
+    def refresh_session():
+        session.permanent = True
+        session.modified = True
```
pcraig3: What does `session.modified` do?

app/main/views/marketplace.py
```diff
@@ -0,0 +1,66 @@
+# -*- coding: utf-8 -*-
```
pcraig3: How about `app/main/views/GOVUKdigimkt.py`?

app/templates/_proposition_header.html
TheDoubleK: Not yet I don't think - Ralph is working on the design still, but I think the link to "Create buyer account" will come as part of the create-a-brief frontend flow.

app/__init__.py
TheDoubleK: It's a straight copy from the supplier app but I believe that having `session.modified = True` effectively resets the session timer with each page view, so it works as "time since last page view" rather than "time since login".

But I could be wrong about that.

app/main/views/marketplace.py
TheDoubleK: :trollface:

app/templates/auth/create-user.html
```diff
@@ -0,0 +1,84 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Create account – Digital Marketplace{% endblock %}
+
+{% block main_content %}
+
+{% if not token %}
```
pcraig3: Is it possible to arrive at this page without a token?  
It seems like in both places this template is called, the token has already been checked.

app/templates/auth/create-user.html
```diff
@@ -0,0 +1,84 @@
...
+            <div class="column-two-thirds">
+                {{ form.hidden_tag() }}
+
+                {% if form.name.errors %}
+                <div class="validation-wrapper">
+                    {% endif %}
+                    <div class="question">
+                        {{ form.name.label(class="question-heading-with-hint") }}
+                        <p class="hint">
+                            Enter the name to be referred to on the Digital Marketplace
```
pcraig3: Seems kind of a funny way to ask for a name.

app/templates/auth/create_your_account_complete.html
```diff
@@ -0,0 +1,39 @@
+{% extends "_base_page.html" %}
```
pcraig3: `create-your-account-complete.html`?  
_(serious suggestion this time)_

app/__init__.py
pcraig3: Okay, so it doesn't actually care if the session has been modified by anything?

app/templates/auth/create-user.html
TheDoubleK: Again, a straight take from the supplier app and I wondered exactly the same thing.
I thought there might be something I was missing and left it there as a belt-and-braces type approach to error handling (if it never gets called then what's the problem), but now you've said the same thing it makes me more confident in ditching it - I have a feeling it must be a left-over from before the error handling got moved out to `create-buyer-user-error.html`

app/__init__.py
TheDoubleK: No, nothing actually needs to be changed - but it will refresh the session timer.  At least I'm pretty sure that's what it does.

app/main/views/login.py
TheDoubleK: Yep that's better. We don't have a `buyers` set of views at the moment so I'll leave it here for now with the other `login` routes, but once we get buyer dashboard, etc. then it should probably move to a collection of `/buyer` pages.

app/templates/auth/create-user.html
TheDoubleK: Again, a lift from the supplier app to keep it consistent - the copy must have been OK'd at some point, but maybe we should take this opportunity to get it reviewed by @superroz or Sue for both apps.

app/main/views/login.py
TheDoubleK: I've made the change - see pull requests here:
utils - https://github.com/alphagov/digitalmarketplace-utils/pull/235
apiclient - https://github.com/alphagov/digitalmarketplace-apiclient/pull/8

TheDoubleK:
Finally! Travis is happy.
Fingers crossed this is ready to go now.

robyoung:
:+1:

PR #211, opened 2016-01-29
Framework renamed to g-cloud

robyoung:
See: https://github.com/alphagov/digitalmarketplace-api/pull/336

### DIFF
--- a/app/main/suppliers.py
+++ b/app/main/suppliers.py
@@ -63,7 +63,7 @@ def suppliers_list_by_prefix():
     page = process_page(request.args.get('page', default=u"1"))
 
     try:
-        api_result = data_api_client.find_suppliers(api_prefix, page, 'gcloud')
+        api_result = data_api_client.find_suppliers(api_prefix, page, 'g-cloud')
         suppliers = api_result["suppliers"]
         links = api_result["links"]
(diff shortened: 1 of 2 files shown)


app/main/suppliers.py
```diff
@@ -63,7 +63,7 @@ def suppliers_list_by_prefix():
     page = process_page(request.args.get('page', default=u"1"))
 
     try:
-        api_result = data_api_client.find_suppliers(api_prefix, page, 'gcloud')
+        api_result = data_api_client.find_suppliers(api_prefix, page, 'g-cloud')
```
TheDoubleK: Should we be properly paramaterize these supplier listing pages with `<framework>` in the route while we're here?  Or we don't know if we're going to want these pages for DOS at all yet? (From discussion so far it sounds like we will want something that allows people to see the suppliers on DOS).

robyoung: I think this should go in as is. It's fixing a specific bug and we're not absolutely sure we know how the DOS pages are going to work yet.

TheDoubleK: OK fine, agreed.

TheDoubleK:
:+1:

PR #212, opened 2016-02-05
Make generated external urls use https for prod and staging

TheDoubleK:
Email links to create an account should be https but aren't without this in the config.

You can see how it's done in the supplier app here: 
https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/config.py#L112


robyoung:
:+1:

PR #213, opened 2016-02-08
Fail the build if one of the frontend steps fails

allait:
Build script should exit immediately if one of the commands exists
with a non-zero status.


robyoung:
:+1:

PR #214, opened 2016-02-09
Add Makefile

robyoung:
This follows the lead taken in the other apps to move commands into a
Makefile. This allows us to better control the tasks that we want to
run.

This change also fixes the config options used and moves defaults into
config.py


TheDoubleK:
Works :+1:

PR #215, opened 2016-02-11
Add 503 errorhandler

pcraig3:
*Note: not sure adding more tests is necessary*
***

@andrewchong found that the password reset error page wasn't being properly called.
Turns out there was no `503` code errorhandler in the buyer app.

This pull request adds one, as well as the capacity to pass error messages through to our `503.html` template.

You can read the commit message if you want more detail, but it's honestly not very interesting.

[>> Bug on Pivtotal](https://www.pivotaltracker.com/story/show/113540415)

### DIFF
--- a/app/main/errors.py
+++ b/app/main/errors.py
@@ -17,7 +17,7 @@ def page_not_found(e):
 
 
 @main.app_errorhandler(500)
-def exception(e):
+def internal_server_error(e):
     return _render_error_page(500)
 
 
--- a/app/main/errors.py
+++ b/app/main/errors.py
@@ -23,10 +23,10 @@ def exception(e):
 
 @main.app_errorhandler(503)
 def service_unavailable(e):
-    return _render_error_page(503)
+    return _render_error_page(503, e.response)
 
 
-def _render_error_page(status_code):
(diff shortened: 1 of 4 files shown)


app/main/errors.py
```diff
@@ -17,11 +17,19 @@ def page_not_found(e):
...
 @main.app_errorhandler(500)
-def page_not_found(e):
+def exception(e):
     return _render_error_page(500)
 
 
-def _render_error_page(status_code):
+@main.app_errorhandler(503)
+def service_unavailable(e):
+    if e.response:
```
TheDoubleK: I don't think we need this check - `_render_error_page` will handle the case if `error_message=None` anyway. So looks like this method should be able to just `return _render_error_page(503, e.response)` without the conditional.

TheDoubleK: And it might help coveralls be a bit happier too - one less untested branch :)

pcraig3: That's a really good point, Kev. A+

TheDoubleK:
I might prefer `internal_server_error` to `exception` as the 500 handler's method name, but it isn't a deal breaker.  :+1:

PR #218, opened 2016-02-26
Buyer publishes clarification questions and answers

robyoung:
[#112340959](https://www.pivotaltracker.com/story/show/112340959)

## Depends on
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/354
- [x] https://github.com/alphagov/digitalmarketplace-apiclient/pull/16
- [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/201

This adds clarification questions and answering clarification questions into the buyer view of briefs.

## Show publish timestamp on live briefs
Draft:
screen shot 2016-02-26 at 14 08 59

Live:
screen shot 2016-02-26 at 14 10 16

## Show clarification questions on live briefs
Draft:
screen shot 2016-02-26 at 14 11 16

Live:
screen shot 2016-02-26 at 14 12 09

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -268,20 +268,33 @@ def add_clarification_question(framework_slug, lot_slug, brief_id):
     if not clarification_questions_open(brief):
         abort(404)
 
-    if request.method == "GET":
-        return render_template(
-            "buyers/answer_question.html",
-            framework=framework,
-            lot=lot,
-            brief=brief,
-            **dict(buyers.config["BASE_TEMPLATE_DATA"])
-        ), 200
-    else:
-        data_api_client.add_brief_clarification_question(brief_id,
-                                                         request.form.get("question", ""),
-                                                         request.form.get("answer", ""),
-                                                         current_user.email_address)
+    content = content_loader.get_manifest(framework_slug, "clarification_question")
+    section = content.get_section(content.get_next_editable_section_id())
 
-        return redirect(
-            url_for('.view_brief_summary', framework_slug=framework_slug, lot_slug=lot_slug,
-                    brief_id=brief_id) + "#clarification-questions")
+    errors = {}
+
+    if request.method == "POST":
+        try:
+            data_api_client.add_brief_clarification_question(brief_id,
+                                                             request.form.get("question", ""),
+                                                             request.form.get("answer", ""),
+                                                             current_user.email_address)
+
+            return redirect(
+                url_for('.view_brief_summary', framework_slug=framework_slug, lot_slug=lot_slug,
+                        brief_id=brief_id) + "#clarification-questions")
+        except HTTPError as e:
+            if e.status_code != 400:
+                raise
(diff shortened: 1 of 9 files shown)


robyoung:
Coveralls is lying. If you look at it coverage has gone up.

app/buyers/views/buyers.py
```diff
@@ -246,3 +249,54 @@ def delete_a_brief(framework_slug, lot_slug, brief_id):
             url_for('.view_brief_summary', framework_slug=framework_slug, lot_slug=lot_slug,
                     brief_id=brief_id, delete_requested=True)
         )
+
+
+@buyers.route("/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/answer-question",
```
TheDoubleK: Seems like we might not need `framework_slug` and `lot_slug` for this route - `brief_id` is the only thing that really matters for adding clarifications.  But given that all the other routes here follow the same pattern maybe we should stick with it.  I'm not sure what's best... for the supplier responses to a brief we went with just including the `brief_id`.

app/buyers/views/buyers.py
```diff
@@ -246,3 +249,54 @@ def delete_a_brief(framework_slug, lot_slug, brief_id):
...
+    content = content_loader.get_manifest(framework_slug, "clarification_question")
+    section = content.get_section(content.get_next_editable_section_id())
+
+    errors = {}
+    status_code = 200
+
+    if request.method == "POST":
+        try:
+            data_api_client.add_brief_clarification_question(brief_id,
+                                                             request.form.get("question", ""),
```
TheDoubleK: We probably don't want to bother sending empty questions or answers to the API?

TheDoubleK: Although I suppose it is following the same pattern as elsewhere to let the API do all validation, so perhaps this is right.

app/templates/buyers/_base_brief_summary_page.html
```diff
@@ -57,17 +57,38 @@
...
+      {% endcall %}
+    </div>
+
+    {% if brief_data.status == "live": %}
+    <div class="column-one-whole">
+
+      {{ summary.heading("Clarification questions", id="clarification-questions") }}
+      {% call(question) summary.list_table(
+        brief_data.clarificationQuestions,
+        field_headings=[
```
TheDoubleK: Can these headings come from the question content at all?  Or probably not worth it?

TheDoubleK:
Couple of comments to think about maybe, but essentially all looks good. :+1: with or without changes.

app/buyers/views/buyers.py
allait: I think it's nice that the URL structure matches the one we have in service submissions. It makes it obvious where to put framework/lot dashboards if we ever need them and it makes the URLs showing up in logs/analytics a lot more useful.

I'm not sure if it's worth moving supplier responses/questions to a similar structure right now though.

PR #221, opened 2016-03-02
Showing `after_content` block should not be conditional on brief status

TheDoubleK:
Previously the "Delete" and "Publish" buttons were not shown for draft briefs.  Now they will be.


robyoung:
:+1:

PR #223, opened 2016-03-02
Doh don't mutate the base template data

TheDoubleK:
The `get_template_data` method was mutating the base template data.  We've been lucky up until now that no name clashes or unexpected effects showed up.  But the new info page for creating a brief has a `lot` keyword argument, and it was possible to get a name clash if the base data had previously been updated with a `lot` from elsewhere.

Everywhere else we just pass in keyword args directly to the templates, so rather than fix the method to make a copy of the base data before adding keys I've removed it altogether and passed in parameters to the template as we do elsewhere.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -40,11 +40,6 @@ def create_app(config_name):
     application.register_blueprint(main_blueprint)
     application.register_blueprint(buyers_blueprint)
 
-    main_blueprint.config = {
-        'BASE_TEMPLATE_DATA': application.config['BASE_TEMPLATE_DATA'],
-    }
-    buyers_blueprint.config = main_blueprint.config
-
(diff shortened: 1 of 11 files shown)


quis:
You can avoid have to pass ‘base’ data to every template by putting it in your `init_app()` instead. See: https://github.com/alphagov/notifications-admin/blob/master/app/__init__.py#L111-L117

TheDoubleK:
Nice - cheers @quis - I'll give it a go.

TheDoubleK:
That's got it tidied up nicely.

This PR now depends on:
 - [ ] https://github.com/alphagov/digitalmarketplace-utils/pull/246

allait:
:+1:

PR #227, opened 2016-03-08
[#115026147] Add dashboard link to homepage sidebar

tombye:
Currently if suppliers log in and then click the 'Digital Marketplace' link to the home page they cannot return to their dashboard without hacking the URL.

(Described in [story #115026147](https://www.pivotaltracker.com/story/show/115026147).

This adds a link going back to the supplier dashboard to the sidebar. It requires the following changes to be merged in the toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/227

It also contains changes to the sidebar HTML to allow it to contain multiple message modules. After talking with @ralph-hawkins it was decided the sidebar should be an area for messages to suppliers (as the main area is one for those to buyers) so the new HTML reflects this.

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -16,7 +16,6 @@ from app import data_api_client, content_loader
 @main.route('/')
 def index():
     temporary_message = {}
-    logged_in = current_user.is_authenticated()
 
     try:
         frameworks = data_api_client.find_frameworks().get('frameworks')
@@ -43,15 +42,11 @@ def index():
             "framework {} status {}".format(framework.get('slug'), framework.get('status')))
         abort(500)
 
-    template_data = {
-        'frameworks': {framework['slug']: framework for framework in frameworks},
-        'temporary_message': temporary_message
-    }
-    if logged_in:
-        template_data['logged_in'] = True
     return render_template(
         'index.html',
-        **template_data
+        frameworks={framework['slug']: framework for framework in frameworks},
+        temporary_message=temporary_message,
+        logged_in=current_user.is_authenticated()
     )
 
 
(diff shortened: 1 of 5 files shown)


app/main/views/marketplace.py
```diff
@@ -41,10 +43,15 @@ def index():
...
+        'frameworks': {framework['slug']: framework for framework in frameworks},
+        'temporary_message': temporary_message
+    }
+    if logged_in:
+        template_data['logged_in'] = True
     return render_template(
         'index.html',
-        frameworks={framework['slug']: framework for framework in frameworks},
-        temporary_message=temporary_message
+        **template_data
```
pcraig3: Is there any reason why we weren't doing this before?
ie, is this the way we should prefer to be passing variables to the template or is it something we should avoid?

pcraig3:
Looks a bit funny when there's also a temporary framework message.  More specifically, it looks like there should be a more distinct separation between the message and the link.
We don't really have any coming frameworks for right _now_; is this something we look at once we do?

screen shot 2016-03-10 at 12 17 26

[(Full page screenshot)](https://cloud.githubusercontent.com/assets/2454380/13669280/503c55a4-e6ba-11e5-8a2b-778bc782a306.png)

app/main/views/marketplace.py
tombye: I think this approach is wrong actually. We used it before when we used to send a standard dict into each template for asset paths, etc...

I'll amend it to the existing pattern.

tombye:
@pcraig3 I talked to @ralph-hawkins about how the existing DOS messages look with the new column structure and they'll be re-designed when they're needed again in a more standardised format so can look odd for now.

pcraig3:
If @ralph-hawkins is okay with it, then :+1: from me

PR #231, opened 2016-03-15
Add separate CSS for index page browse lists

tombye:
### With old browse list

homepage_existing_browse_list

### With new browse list

homepage_new_browse_list

### With old browse list

g-cloud_index_existing_browse_list

### With new browse list

g-cloud_index_new_browse_list

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -2,7 +2,6 @@
 from __future__ import unicode_literals
 
 from flask import abort, render_template, current_app
-from flask_login import current_user
 
 from dmapiclient import APIError
 from dmutils.content_loader import ContentNotFoundError
@@ -42,15 +41,10 @@ def index():
             "framework {} status {}".format(framework.get('slug'), framework.get('status')))
         abort(500)
 
-    if current_user.is_authenticated():
-        logged_in_user = current_user.role
-    else:
-        logged_in_user = None
     return render_template(
         'index.html',
         frameworks={framework['slug']: framework for framework in frameworks},
-        temporary_message=temporary_message,
-        logged_in_user=logged_in_user
+        temporary_message=temporary_message
     )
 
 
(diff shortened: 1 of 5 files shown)


app/templates/index.html
```diff
@@ -104,7 +104,7 @@ <h2 id="supplier-message-heading">Sell services</h2>
           {% if logged_in %}
             <p>
               <a href="/suppliers" class="top-level-link">
-                <span>View your services and account details</span>
+                View your services and account details
```
TheDoubleK: I think the `{% if logged_in %}` condition here needs to be more specific - it should only show this link/sidebar if logged in as a supplier, not as a buyer.

TheDoubleK: We might want a different link for logged-in buyer users, to take them to their buyer dashboard `/buyers` instead.

tombye: That's a good point. @ralph-hawkins has a design for a link for buyers too so I'll put it in and make the condition more specific.

tombye:
@TheDoubleK I've made some changes in line with your comments:

1. `index.html` now gets sent the `logged_in_user` variable which is the role or `None`
2. the link to the supplier dashboard is now tied to the logged in user being a supplier
3. the link to the buyer dashboard (which was already in the main browse-list) now only appears if a       buyer user is logged in
4. I changed the text in the buyer dashboard link to match that provided by @ralph-hawkins

app/main/views/marketplace.py
```diff
@@ -42,11 +42,15 @@ def index():
             "framework {} status {}".format(framework.get('slug'), framework.get('status')))
         abort(500)
 
+    if current_user.is_authenticated():
+        logged_in_user = current_user.role
```
TheDoubleK: I don't think you need to pass anything into the template at all here, as templates have access to `current_user`.

So, e.g. in _proposition_header.html we have `{% if current_user.email_address %}` and for the new links we can do in the template: `{% if current_user.role == 'supplier' %}` with no need for an additional parameter.

tombye:
@TheDoubleK I've made some changes so `current_user` is used and it's not necessary to mock it in the tests.

TheDoubleK:
:+1:  :shipit:  :+1:

PR #233, opened 2016-03-16
Tweak error page CSS

pcraig3:
Create a new list (container) style with bullets but without extra spacing.  Only visible onscreen change is the bullets.
If we're happy, I'll bring this into the supplier app next.

***
#### Old
screen shot 2016-03-16 at 14 19 33

#### New
screen shot 2016-03-16 at 14 12 59


carolinegreen:
If this is a general pattern and going to be applied to the supplier app as well, it would be better for it be in the frontend toolkit rather than being repeated across apps.

TheDoubleK:
:+1:  and screw what coveralls says.

tombye:
+1 to @carolinegreen's comment. This code is great and should be made available to all the apps.

pcraig3:
While, that *could* be the answer, it's not really what we've been using the frontend toolkit for up to this point. 
To my knowledge, we're only using the frontend toolkit for components: tightly coupled CSS and markup that we pass variables into.  Whereas all we want generalized here is the CSS classes, so is that best done through the toolkit?

Have we done that before?

carolinegreen:
That is a component. We have done this with other lists:

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/0b48e0df5d7844a6c3c33d4df2711fe77be19efb

I don't see what you think the advantage is of duplicating classes?

pcraig3:
Cool!  All ported over to the frontend toolkit.

tombye:
The component is now a part of the front-end toolkit so this looks good to go :+1:

pcraig3:
Thanks for the :+1:, Tom!

PR #235, opened 2016-03-17
Use a single UA code for Buyer App analytics

tombye:
Missed out of this story (which only changed the Supplier App):

https://www.pivotaltracker.com/story/show/106115458


robyoung:
:+1:

PR #237, opened 2016-03-21
Add brief clarification question numbers

robyoung:
[#115944077](https://www.pivotaltracker.com/story/show/115944077)

Numbers are needed in the buyer and public view of the clarification questions so that later questions can easily refer to earlier questions.

The fact that summary tables are done with macros makes it more difficult to just use Jinja2's `loop.counter` variable. The simplest way of getting a counter actually seems to be to pass it in to the template. While this number could be generated in the API it is just for display, ideally it would be generated in the view but that is not easy because of how we use macros.

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -206,6 +206,11 @@ def view_brief_summary(framework_slug, lot_slug, brief_id):
 
             flattened_brief.append(question)
 
+    brief['clarificationQuestions'] = [
+        dict(question, number=index+1)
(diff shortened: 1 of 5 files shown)


allait:
We could add the index to the `call(question, index)` in the summary macro later, but that's quite a bit more work, so this is good :+1:

robyoung:
Yes, I looked at doing that. It would mean a major version bump on the toolkit and changing all the call sites. Something like that is probably the right thing to do ultimately though.

PR #240, opened 2016-03-30
Fix specialist role display on brief page

allait:
### Use value label lookup from dmutils
Removes label lookup code from the brief summary view since the label is now returned by question_summary.value if there's a matching option.

Fixes test specialistRole values to match the new content values and adds a test for label display on the brief page.


robyoung:
:+1:

PR #242, opened 2016-04-04
Buyer can post clarification questions when brief is `live`

TheDoubleK:
This removes the now-unnecessary stubbed-out function that was going to check if clarification questions are open.  That information is included in the API response now, so no need to compute it in the front end.

Also removes the condition that clarificaton period should be open in order for buyers to submit a question/answer pair - buyers should be able to do this the whole time the brief is live.


allait:
👍

PR #244, opened 2016-04-07
[#117143511] Display flashed messages on the view-opportunity page

TheDoubleK:
Part of the fix for this bug: https://www.pivotaltracker.com/story/show/117143511

Rather than 404 when a supplier has already responded to a brief, we will show a flashed message on the brief page that explains they have already submitted and can't do so twice.  The message itself will be flashed through from the supplier app.

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -19,6 +19,18 @@
 
 {% block main_content %}
 
+{% with messages = get_flashed_messages(with_categories=True) %}
+  {% for category, message in messages %}
+    {%
+      with
+      message = message,
+      type = "destructive" if category == "error" else "success"
+    %}
+    {% include "toolkit/notification-banner.html" %}
+    {% endwith %}
+  {% endfor %}
+{% endwith %}
+
 {% if brief.status == 'closed' %}
 <div class="grid-row">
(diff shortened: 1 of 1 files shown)


allait:
I'm not sure where this requirement is coming from, but I'd rather we show a message on the error page itself. I'm not sure if flashed messages between apps rely on the shared secret (which we do for user logins right now, but it could change if we split a user app), but even if not this introduces a new interaction between the apps that's hard to unit test and will require multi-stage deploys if changed.

TheDoubleK:
There isn't an error page to flash on though - trying to submit a second response to a brief shouldn't be a 404.
The problem is that the buyer app's `view-a-brief` page that has the link to respond to the brief links to the supplier app, and it's only once a user is logged-in to the supplier app that we can check whether they have already submitted a response or not.

I spoke to Ralph about this and if they have already responded to a brief then they should stay on the brief overview page but get a message that they can't submit twice. 

So this is doing this existing TODO: https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/briefs.py#L65

TheDoubleK:
Closing - going to deal with this all within supplier app.

PR #245, opened 2016-04-07
Exclude app/content from PEP8 checks

tombye:
As suggested by @allait in https://github.com/alphagov/digitalmarketplace-frameworks/pull/219#issuecomment-206917506


allait:
👍

PR #246, opened 2016-04-11
Don't iterate all (closed) briefs for dashboard

TheDoubleK:
This should fix the problems with preview API running out of credit because the functional test user has 1000s of briefs.

Iterating all closed briefs will soon (now in the case of the functional tests) result in too many API requests per page.

Instead we're going to add the response count to the API response.

### DIFF
--- a/tests/app/views/test_buyers.py
+++ b/tests/app/views/test_buyers.py
@@ -40,37 +40,40 @@ class TestBuyerDashboard(BaseApplicationTest):
             assert live_row[0] == "A live brief"
             assert live_row[1] == "Thursday 04 February 2016"
 
-    def test_closed_brief_response_count(self, data_api_client):
-        with self.app.app_context():
-            self.login_as_buyer()
-            data_api_client.find_briefs.return_value = {
-                "briefs": [
-                    {"status": "closed",
-                     "id": 12,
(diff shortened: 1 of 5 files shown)


tests/app/views/test_buyers.py
```diff
@@ -40,37 +40,40 @@ def test_buyer_dashboard(self, data_api_client):
...
-
-            assert res.status_code == 200
-
-            cell = document.xpath(
-                "//caption[contains(text(), 'Closed requirements')]"
-                "//following-sibling::tbody/tr[1]/td[last()]"
-            )[0]
-
-            assert "1 responses" in cell.text_content()
+    # TODO: reinstate this test when reponse counts are back on the buyer dashboard page
```
allait: Let's mark this as skipped https://pytest.org/latest/skipping.html#marking-a-test-function-to-be-skipped instead of commenting, otherwise it'll probably get lost

TheDoubleK: Nice - hadn't seen that before.

TheDoubleK:
Took me a while to realise we need to update pytest for this, but fixed now.

allait:
👍

TheDoubleK:
npm error in Travis...

PR #247, opened 2016-04-12
Add question and answer session details link to opportunity page

allait:
### Add question and answer session details link to opportunity page
Adds a link to view q+a session (webinar) details to the brief page. Adding this to the _brief_q_and_a template since it seems to be closely related to the clarification questions (one of the main points during the discussion was whether it should be a separate link or the same one as "Ask a question")

#### [#117296739] Update clarification questions closed message

screen shot 2016-04-14 at 15 13 43

### DIFF
--- a/tests/app/views/test_marketplace.py
+++ b/tests/app/views/test_marketplace.py
@@ -346,6 +346,20 @@ class TestBriefPage(BaseApplicationTest):
 
         assert qa_session_link_text == "Log in to view question and answer session details"
 
+    def test_dos_brief_question_and_answer_session_details_hidden_when_questions_closed(self):
+        self.brief['briefs']['clarificationQuestionsAreClosed'] = True
+        brief_id = self.brief['briefs']['id']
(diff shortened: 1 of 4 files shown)


tests/app/views/test_marketplace.py
```diff
@@ -335,6 +335,17 @@ def test_dos_brief_has_at_least_one_section(self):
         assert_equal(contract_length_key[0], 'Contract length')
         assert_equal(contract_length_value[0], '4 weeks')
 
+    def test_dos_brief_has_question_and_answer_session_details_link(self):
```
TheDoubleK: Dare I say we could have a couple more tests... maybe one where the questions deadline has closed and one where there aren't any session details - in each case testing that the section isn't there?

allait: Added tests in b6a6a64

TheDoubleK:
:+1: once Travis manages to NPM install OK.

PR #248, opened 2016-04-12
Add publish page 

carolinegreen:
Add publish page.  Page only displays links to questions with no publish button is required questions have not been completed, otherwise it only shows the publish button. 

Pulled out flattened_brief method and put it into a helper as this is common to several methods.

The link to the publishing page itself is a placeholder for now and will be replaced when the summary page is replaced with an overview page.

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -8,9 +8,9 @@ from flask_login import current_user
 from app import data_api_client
 from .. import buyers, content_loader
 from ...helpers.buyers_helpers import (
-    add_unanswered_counts_to_briefs, all_essentials_are_true, brief_can_be_edited, count_unanswered_questions,
-    counts_for_failed_and_eligible_brief_responses, get_framework_and_lot, get_sorted_responses_for_brief,
-    is_brief_associated_with_user
+    all_essentials_are_true, counts_for_failed_and_eligible_brief_responses,
+    get_framework_and_lot, get_sorted_responses_for_brief, is_brief_associated_with_user, count_unanswered_questions,
+    brief_can_be_edited, add_unanswered_counts_to_briefs, get_flattened_brief
 )
 
 from dmapiclient import HTTPError
@@ -192,14 +192,9 @@ def view_brief_summary(framework_slug, lot_slug, brief_id):
         {'lot': lot['slug']}
     )
     sections = content.summary(brief)
+    flattened_brief = get_flattened_brief(sections)
     unanswered_required, unanswered_optional = count_unanswered_questions(sections)
     delete_requested = True if request.args.get('delete_requested') else False
-
-    flattened_brief = []
-    for section in sections:
-        for question in section.questions:
-            question.section_id = section.id
-            flattened_brief.append(question)
 
     brief['clarificationQuestions'] = [
         dict(question, number=index+1)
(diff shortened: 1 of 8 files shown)


carolinegreen:
This is temporary as this entire page is going to be replaced:
screen shot 2016-04-12 at 17 36 31

carolinegreen:
Page when questions have been answered:
screen shot 2016-04-12 at 17 37 40

carolinegreen:
Page with unanswered questions:
screen shot 2016-04-12 at 17 37 16

app/buyers/views/buyers.py
```diff
@@ -310,6 +305,45 @@ def read(self):
...
+        abort(404)
+    brief_users = brief['users'][0]
+    brief_name = brief_users['name']
+    email_address = brief_users['emailAddress']
+
+    content = content_loader.get_manifest(framework_slug, 'edit_brief').filter(
+        {'lot': lot['slug']}
+    )
+    sections = content.summary(brief)
+    flattened_brief = get_flattened_brief(sections)
```
TheDoubleK: The flattened brief is only required for the GET, so this call could move to the `else` part below.

app/buyers/views/buyers.py
```diff
@@ -310,6 +305,45 @@ def read(self):
...
+
+    unanswered_required, unanswered_optional = count_unanswered_questions(sections)
+    if request.method == 'POST':
+        if (unanswered_required > 0):
+            abort(404)
```
TheDoubleK: A different response code (maybe 400?) seems more appropriate here, maybe with a message explaining the error.

app/buyers/views/buyers.py
```diff
@@ -310,6 +305,45 @@ def read(self):
...
+        data_api_client.update_brief_status(brief_id, 'live', brief_name)
```
TheDoubleK: I think the user should be the currently logged in user. For now this will be the same as the brief user, but in future there may be multiple users per brief - the auditing should capture who was logged in when the button was pressed.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -0,0 +1,64 @@
...
+    {%
+        with items = [
+        {
+            "link": "/",
+            "label": "Digital Marketplace"
+        }
+        ]
+    %}
+    {% include "toolkit/breadcrumb.html" %}
+    {% endwith %}
```
TheDoubleK: We probably want more breadcrumbs on this page - one for burr dashboard and one to the brief grandparent page? Probably best to check with @ralphhawkins.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -0,0 +1,64 @@
...
+<div class='grid-row'>
+  <div class='column-two-thirds large-paragraph'>
+    {% with
+        heading = "Publish your requirements and evaluation criteria",
+        smaller = True,
+        with_breadcrumb = True
+    %}
+        {% include "toolkit/page-heading.html" %}
+    {% endwith %}
+    <p class="padding-bottom-small">All requirements are published on the Digital Marketplace where anyone can see them.</p>
```
TheDoubleK: There is a new secret style `dmspeak` in the toolkit that I think if you put it on the wrapper div will mean we don't need the `padding-bottom-small`s on each paragraph.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -0,0 +1,64 @@
...
+        label = "Publish Requirements"
+        %}
+        {% include "toolkit/button.html" %}
+        {% endwith %}
+      </form>
+    {% endif %}
+  </div>
+</div>
+
+{% endblock %}
```
TheDoubleK: Newline at EoF

tests/app/views/test_buyers.py
```diff
@@ -587,6 +587,77 @@ def test_404_if_framework_status_is_not_live_(self, data_api_client):
 
 
 @mock.patch('app.buyers.views.buyers.data_api_client')
+class TestPublishBrief(BaseApplicationTest):
+    def test_publish_brief(self, data_api_client):
```
TheDoubleK: Would be nice to have some tests on the GET route too - e.g. check publish button is there if brief is complete, isn't there if there are still unanswered questions.

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: There is no brief grandparent page currently.

app/buyers/views/buyers.py
carolinegreen: Could do, this should never actually happen as the button shouldn't appear, so I was just putting in something very generic.

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: Yeah, it being secret really isn't helpful! Thanks for letting me know about it.

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: Got to stop doing that.

tests/app/views/test_buyers.py
carolinegreen: Yes, that is there now.

app/buyers/views/buyers.py
carolinegreen: That was already there, but the variable name wasn't the best, so I've changed that now.

TheDoubleK:
:+1:

PR #243, opened 2016-04-06
[#116130103] Participants start page

TheDoubleK:
Completes this story: https://www.pivotaltracker.com/story/show/116130103

The requirements start page design and content has changed quite a lot, and this brings it up to date with the prototype (e.g., http://dm-prototype.herokuapp.com/participants-start), and also adds an introductory sentence specific to each lot at the top of the page (words by Roz, OK'd by Sarah).

The "Find user research participants" link from the homepage now takes the user to this start page rather than straight into creating the requirements.

This is how the three start pages look now...

SPECIALISTS
===========
screen shot 2016-04-06 at 16 19 48

OUTCOMES
==========
screen shot 2016-04-06 at 16 20 09

PARTICIPANTS
=============
screen shot 2016-04-06 at 16 20 37

PR #249, opened 2016-04-12
Remove padding class

carolinegreen:
Padding class doesn't exist anymore, so it should be removed.


TheDoubleK:
This is already done in #243, so probably best to just get that one in.

PR #250, opened 2016-04-13
Re-adding padding class 

carolinegreen:
Re-adding padding class as dmspeak in its current form does not work foor this page.  Can look at how dmspeak and padding classes should be used later.

### DIFF
--- a/app/templates/buyers/brief_publish_confirmation.html
+++ b/app/templates/buyers/brief_publish_confirmation.html
@@ -26,17 +26,17 @@
     %}
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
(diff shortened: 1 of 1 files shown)


TheDoubleK:
The `dmspeak` class needs to go on the wrapper rather than the individual `p`s, e.g. 
```
<div class="dmspeak">
  <p>Blah...</p>
  <p>Blah...</p>
</div>
```

carolinegreen:
Yes, and then everything is spaced out, so I'll have to change a class that doesn't even have an example.

TheDoubleK:
OK cool :+1:

TheDoubleK:
Merging this as going to release the buyer app now.

PR #252, opened 2016-04-14
[#116129469] Studios catalogue

TheDoubleK:
With: @Wynndow 

Fixes this bug: https://www.pivotaltracker.com/story/show/117521463
And completes this story: https://www.pivotaltracker.com/story/show/116129469

This moves the start pages from the `buyers` blueprint to the `main` blueprint, thus making them accessible without logging in.

The CSV of labs itself is not yet finalised, and should only be put up at the live URL when DOS is live.  There's a holding file there for now, just to make sure the link/download works OK:
https://assets.digitalmarketplace.service.gov.uk/catalogues/digital-outcomes-and-specialists/user-research-studios.csv

Also adds the studios start page, which ~~looks like this~~:
       _See comment below for latest screenshot_

~~The link to the guidance doesn't go anywhere yet - waiting for guides to be live on GOV.UK to put the proper URL in.~~ Proper URL now added from the spreadsheet of guidance links.

### DIFF
--- a/app/main/views/digital_outcomes_and_specialists.py
+++ b/app/main/views/digital_outcomes_and_specialists.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+from app import data_api_client
+
+from flask import render_template
+from ...helpers.buyers_helpers import get_framework_and_lot, count_suppliers_on_lot
+from ...main import main
+
+
+@main.route('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>', methods=['GET'])
+def info_page_for_starting_a_brief(framework_slug, lot_slug):
+    framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client,
+                                           status='live', must_allow_brief=True)
+    print("LOT: {}".format(lot))
+    print("FRAMEWORK: {}".format(framework))
+    return render_template(
+        "buyers/start_brief_info.html",
+        framework=framework,
+        lot=lot,
+        supplier_count=count_suppliers_on_lot(framework, lot)
+    ), 200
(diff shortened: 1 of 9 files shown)


TheDoubleK:
Content is finalised - the page now looks like:
screen shot 2016-04-14 at 17 31 28

app/main/views/digital_outcomes_and_specialists.py
```diff
@@ -0,0 +1,29 @@
...
+from app import data_api_client
+
+from flask import abort, render_template
+from ...helpers.buyers_helpers import get_framework_and_lot
+from ...main import main
+
+
+@main.route('/buyers/frameworks/<framework_slug>/requirements/user-research-studios', methods=['GET'])
+def studios_start_page(framework_slug):
+    framework = data_api_client.get_framework(framework_slug)['frameworks']
```
allait: `user-research-studios` are tied to a specific framework, I think we still need to check that the framework has the lot

app/templates/buyers/studios_start_page.html
```diff
@@ -0,0 +1,99 @@
...
+            <p>Read more about <a href="https://www.gov.uk/guidance/how-to-hire-user-research-labs-on-the-digital-marketplace">how to hire user research labs</a>.</p>
+        </div>
+        
+        <h2>Download the list of labs</h2>
+
+        {%
+        with
+            items = [
+            {
+            "title": "List of labs (???K)",
```
allait: Let's just remove the size for now. We can add it if we have time to do a release once we have the final files.

app/main/views/digital_outcomes_and_specialists.py
TheDoubleK: Yep good point.

allait:
👍

PR #255, opened 2016-04-15
[#115253573] Overview page and separate brief into sections

pcraig3:
This pull request redoes buyer brief interface so that it looks like the prototype <sub>sort of</sub>.
It turned out to be a large 2 point story.

Diff is probably a nightmare.

Things we've done: 
- add grandparent, parent, and child pages
  - one question per page (instead of all of a section's questions per page)
  - section descriptions look good
- add links on the overview page take you to section summaries if they have more than one question or a description, otherwise they link straight to the question page
- handle multiquestions correctly
- add really dope ticks (which show up if you complete all required questions in a section)
- kill a bug where it was possible to manually type in the wrong lot to your brief url and it wouldn't really ever 404 you
  - stop passing so many variables to our templates
  - fix some tests that were unintentionally relying on this

Things we haven't done:
- add the correct clarifications question page (see below)
- there is no longer any interface elements that would let you delete a draft brief :/
- made the overview page styling exactly like Ralph's (something about hanging indents)
- implemented overview state changes (this is [another story](https://www.pivotaltracker.com/story/show/116125701))
- fixed the breadcrumbs
  - they don't ever link back to the buyer dashboard and we never did anything about it
- render multiquestions correctly
  - the descriptions for multiquestions don't show up
  - the values you input don't show up properly on the section summary table (see below)

***

#### brief overview (updated)
screen shot 2016-04-18 at 17 01 30

#### section summary
screen shot 2016-04-15 at 16 32 40

#### (multi)question 
screen shot 2016-04-15 at 15 50 54

#### multiquestions don't show up properly
screen shot 2016-04-15 at 15 51 02

#### clarifications questions page which I know is supposed to be different but I didn't know when I was doing it
screen shot 2016-04-15 at 15 52 04


app/buyers/views/buyers.py
```diff
@@ -92,163 +92,205 @@ def create_new_brief(framework_slug, lot_slug):
...
+def view_brief_overview(framework_slug, lot_slug, brief_id):
+    get_framework_and_lot(framework_slug, lot_slug, data_api_client, must_allow_brief=True)
+    brief = data_api_client.get_brief(brief_id)["briefs"]
 
-    framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client,
-                                           status='live', must_allow_brief=True)
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
+        abort(404)
+
+    content = content_loader.get_manifest(brief['frameworkSlug'], 'edit_brief').filter({'lot': brief['lotSlug']})
```
allait: This ties a bit with the `lot`/`lotSlug` move @TheDoubleK was talking about this morning. Normally, this should be `.filter(brief)`, but it looks like we're trying to move away from `brief['lot']` containing a slug while all questions still use "depends on lot" for lot checks.

We should probably leave this as is right now, but it's worth revisiting later since this limits depends to only work on 'lot', which isn't mentioned anywhere in frameworks repo.

app/buyers/views/buyers.py
```diff
@@ -92,163 +92,205 @@ def create_new_brief(framework_slug, lot_slug):
...
+                brief_id=brief['id']))
 
 
-@buyers.route(
-    '/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/edit/<section_id>',
-    methods=['GET'])
-def edit_brief_submission(framework_slug, lot_slug, brief_id, section_id):
+@buyers.route('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>', methods=['GET'])
+def view_brief_overview(framework_slug, lot_slug, brief_id):
+    get_framework_and_lot(framework_slug, lot_slug, data_api_client, must_allow_brief=True)
```
allait: This probably should still check for 'live' framework status

app/buyers/views/buyers.py
pcraig3: Yeah, I was using `.filter(brief)` and it was working when going through the app, but then a bunch of tests started breaking.  Turns out that the `brief()` returned [from the stubs in the apiclient](https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/api_stubs.py#L33) doesn't have a `lot`, so I just left it as it was.

app/buyers/views/buyers.py
```diff
@@ -80,163 +80,205 @@ def create_new_brief(framework_slug, lot_slug):
...
-                                           status='live', must_allow_brief=True)
+    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id) or not brief_can_be_edited(brief):
+        abort(404)
+
+    content = content_loader.get_manifest(brief['frameworkSlug'], 'edit_brief').filter(
+        {'lot': brief['lotSlug']}
+    )
+    section = content.get_section(section_slug)
+    # https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/views/services.py#L381
+    # if section and (question_id is not None):
```
allait: What's this?

pcraig3: That's what we do with multiquestions in the supplier app: we intercept any multiquestions in the view and turn them into section-like objects.
Here, we're unpacking multiquestions in the template file itself.

It would be nice to keep some of this as non-tacit knowledge, but, yeah, I don't know that this commented out stuff is the way to do it.

app/buyers/views/buyers.py
```diff
@@ -80,163 +80,205 @@ def create_new_brief(framework_slug, lot_slug):
...
         return render_template(
-            "buyers/edit_brief_section.html",
-            framework=framework,
-            data=update_data,
+            "buyers/edit_brief_question.html",
+            brief=brief,
             section=section,
+            question=question,
             errors=errors
         ), 200
```
allait: 400

app/buyers/views/buyers.py
```diff
@@ -80,163 +80,205 @@ def create_new_brief(framework_slug, lot_slug):
...
-        delete_requested=delete_requested
-    ), 200
 
 
 @buyers.route('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses', methods=['GET'])
 def view_brief_responses(framework_slug, lot_slug, brief_id):
-
-    framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client, must_allow_brief=True)
-
+    get_framework_and_lot(framework_slug, lot_slug, data_api_client, must_allow_brief=True)
```
allait: Live framework only.

app/templates/buyers/brief_overview.html
```diff
@@ -0,0 +1,139 @@
...
+                  'text': 'How to assess suppliers'
+                }
+              ]
+            },
+            {
+              'title': 'Award the supplier that best meets your needs',
+              'links': [
+                {
+                  'href': '/',
+                  'text': 'How to award a contract'
```
TheDoubleK: https://www.gov.uk/guidance/how-to-award-a-contract-when-you-buy-services

app/templates/buyers/brief_overview.html
```diff
@@ -0,0 +1,139 @@
...
+                  'text': 'How to shortlist suppliers',
+                }
+              ]
+            },
+            {
+              'title': 'Assess the shortlist',
+              'links': [
+                {
+                  'href': '/',
+                  'text': 'How to assess suppliers'
```
TheDoubleK: https://www.gov.uk/guidance/ways-to-assess-digital-outcomes-and-specialists-suppliers

app/templates/buyers/brief_overview.html
```diff
@@ -0,0 +1,139 @@
...
+                }
+              ]
+            },
+            {
+              'title': 'Choose the shortlist',
+              'description': 'After the application deadline, shortlist the suppliers who applied.',
+              'links': [
+                {
+                  'href': '/',
+                  'text': 'How to shortlist suppliers',
```
TheDoubleK: Will be `https://www.gov.uk/guidance/how-to-shortlist-digital-outcomes-and-specialists-suppliers`

app/templates/buyers/brief_overview.html
```diff
@@ -0,0 +1,139 @@
...
+            {
+              'title': 'Publish requirements',
+              'description': 'Before you can publish your requirements, you must complete:'
+            },
+            {
+              'title': 'Answer supplier questions',
+              'description': 'When you’ve published your requirements, you must answer all supplier questions.',
+              'links': [
+                {
+                  'href': '/',
```
TheDoubleK: Will be `https://www.gov.uk/guidance/how-to-answer-supplier-questions-on-the-digital-marketplace`

TheDoubleK:
Guidance URLs from: https://docs.google.com/spreadsheets/d/16_H7AadKnWG6Gc1sjW-TUj_fxYqhrdITwr2iZgwyEv8/edit?ts=56f3ed55#gid=1943759623

TheDoubleK:
This looks odd with the fist link not capitalised and the following ones are:
screen shot 2016-04-18 at 15 50 05

TheDoubleK:
I don't get a tick for the "describe question and answer session" question once it's completed. Not sure why? Is it because it's optional?  If so, it still feels like it should get a tick somehow...

TheDoubleK:
The link to "Publish requirements" doesn't appear anywhere on the page - it's in section 3 in the prototype.

Also, missing a section heading - section 2 should be "Set how you’ll evaluate specialists", 3. "Publish requirements", etc.

screen shot 2016-04-18 at 16 02 53

allait:
👍

pcraig3:
We talked about the ticks and when they should show up last week, @TheDoubleK.

That particular section only has one question, and it's optional.

So we could write in some logic such that:
 - tick if all required are answered
 - else, if ! required:
  - tick if all optional are answered

But then it's a question of what we want ticks to signify.

As it stands, the tick means
"You have completed all the things you need to in this section in order to be able to publish a brief"

If we did it on the optional question, it would mean
"You get a tick for filling out some contingent grouping of questions in this section",
or even
"You get a tick because there is nothing left to do in this section"

It was decided to let it never be ticked because it would be consistent with the required questions thing, but the fact that you can never get a tick for that section even after you've answered it (and you have ticks for every other section) seems a bit counterintuitive.

I don't think anyone's thrilled, but I do think it's a reasonable approach.

pcraig3:
💣  💣  💣

PR #256, opened 2016-04-18
[#117542773] Remove `yet` from home page

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/117542773

I've checked the functional tests and this change shouldn't break them as far as I can see.


pcraig3:
👍

PR #259, opened 2016-04-18
Temporary message subheadings are safe

TheDoubleK:
Sometimes we put HTML markup in our messages - it shouldn't be escaped.

### Before:
screen shot 2016-04-18 at 16 18 58

### After:
screen shot 2016-04-18 at 16 19 19

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -101,7 +101,7 @@
         {%
           with
             heading = temporary_message.heading,
-            subheading = temporary_message.subheading,
+            subheading = temporary_message.subheading|safe,
             messages = temporary_message.messages,
             message = temporary_message.message
         %}


pcraig3:
Nice try going down the safe route, Kev.

1. [markdown filter extended to support abbreviations](https://github.com/alphagov/digitalmarketplace-utils/pull/257)
2. [frameworks updated to new abbreviation syntax](https://github.com/alphagov/digitalmarketplace-frameworks/pull/228)
3. [toolkit updated to use markdown filter for framework notice subheadings](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/237)

TheDoubleK:
Closing because three pull requests and reliance on new external extension libraries is better than a five character change in a template file.

pcraig3:
Ha!

It's not an external extension though.

screen shot 2016-04-19 at 15 47 30

PR #261, opened 2016-04-19
Overview page styling amends

tombye:
The overview page was styled so its numbers and ticks were positioned outside the main list.

This brings them inside the list to match the prototype.

### Before

overview_old

### After

overview_new


pcraig3:
[Prototype also has black horizontal lines](http://dm-prototype.herokuapp.com/outcome-overview).

I like grey better personally, but is this a decision we've actually made or a thing we never fixed?

pcraig3:
I think the rest of it looks good though!

tombye:
@pcraig3 I spoke to @ralph-hawkins about this and he said he's ok for us to keep using the `$border-colour` Sass variable (as we do now) for this. We will add a DM-level border variable for the darker look at a later point which can then be brought into this pattern but it'll still be done at the DM toolkit level.

pcraig3:
👍

pcraig3:
thumbs_up

PR #262, opened 2016-04-19
Sort out the breadcrumbs

pcraig3:
Breadcrumbs have always been fine for navigating around a brief, but
there's never been a link back to the buyer dashboard (aka, "Your
requirements") in any of the breadcrumbs.
Added them in where they were missing/necessary.

Talked to @ralph-hawkins about them, so they should be okay.

Did a cool thing around only doing breadcrumbs back to section summary pages if `section.has_summary_page` is `True` and added tests for it.

Only pages I didn't update the breadcrumbs on are those clarification question pages because they have to be updated anyway.

***

### brief overview
screen shot 2016-04-19 at 12 40 41

***

### section summary 
screen shot 2016-04-19 at 12 41 33

***

### question page (with section summary)
screen shot 2016-04-19 at 12 41 50

***

### question page (w/o section summary)
screen shot 2016-04-19 at 12 41 12

### DIFF
--- a/app/templates/buyers/clarification_questions.html
+++ b/app/templates/buyers/clarification_questions.html
@@ -10,8 +10,8 @@
         "label": "Digital Marketplace"
       },
       {
-        "link": url_for(".view_brief_overview", framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id),
-        "label": brief.title
+        "link": url_for(".buyer_dashboard"),
+        "label": "Your account"
       }
     ]
   %}
@@ -60,7 +60,7 @@
     </div>
   </div>
 
-  <a href="{{ url_for(".add_clarification_question", framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Answer a clarification question</a>
+  <a href="{{ url_for('.add_clarification_question', framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id) }}">Answer a clarification question</a>
 
   {%
     with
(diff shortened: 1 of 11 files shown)


app/templates/buyers/clarification_questions.html
```diff
@@ -3,15 +3,16 @@
 {% block page_title %}Clarification questions – Digital Marketplace{% endblock %}
 
 {% block breadcrumb %}
+  {# TODO: this is the wrong label, probably #}
```
TheDoubleK: It's the same label as all the other links back to `.buyer_dashboard` so it's probably right?  Would be weird to use different words for the same destination page.

TheDoubleK: Either way, we should remove the comment here.

app/templates/buyers/create_brief_question.html
```diff
@@ -2,12 +2,24 @@
...
+
 {% block breadcrumb %}
   {%
     with items = [
       {
         "link": "/",
         "label": "Digital Marketplace"
+      },
+      {
+        "link": url_for("main.info_page_for_starting_a_brief", framework_slug=framework.slug, lot_slug=lot.slug),
```
TheDoubleK: This seems like an odd page to go back to, but guess it makes sense as it's the only route to this page.

TheDoubleK:
Other than the comment that needs to be removed this looks good :+1: (at least, better than what we currently have ;)

app/templates/buyers/clarification_questions.html
pcraig3: Yeah, whoops.  Forgot to remove that.

app/templates/buyers/create_brief_question.html
pcraig3: It's already been RalphApproved™, Kev.

PR #263, opened 2016-04-19
Update brief questions content

allait:
### Use list question number_of_items for list form macro
Changes the number of list fields based on the question content data.


TheDoubleK:
:+1: if Travis doesn't choke.

PR #266, opened 2016-04-20
Pc supplement steps with states

allait:
### Set allowed statuses for brief overview page steps
Hides the step description based on step.allowed_statuses. Shows 'Done' instead of the step description for live briefs on draft brief steps.

allowed_statuses defaults to all possible statuses, so when not set the link/step will be always displayed (eg for guidance links).

### Split brief overview page step descriptions for each brief status
Sets a separate step descriptions for each available brief status. This simplifies the 'Done' description logic and allows setting different descriptions for any brief state, but requires some duplication for descriptions that don't change.

# Draft brief

screen shot 2016-04-20 at 13 38 11

# Live brief

screen shot 2016-04-20 at 13 38 23

# Closed brief

screen shot 2016-04-20 at 13 38 33


tombye:
👍

PR #267, opened 2016-04-20
Add delete button to brief overview

allait:
#### Add UI to allow deletion of draft briefs
Replace the delete button previously there with a link to the same page so users can confirm through the banner form that is displayed then.


tombye:
👍

PR #272, opened 2016-04-20
[#117813351] Render questions (with descriptions) in the buyer app like they are in the prototype

pcraig3:
## Question descriptions look amaze now

 Question descriptions look amaze now

👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 

[>> Story in Pivotal](https://www.pivotaltracker.com/story/show/117813351)

***

screen shot 2016-04-20 at 18 31 35

***

screen shot 2016-04-20 at 18 34 16

### DIFF
--- a/app/templates/macros/toolkit_forms.html
+++ b/app/templates/macros/toolkit_forms.html
@@ -2,7 +2,9 @@
   {%
     with
     question=question_content.question|markdown,
-    hint=(question_content.hint or '')|markdown,
+    question_advice=question_content.question_advice,
+    hint=question_content.hint,
+    optional=question_content.optional,
     name=question_content.id,
     value=data[question_content.id],
     error=errors.get(question_content.id)['message']
@@ -15,7 +17,9 @@
   {%
     with
     question=question_content.question|markdown,
-    hint=(question_content.hint or '')|markdown,
+    question_advice=question_content.question_advice,
+    hint=question_content.hint,
+    optional=question_content.optional,
     name=question_content.id,
     value=data[question_content.id],
     large=True,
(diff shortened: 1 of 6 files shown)


app/templates/macros/toolkit_forms.html
```diff
@@ -2,7 +2,9 @@
   {%
     with
     question=question_content.question|markdown,
-    hint=(question_content.hint or '')|markdown,
+    question_advice=question_content.question_advice,
+    hint=question_content.hint,
```
TheDoubleK: Are we sure about losing the `|markdown` from `hint`?
I'm pretty sure that some of the hints have markdown-style links in them. Perhaps only on the supplier-app questions, I'm not sure, but even so it might be nice to have consistency of treatment of the content keys across apps - we don't want to have to remember that if we're doing hints that are rendered by the supplier app we can use markdown but if in the buyer app we can't.

TheDoubleK:
Looking very smart - nice one.  That's some old content you're working with though, with the "15 points" stuff and all that.

app/templates/macros/toolkit_forms.html
pcraig3: Latest toolkit update adds a bunch of styling to the `question_advice` and `hint` blocks, and then I also added the markdown filter to them because otherwise all that beautiful styling doesn't get applied (ie, we defined all these rules for `ul`s and `p`s, but without the markdown filter, all that would get rendered is a string).  So it seemed like a toolkit responsibility.

ie, https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/239/files#diff-1de52a43ce5f6aa6acf2d20f48f7c74e

Does that seem alright?

TheDoubleK:
Hmm, yeah that should be OK I think - although I wonder about supplier app stuff getting markdowned twice if it pulls in the new toolkit.

We can fix that later though..

:+1:

pcraig3:
Yeah, I've tried the double markdown thing and it seems fine.

TheDoubleK:
Nice work - let's merge this now.

PR #275, opened 2016-04-21
Fix dateformat being converted to local time

allait:
Update dmutils to 19.7.2 to remove localization when displaying dates.


pcraig3:
👍

PR #277, opened 2016-04-21
Clarification page update

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/117811455

Now depends on:
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/241

The content of the page is all in the updated frameworks repository. This updates the page to display the description and question name properly.


pcraig3:
I *think* that by making the clarifications question into a multiquestion, you could just use the regular edit question template, but other than losing one template file I'm not sure what the utility of that would be.

pcraig3:
screen shot 2016-04-21 at 16 05 32

Not as good as Tom's but still worth a 💥

pcraig3:
👍  after a rebase.

PR #278, opened 2016-04-21
reworded link from 'View a list of suppliers'

ralph-hawkins:
to 'Download list of suppliers'


TheDoubleK:
Makes sense :+1: when/if Travis goes green.

PR #279, opened 2016-04-21
Add publishing brief dates

carolinegreen:
Add publishing brief dates and new summary table using current version of frontend toolkit.

Update to use dmutils with short date format without localised time format.

screen shot 2016-04-20 at 18 48 03

screen shot 2016-04-20 at 18 48 12

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -10,7 +10,7 @@ from .. import buyers, content_loader
 from ...helpers.buyers_helpers import (
     all_essentials_are_true, counts_for_failed_and_eligible_brief_responses,
     get_framework_and_lot, get_sorted_responses_for_brief, count_unanswered_questions,
-    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct
+    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct, get_publishing_dates
 )
 
 from dmapiclient import HTTPError
@@ -349,6 +349,13 @@ def publish_brief(framework_slug, lot_slug, brief_id):
     brief_user_name = brief_users['name']
 
     sections = content.summary(brief)
+    question_and_answers = {}
+    question_and_answers_content = sections.get_question('questionAndAnswerSessionDetails')
+    question_and_answers['id'] = question_and_answers_content['id']
+
+    for section in sections:
+        if section.get_question('questionAndAnswerSessionDetails') == question_and_answers_content:
+            question_and_answers['slug'] = section['id']
 
     unanswered_required, unanswered_optional = count_unanswered_questions(sections)
     if request.method == 'POST':
@@ -360,13 +367,16 @@ def publish_brief(framework_slug, lot_slug, brief_id):
                     brief_id=brief['id']))
     else:
         email_address = brief_users['emailAddress']
+        dates = get_publishing_dates()
 
         return render_template(
             "buyers/brief_publish_confirmation.html",
             email_address=email_address,
+            question_and_answers=question_and_answers,
             unanswered_required=unanswered_required,
             sections=sections,
-            brief=brief
+            brief=brief,
+            dates=dates
         ), 200
 
 
--- a/app/helpers/buyers_helpers.py
+++ b/app/helpers/buyers_helpers.py
@@ -1,4 +1,5 @@
 from flask import abort
+from datetime import datetime, timedelta
 
 
 def get_framework_and_lot(framework_slug, lot_slug, data_api_client, status=None, must_allow_brief=False):
@@ -82,3 +83,19 @@ def get_sorted_responses_for_brief(brief, data_api_client):
         )
     else:
         return brief_responses
+
+
+def get_publishing_dates():
+    application_open_days = 14
+    questions_open_days = application_open_days - 7
+    answers_open_days = application_open_days - 1
+
+    dates = {}
+
+    dates['today'] = datetime.utcnow().replace(hour=23, minute=59, second=59, microsecond=0)
+    dates['questions_close'] = dates['today'] + timedelta(days=questions_open_days)
+    dates['answers_close'] = dates['today'] + timedelta(days=answers_open_days)
+    dates['closing_date'] = dates['today'] + timedelta(days=application_open_days)
+    dates['application_open_weeks'] = application_open_days//7
+    dates['closing_time'] = '{d:%I:%M %p}'.format(d=dates['closing_date']).lower()
+    return dates
--- a/app/templates/buyers/brief_publish_confirmation.html
+++ b/app/templates/buyers/brief_publish_confirmation.html
@@ -35,12 +35,10 @@
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
     <p class="padding-bottom-small">All requirements are published on the Digital Marketplace where anyone can see them.</p>
+    <p class="padding-bottom-small">All requirements are open for {{ dates.application_open_weeks }} weeks.</p>
+    <p class="padding-bottom-small">If you publish your requirements today ({{ dates.today | shortdateformat }}), suppliers will be able to apply until {{ dates.closing_time }}, {{ dates.closing_date | shortdateformat}}.</p>
 
-    <p><strong>Suppliers will be able to apply for:</strong></p>
-    <p class="padding-bottom-small">2 weeks ending at midnight.</p>
-
-    <p><strong>Supplier questions will be sent to:</strong></p>
-    <p class="padding-bottom-small">{{ email_address }}</p>
+    <p class="padding-bottom-small"><strong>Supplier questions will be sent to:</strong> {{ email_address }}</p>
     <p class="padding-bottom-small">Ensure this email address will be monitored. If you’re away while suppliers can still ask questions, you should make sure your emails are forwarded to a colleague.</p>
 
     {% if unanswered_required > 0 %}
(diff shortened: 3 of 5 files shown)


app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -57,6 +56,49 @@
...
+        {% call summary.row(no_border=True) %}
+          {{ summary.field_heading_date("Today", scope='row') }}
+          {{ summary.text("Suppliers can apply and ask questions about your requirements.") }}
+        {% endcall %}
+        {% call summary.row(bold_border=True) %}
+          {{ summary.field_heading_date("Before {}".format(dates.questions_close), rowspan='3', scope='row') }}
+          {{ summary.text_bold("Details of your question and answer session") }}
+        {% endcall %}
+        {% call summary.row(no_border=True) %}
+          {{ summary.text("Question and answer session information.") }}
```
allait: This should be the actual session information

app/helpers/buyers_helpers.py
```diff
@@ -82,3 +84,23 @@ def get_sorted_responses_for_brief(brief, data_api_client):
...
+
+
+def get_publishing_dates():
+    application_open_days = 14
+    questions_open_days = application_open_days - 7
+    answers_open_days = application_open_days - 1
+
+    dates = {}
+
+    dates['today'] = datetime.now().replace(hour=23, minute=59, second=59, microsecond=0)
```
allait: Should be `.utcnow()` instead of now, to avoid any potential timezone issues

app/helpers/buyers_helpers.py
```diff
@@ -82,3 +84,23 @@ def get_sorted_responses_for_brief(brief, data_api_client):
...
+    dates['questions_close'] = dates['today'] + timedelta(days=questions_open_days)
+    dates['answers_close'] = dates['today'] + timedelta(days=answers_open_days)
+    closing_date = dates['today'] + timedelta(days=application_open_days)
+    dates['closing_date'] = closing_date
+
+    for date in dates:
+        dates[date] = shortdateformat(dates[date])
```
allait: Usually, we'd apply this as a Jinja filter directly in the template, eg `dates.answers_close | shortdateformat`, but since this is temporary I'm ok with keeping this for now

app/helpers/buyers_helpers.py
carolinegreen: Yes

app/helpers/buyers_helpers.py
```diff
@@ -82,3 +84,23 @@ def get_sorted_responses_for_brief(brief, data_api_client):
...
+    dates['application_open_weeks'] = application_open_days/7
```
TheDoubleK: This will be fine at the moment because we have a multiple of seven for the number of days things are open, but having integer division in our code makes me nervous. eg 20/7 will still give "2 weeks" for this.

app/helpers/buyers_helpers.py
```diff
@@ -82,3 +84,23 @@ def get_sorted_responses_for_brief(brief, data_api_client):
...
+
+    dates = {}
+
+    dates['today'] = datetime.now().replace(hour=23, minute=59, second=59, microsecond=0)
+    dates['questions_close'] = dates['today'] + timedelta(days=questions_open_days)
+    dates['answers_close'] = dates['today'] + timedelta(days=answers_open_days)
+    closing_date = dates['today'] + timedelta(days=application_open_days)
+    dates['closing_date'] = closing_date
+
+    for date in dates:
```
TheDoubleK: This will work, but something like
`dates.update({date : shortdateformat(dates[date]) for date in dates.iterkeys()})`
might be more "pythonic".

Having typed it though, I'm not sure I like it.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -35,9 +35,8 @@
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
     <p class="padding-bottom-small">All requirements are published on the Digital Marketplace where anyone can see them.</p>
-
-    <p><strong>Suppliers will be able to apply for:</strong></p>
-    <p class="padding-bottom-small">2 weeks ending at midnight.</p>
+    <p class="padding-bottom-small">All requirements are open for {{ dates.application_open_weeks }} weeks.</p>
+    <p class="padding-bottom-small">If you publish your requirements today ({{ dates.today }}), suppliers will be able to apply until {{ dates.closing_time }}, {{ dates.closing_date }}.</p>
```
TheDoubleK: It might be nice to stick to our `datetimeformat` from utils rather than making an ad hoc display format.  I don't see why we'd want a different format here than other places where we show dates to users. (Maybe ask @ralph-hawkins if it really must be `{{ dates.closing_time }}, {{ dates.closing_date }}`?)

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -57,6 +56,49 @@
...
+        {% call summary.row(bold_border=True) %}
+          {{ summary.field_heading_date(dates.questions_close, scope='row') }}
+          {{ summary.text("Suppliers can no longer ask questions.") }}
+        {% endcall %}
+        {% call summary.row(bold_border=True) %}
+          {{ summary.field_heading_date(dates.answers_close, rowspan='2', scope='row') }}
+          {{ summary.text("You must have published answers to all suppliers’ questions.") }}
+        {% endcall %}
+        {% call summary.row(no_border=True) %}
+          {{ summary.text("If dates.answers_close is a Saturday or Sunday, you need to have published all questions and answers by the Friday before.", border=True) }}
```
TheDoubleK: `dates.answers_close` is just part of the string here - needs to be pulled out into a `.format()`

app/templates/buyers/brief_publish_confirmation.html
allait: It's the time localisation again. Since we removed it for dates, we can run the date through the display filter but we can't do the same for time since it'll be transformed into 00:59. And the same will happen if we run it through any of the date and time filters. That's why we decided to display the time like this for now.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -35,9 +35,8 @@
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
     <p class="padding-bottom-small">All requirements are published on the Digital Marketplace where anyone can see them.</p>
-
-    <p><strong>Suppliers will be able to apply for:</strong></p>
-    <p class="padding-bottom-small">2 weeks ending at midnight.</p>
+    <p class="padding-bottom-small">All requirements are open for {{ dates.application_open_weeks }} weeks.</p>
+    <p class="padding-bottom-small">If you publish your requirements today ({{ dates.today }}), suppliers will be able to apply until {{ dates.closing_time }}, {{ dates.closing_date }}.</p>
 
     <p><strong>Supplier questions will be sent to:</strong></p>
```
tombye: Can we use the `.dmspeak` styles for this block? The `padding-bottom-small` classes came across from the prototype so we should remove them in app code when possible.

Also, I'd argue `Supplier questions will be sent to:` should be in the same paragraph as the email address so there's no break in the sentence between the two.

app/helpers/buyers_helpers.py
carolinegreen: That was the plan, as we always want it to be full weeks currently and having it handle days was outside the scope.  I could put in a test to fail if it's not a multiple of 7 though.

app/helpers/buyers_helpers.py
carolinegreen: What is the function for utils normally used for, out of curiosity?

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: Did realise that was in this story too!  Will put this in.

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: I wasn't putting in dmspeak at currently, as it still hasn't really got enough of an explanation to make sense as a class or examples in the toolkit.  It will also have to be changed to work for this example.  I wanted to add this later on when we weren't stuck for time.

I thought the email being on a separate line was a design decision, but I agree that it does seem weird and will change it.

app/templates/buyers/brief_publish_confirmation.html
carolinegreen: Yes, missed that.

app/helpers/buyers_helpers.py
allait: The utils function is the one registered as Jinja filter, so it's exactly the same. The reason it's better to use it in a template is that `datetime` object is generally more useful than a string representation, so you usually try to convert it for display at the last possible moment.

app/helpers/buyers_helpers.py
carolinegreen: Changing it to // as well.

app/helpers/buyers_helpers.py
carolinegreen: Right, thanks, will change it.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -57,6 +56,49 @@
...
+        {% call summary.row(no_border=True) %}
+          {{ summary.field_heading_date("Today", scope='row') }}
+          {{ summary.text("Suppliers can apply and ask questions about your requirements.") }}
+        {% endcall %}
+        {% call summary.row(bold_border=True) %}
+          {{ summary.field_heading_date("Before {}".format(dates.questions_close | shortdateformat), rowspan='3', scope='row') }}
+          {{ summary.text_bold("Details of your question and answer session") }}
+        {% endcall %}
+        {% call summary.row(no_border=True) %}
+          {{ summary.text("Question and answer session information.") }}
```
allait: The previous comment got hidden, but this still needs to be replaced with the actual brief data

carolinegreen: I know, it's not up yet.

carolinegreen:
Question and answers session information is now in as well.

carolinegreen:
screen shot 2016-04-22 at 17 24 05

app/buyers/views/buyers.py
```diff
@@ -347,8 +347,16 @@ def publish_brief(framework_slug, lot_slug, brief_id):
     content = content_loader.get_manifest(brief['frameworkSlug'], 'edit_brief').filter({'lot': brief['lotSlug']})
     brief_users = brief['users'][0]
     brief_user_name = brief_users['name']
+    print('>>>>>brief', brief)
```
allait: ☝️

allait:
Looks good 👍

PR #280, opened 2016-04-21
removed use of 'clarification questions'

ralph-hawkins:
changed most of the uses 'clarification' to 'supplier' except where the table where it now says 'No questions or answers have been published'


allait:
👍

PR #285, opened 2016-04-22
Show the correct application deadline for closed briefs

allait:
'This opportunity is closed' banner showed the supplier questions
deadline instead of the application deadline.

Also removes the use of `| safe` on the banner message, since it
shouldn't contain any HTML entities.


TheDoubleK:
:+1:

tombye:
👍

PR #286, opened 2016-04-22
Breadcrumbs on brief page link back to catalogue

pcraig3:
Add extra breadcrumb to a brief page (pointing back to the catalogue of opportunities).

Didn't add tests because there's nothing weird going on here: no url string formatting or additional logic.

***

screen shot 2016-04-22 at 14 54 30


allait:
👍

PR #269, opened 2016-04-20
[#118019637] Bugfix for brief responses with zero nice-to-haves

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/118019637

Currently the CSV generation for downloading responses to briefs always tries to sort them in descending order of the number of nice-to-have requirements.

But if the brief has NO nice-to-haves the call to sort causes an error. (e.g., see excerpt from AWS logs below).

This adds a check that niceToHaveRequirements exists before attempting to sort.  If it doesn't then we just return them in the order they come back from the API.

```
{
    "name": "app",
    "levelname": "ERROR",
    "message": "Exception on /buyers/frameworks/digital-outcomes-and-specialists/requirements/user-research-participants/48/responses/download [GET]",
    "pathname": "/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py",
    "lineno": 1423,
    "exc_info": "Traceback (most recent call last):\n  File \"/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py\", line 1817, in wsgi_app\n    response = self.full_dispatch_request()\n  File \"/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py\", line 1477, in full_dispatch_request\n    rv = self.handle_user_exception(e)\n  File \"/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py\", line 1381, in handle_user_exception\n    reraise(exc_type, exc_value, tb)\n  File \"/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py\", line 1475, in full_dispatch_request\n    rv = self.dispatch_request()\n  File \"/opt/python/run/venv/lib/python2.7/site-packages/flask/app.py\", line 1461, in dispatch_request\n    return self.view_functions[rule.endpoint](**req.view_args)\n  File \"/opt/python/current/app/app/buyers/views/buyers.py\", line 237, in download_brief_responses\n    sorted_brief_responses = get_sorted_responses_for_brief(brief_id, data_api_client)\n  File \"/opt/python/current/app/app/helpers/buyers_helpers.py\", line 83, in get_sorted_responses_for_brief\n    reverse=True\n  File \"/opt/python/current/app/app/helpers/buyers_helpers.py\", line 82, in <lambda>\n    key=lambda k: len([nice for nice in k['niceToHaveRequirements'] if nice is True]),\nKeyError: 'niceToHaveRequirements'",
[...]

PR #287, opened 2016-04-22
Fix download of responses if no nice-to-have requirements

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/118019637

I thought I'd fixed it here: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/269

But it turns out that I missed a bit and that was only half of the fix.  The change here prevents this error:
```
File \"/opt/python/current/app/app/buyers/views/buyers.py\", line 295, in download_brief_responses
    column_headings.extend(brief[question.id])
KeyError: 'niceToHaveRequirements'"
```

Properly tested locally this time and this really does fix the bug.


allait:
👍

PR #288, opened 2016-04-22
[#118169341] Fix empty field data on supplier q and a validation error

allait:
Keeps the submitted form data when rendering the page with validation error so that the submitted value is not lost.

We also don't need `.get` after using `section.get_data` since the keys for text fields will always be present (possibly with `None` values that API needs to handle).


pcraig3:
Legit 👍

PR #289, opened 2016-04-22
Bump frameworks to 0.30.4

allait:
Fixes the participant accessibility needs question id typo and the
ordered list formatting in specialistWork question example.


TheDoubleK:
:+1:

PR #291, opened 2016-04-25
Update frameworks to 0.30.6

allait:
Adds led-in text for requirements questions and early market engagement
question for all lots.


TheDoubleK:
:+1:

PR #293, opened 2016-04-25
Reword homepage opportunities link

tombye:
We’ve decided that the 'view opportunities' page would be more helpful if the page title said 'Digital Outcomes and Specialists opportunities' because suppliers think about lots'. The link to this page should reflect this.

https://www.pivotaltracker.com/story/show/118280489

### Old version

sidebar_link_old

### New version

sidebar_link_new


allait:
👍

PR #295, opened 2016-04-25
Hide question and answers, and minor changes.

carolinegreen:
- Hide question and answer session details if these have not been filled in, as this section is optional.
- Add return to overview link at bottom of the page
- Change copy to reflect what the suppliers can do rather than what they can't.

Overview link:
screen shot 2016-04-25 at 17 12 56

With question and answer details, when these are available:
screen shot 2016-04-25 at 17 13 33

Without question and answer details, when this section has not been filled in:
screen shot 2016-04-25 at 17 14 58


allait:
👍

PR #297, opened 2016-04-26
Move catalogues to communications bucket

allait:
Changes the catalogues CSV URLs to match communications bucket.
Only the admin app uploads files to the communications bucket, so it feels more appropriate to keep catalogue files there.


TheDoubleK:
:+1:

TheDoubleK:
Sorry about that. This is better.

PR #298, opened 2016-04-27
Change button text on publish a supplier Q&A page

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/118417371

The Q&A page now re-uses the `edit_brief_question` template, which has "Save and continue" as the button text.
By passing in `button_label` to the template this changes the text to the required "Publish question and answer".

The button now looks like this:
screen shot 2016-04-27 at 08 25 47


pcraig3:
👍

PR #300, opened 2016-04-27
Fix return to overview link on publish page

pcraig3:
Now using the macro instead of just a link.
Means we have the right spacing.

[>> Bug on Pivotal](https://www.pivotaltracker.com/story/show/118406997)

***

### not nice
screen shot 2016-04-27 at 17 01 51

### nice 
screen shot 2016-04-27 at 17 01 20


TheDoubleK:
:+1:

PR #302, opened 2016-04-28
opportunities show summary

ralph-hawkins:
Each opportunity on /digital-outcomes-and-specialists/opportunities needs to show the 'summary' for all lots. Was 'backgroundInfromation' answer, which is specific to Outcomes, so nothing was showing for either specialists or participants opportunities.


TheDoubleK:
:+1:

PR #304, opened 2016-05-03
Add temporary messages for G-Cloud-8

carolinegreen:
Add temporary messages for G-Cloud-8

Move shared placeholders above frontend toolkit styles to prevent extensions of shared placeholders from being overwritten.

Also add new verions of frontend toolkit, with temporary message variant and digital marketplace framework with new content.

Remove h3 class from suppliers scss as this is obsolete.

Remove framework notice option as this has been superceded by the temporary messages pattern.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -20,45 +20,47 @@
 <div class="index-page grid-row">
   <div class="column-two-thirds">
     <h2 class="marketplace-homepage-subheading">Find technology or people for digital projects in the public sector</h2>
+    {% set buyer_dashboard = [] %}
+    {% set dss_items = [] %}
+    {% set dos_items = [] %}
+    {% set g_cloud_subtext = '' %}
+    {% set crown_hosting_subtext = '' %}
+
     {% if dos_is_live %}
       {% set
         dos_items = [
           {
-            "link": url_for("main.info_page_for_starting_a_brief", framework_slug='digital-outcomes-and-specialists', lot_slug='digital-specialists'),
+            "link": url_for("main.info_page_for_starting_a_brief", framework_slug=dos_is_live.slug, lot_slug='digital-specialists'),
             "title": "Find an individual specialist",
             "body": "eg a developer or user researcher",
           },
           {
-            "link": url_for("main.info_page_for_starting_a_brief", framework_slug='digital-outcomes-and-specialists', lot_slug='digital-outcomes'),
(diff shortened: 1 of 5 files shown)


app/templates/index.html
```diff
@@ -92,33 +92,23 @@ <h2 class="marketplace-homepage-subheading">Find technology or people for digita
...
-            message = temporary_message.message
-        %}
-          {% include "toolkit/{}.html".format(temporary_message.template_name) %}
-        {% endwith %}
-      {% endif %}
-        
+
       {% if dos_is_live %}
+
+      {% set opportunities %}
```
allait: What are the `set` blocks for?

carolinegreen: It was too make things clearer instead of alternating between two sets of ifs for the same thing.

carolinegreen: It was to make things clearer, instead of repeating ifs.  They had to be broken up due to the way the messages are laid out in the template but it didn't read well.

allait: I'm still not sure why this is better. For example this `opportunities` block is just used a few lines below with `| safe`, why not put it there directly?

carolinegreen: Yes, okay I think it was for consistency but I can remove it.

app/templates/index.html
```diff
@@ -92,33 +92,21 @@ <h2 class="marketplace-homepage-subheading">Find technology or people for digita
...
           <p>
             <a href="/digital-outcomes-and-specialists/opportunities" class="top-level-link">
-              Digital Outcomes and Specialists opportunities
+              View Digital Outcomes and Specialists opportunities
             </a>
           </p>
         </div>
-        
-        {% if current_user.role != 'supplier' %}
+      {% set selling %}
```
allait: I think we should do the same thing here, inline the block on line 137 by joining the if conditions with `and`.

carolinegreen: I'm not completely sure what you mean, unless you are talking about the statement below this one which is gone.

allait: I think we need to inline the `selling` block the same way we inlined `opportunities`

carolinegreen: That wouldn't work as they are separated by a block which is not within the same condition.  The entire point of writing it this way was to keep things DRY and group together conditionals using set, which is already used numerous times in the apps with conditionals fed into macros.

allait:
Looks good 👍

PR #306, opened 2016-05-03
Update to latest toolkit

pcraig3:
Fixes the bug where inline validation messages sit on the same line as inline text fields.


TheDoubleK:
:+1: if Travis likes it.

PR #308, opened 2016-05-06
[#118568577] Return to description of work

pcraig3:
The logic for deciding whether the secondary action link should return a user to the the overview or the section summary page was right, but not applied consistently.

The generated link was correct (and tested), but the link text itself was using slightly different logic (and wasn't being tested).

Fixed the logic and added some tests for link text on brief question pages.

[>> Bug in Tracker](https://www.pivotaltracker.com/story/show/118568577)


TheDoubleK:
Loks good after a rebase :+1:

PR #309, opened 2016-05-06
Update dependencies with new 'number' question type support

allait:
### Update toolkit form macro to support 'number' question type
Renames percentage form macro to number and passes the number
question 'unit' fields to the form template.

Number fields are always rendered as `smaller` text inputs.

#### Remove leading zeroes from test date formats


pcraig3:
👍

PR #319, opened 2016-05-13
Update frameworks requirement version to 1.3.6

Wynndow:
Updates the frameworks version to make use of [this](https://github.com/alphagov/digitalmarketplace-frameworks/pull/274) PR. It changes the link destination on the 'Additional terms and conditions' question to take the buyer directly to the standard contract.


pcraig3:
👍

PR #321, opened 2016-05-17
Update toolkit, make lot filter text smaller.

pcraig3:
Bring in the latest toolkit and remove local version of `_reset.scss` because the latest toolkit has a copy of this file.

Update `application.scss` to reflect this.

Also lowers the the text size of lot filters on the G-Cloud search page.  Since [the new version of the toolkit](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/262) is going to make the labels smaller, it seems like all filtering options should be the same size so that the association is clear.

***

## FILTERS -> 👎 
screen shot 2016-05-17 at 11 42 57

## <sub>filters</sub> -> 👍 
screen shot 2016-05-17 at 11 42 35


TheDoubleK:
It's all so tiny though.

TheDoubleK:
Having the text for lot selection bigger seems OK to me.

tombye:
LGMT if it’s clean to merge 👍

PR #323, opened 2016-05-23
buyers can view their q + a dates while their brief is live

ralph-hawkins:
Adds a link to 'View question and answer dates' on the overview page that links to `brief_publish_confirmation.html`.

image

The `brief_publish_confirmation.html` changes when a opportunity is live:
- removes the warnings that are no longer relevant and the publish button
- changes content to present tense.

The page isn't visible when the opportunity closes.

image

### DIFF
--- a/app/templates/buyers/brief_publish_confirmation.html
+++ b/app/templates/buyers/brief_publish_confirmation.html
@@ -28,20 +28,22 @@
 <div class='grid-row'>
   <div class='column-two-thirds large-paragraph'>
     {% with
-        heading = "Publish your requirements and evaluation criteria",
+        heading = "Publish your requirements and evaluation criteria" if not published else "Question and answer dates",
         smaller = True,
         with_breadcrumb = True
     %}
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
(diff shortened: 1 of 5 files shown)


allait:
We need tests for the new endpoint. Feel free to assign the PR to yourself if you want to pick this up.

TheDoubleK:
When the opportunity closes the link to the timeline page goes but the page itself is still accessible using a direct URL.  Will fix it to 404 for closed briefs.

TheDoubleK:
I added some tests - should be good to go now I think.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -101,16 +105,18 @@
           {{ summary.text("The last day suppliers can apply.") }}
```
pcraig3: <nitpick> 
In the extra bit of the message that warns about Saturdays and Sundays, we can tell at that point whether the date in question is going to be one of those days, right? 
Wouldn't it be more cool to have a contextual message say something like "Since {} is a Saturday, please publish questions and answers by {}"?  
</nitpick>

PR #324, opened 2016-05-23
Update 'View Opportunities' page to include missing 'and'

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/118472411) story on Pivotal.

This PR updates the view opportunities page to use a new filter (see [this PR](https://github.com/alphagov/digitalmarketplace-utils/pull/265)) which will convert a list of strings to a properly formatted string. It joins all elements with a ', ' except for the last which it joins with ' and '.

If a list has only one element, it just returns that element as a lower case string without any commas or 'and'.

If the list has no elements it will just return an empty string. This should never occur in this use case, but the functionality has been added in case it is needed for another application in the future.

All strings are converted to lowercase.

See the screenshot below.

screen shot 2016-05-23 at 17 09 55

### DIFF
--- a/app/templates/briefs_catalogue.html
+++ b/app/templates/briefs_catalogue.html
@@ -25,7 +25,7 @@
     <div class="grid-row">
         <section class="column-two-thirds">
             <div class="marketplace-paragraph">
-              <p>View buyer requirements for {{ lot_names | map('lower') | join(", ") }}.</p>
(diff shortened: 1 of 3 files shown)


allait:
👍 

(Although looking at it in context now, I'm not sure about `smartjoin` name. Since it's used without an argument it's not really obvious that it's using commas and 'and'. But that's something we can revisit later.)

PR #325, opened 2016-05-24
Fix link to DM buyers' guide

tombye:
As requested in https://www.pivotaltracker.com/story/show/119579917.

### DIFF
--- a/app/templates/_footer_categories.html
+++ b/app/templates/_footer_categories.html
@@ -40,7 +40,7 @@
         </h2>
         <ul>
             <li>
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Does this also need to be fixed in the supplier app?  Footer links there should be the same as here.

TheDoubleK:
:+1: for this one though.

tombye:
Good point, now up: https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/486

PR #327, opened 2016-05-24
Add gunicorn to requirements.txt for PaaS deploys

allait:
Add gunicorn to requirements.txt for running the app on Cloud Foundry
(possibly not ideal see [1]).

Explicitly add cffi to requirements so that Cloud Foundry spots it and
adds libffi.

There doesn't seem to be an easy way to use a different requirements
file for Cloud Foundry, so we'll have to add the dependencies
everywhere.

[1] https://github.com/etianen/django-herokuapp/issues/9


TheDoubleK:
When you say "not ideal" - am I right in thinking this will only be an issue for when apps are deployed to PaaS?  The new dependencies won't have any affect when running in AWS, will they?

allait:
@TheDoubleK yeah, the issue is about gunicorn vs alternatives when running on PaaS. The dependencies aren't used when running in AWS, but they do get installed, which might have an effect on deployments. Ideally, there'd be an easy way to have a different requirements file for PaaS, but I couldn't think of one that wouldn't involve writing wrapper code.

TheDoubleK:
Yeah it's a bit weird but like you say, no easy way round it.
:+1:

PR #329, opened 2016-05-27
Adding phone number to buyer

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/119726147) story on Pivotal.

This PR compliments [this](https://github.com/alphagov/digitalmarketplace-api/pull/399) one for the api.

It updates the app to include a field for phone number when buyer's create an account. It's included on the page with the name and password field.

The phone number is validated on the Regex pattern used in the user json_schema. I suspect that validation message will need changing. Screenshots below.

screen shot 2016-05-27 at 11 56 54
___________________________________________________________________________
screen shot 2016-05-27 at 12 11 40

### DIFF
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -65,8 +65,9 @@ class CreateUserForm(Form):
     phone_number = StringField(
         'Phone number', id="input_phone_number",
         validators=[
-            Regexp("^$|^\\+?([\\d\\s()-]){6,20}$",
-                   message="Phone numbers must be between 6 and 20 characters."
+            Regexp("^$|^\\+?([\\d\\s()-]){9,20}$",
+                   message=("Phone numbers must be at least 9 characters long. "
+                            "They can only include digits, spaces, plus and minus signs, and brackets.")
                    )
         ]
     )
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -1,7 +1,7 @@
 from flask_wtf import Form
 from wtforms import PasswordField
-from wtforms.validators import DataRequired, Email, EqualTo, Length
-from dmutils.forms import StripWhitespaceStringField
+from wtforms.validators import DataRequired, Email, EqualTo, Length, Regexp
+from dmutils.forms import StripWhitespaceStringField, StringField
 
 
 class LoginForm(Form):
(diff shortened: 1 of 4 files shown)


app/main/forms/auth_forms.py
```diff
@@ -62,6 +62,15 @@ class CreateUserForm(Form):
         ]
     )
 
+    phone_number = StringField(
+        'Phone number', id="input_phone_number",
+        validators=[
+            Regexp("^$|^\\+?([\\d\\s()-]){6,20}$",
+                   message="Phone numbers must be between 6 and 20 characters."
```
TheDoubleK: I don't think this message is quite right. Our pattern requires 6-20 digits (not any old characters), but actually could be up to 21 if they have an optional `+` at the start.  We don't want to get into those details in the message though.

How about `You must enter a valid phone number.`?  But probably best to check with @superroz or Sue for the ideal wording.

superroz: Hey,

Let's maybe update to 21 (characters includes digits and + sign doesn't it?). It's phrased like this to be consistent with the other error messages on the page. I'd like to update all the messages on this page so they're active, eg 'You must', but that felt like a separate story and didn't want there to be mission creep. Can talk through alternatives on Monday.

tests/app/views/test_login.py
```diff
@@ -859,6 +859,7 @@ def test_should_create_user_if_user_does_not_exist(self, data_api_client):
             'role': 'buyer',
             'password': 'validpassword',
             'emailAddress': 'test@email.com',
+            'phoneNumber': '',
```
TheDoubleK: Is it worth having some tests where it does pass in a phone number too?

TheDoubleK:
Tested locally and all works OK.  Just need to get the validation message wording straight and should be good to go.

app/main/forms/auth_forms.py
TheDoubleK: I reckon fix all three messages now. 

Our phone number validation is a bit funky, saying 21 characters still wouldn't be true because it depends on what characters they are and in what order. We don't want to get into the details in our messaging though.

superroz: Have talked to Sarah and Chris. We're going to update to this message for now: 

'Phone numbers must be at least 7 characters long. They can only include digits, spaces, plus and minus signs, and brackets.'

But will work on a separate story to make the validation active and consistent on this page and elsewhere in the app.

Wynndow:
Have updated the front end validation to 9 characters after discussions with Sarah and Roz, as well as the validation message that appears. See screenshot below.

I've also updated the tests to actually check a phone number is sent to the api.

screen shot 2016-06-01 at 16 14 47

allait:
I think it might be worth adding tests to check that empty phone number is accepted and that for values that don't match the pattern we get the error message we expect.

allait:
👍

PR #331, opened 2016-06-03
Track the publishing of a brief

tombye:
Publishing a brief redirects to the brief overview page (which has the same URL as normally).

We need to track this action so this adds a `published=true` query string parameter to the URL when it is requested as a result of a publish.

https://www.pivotaltracker.com/story/show/119634461


pcraig3:
Looks 👍  to me.

Might get multiple hits if people refresh the page or whatever (but since they're all tied to brief ids that's not really an issue).

tombye:
@pcraig3 yeah, I ran that past Ash and he wants to try it out with some real requests coming in to see how big an issue it is (whether it needs much filtering). Rebasing now...

tombye:
For the record, there were some questions about whether this could be done through sending a virtual pageview rather than our messing with the URL.

The problem is that publishing a brief doesn't produce a confirmation banner in the page we redirect to but rather switches that page to a state it has from that point onwards. Without having a flash message to protect from page refreshes, we might as well stick to the original approach.

PR #332, opened 2016-06-06
Adjust button width for new padding

tombye:
The 'Show services' button on the G-Cloud index page is broken:

image

This is because of a change made to code the govuk frontend toolkit which came through when we bumped the version this app uses to `14.12.0`.

Story: https://www.pivotaltracker.com/story/show/120939061

The horizontal padding for all buttons using the `button` mixin was increased from 1.2em
(0.6em x 2) to 1.578946em (0.789473em x 2) in version `14.12.0` of the govuk frontend toolkit:

https://github.com/alphagov/govuk_frontend_toolkit/commit/227093d05ea6faa1c3a6dd1aed919ce607b66182

The 'Show services' button has its width set to 7.6em (roughly 183px with the 24px font-size) so
when the padding was increased, this reduced the space available for the text.

button_diff

In the CSS for this button, bearing in mind 1px is 0.04166em (1 / 24), the following is true:

Before the change the text had 6.4em (7.6em - 1.2em) of space (153.592px):

button_old_css

After the change the text had 6.02105em (7.6em - 1.578946em) of space (144.496px):

button_broken_css

This changes the width so it includes the space the text had before + the new padding.

This ends up as 7.97894em (6.4em + 1.57894em) which we round up to a width of 8em.

Here are the dimensions this leaves us with:

button_new_css

...and the button now looks like:

image

### DIFF
--- a/app/assets/scss/_index-page.scss
+++ b/app/assets/scss/_index-page.scss
@@ -44,7 +44,7 @@
       height: 50px;
 
       @include media(tablet) {
-        margin-right: 7.6em;
+        margin-right: 8em;
       }
     }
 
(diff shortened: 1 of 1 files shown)


pcraig3:
Seems legit

👍

PR #334, opened 2016-06-07
Update dmutils to 21.0.0

allait:
Adds a common config class for live environments, similar to how we set up configs for the API. Moves DM_LOG_PATH to the new Live config.


TheDoubleK:
:+1:

PR #336, opened 2016-06-07
Update dmcontent with mandatory question options support

allait:
Adds mandatory options to the assessment methods question. Updates tests to use lists for `evaluationType` values to match the actual data type.


TheDoubleK:
:+1:

Does releasing this need to be coordinated with anything else (API migration?), or are the other pieces in place already?  There's the technical competence stuff to go out too today - let's talk about what order to merge and release.

allait:
I think this is safe to release before API. It will remove the mandatory options from the interface, which works well with existing validation. The only issue is that it will show duplicate "Work history" items in summary tables before API is released.

PR #338, opened 2016-06-10
Fix template footer css

tombye:
We were getting some issues with the footer alignment in chrome:

Was:

image

Now:

image

This includes a fix for that (7e3057a) but also uses the opportunity to bump the govuk template used to the latest version and use the [new grids](https://github.com/alphagov/govuk_frontend_toolkit/pull/136) for the footer's category sections.


TheDoubleK:
:+1:

PR #340, opened 2016-06-13
[#121388173] Fix for tick marks on the brief overview

pcraig3:
There were two problems with tick marks that cropped up.

1. Sections containing only one question would never show ticks.
2. Sections containing no required questions would always show ticks.

Added some more template code to show ticks next to single questions, so that fixed the first case.

In the second case, new logic goes like:

```
if required_questions exist:
  count unanswered required questions
else:
  count unanswered optional questions
```

Also, I gave up on my `is_*` naming convention for methods that return boolean values because it just sounded too weird. (ie, otherwise it would have been called `is_section_has_at_least_one_required_question`)

[>> Bug bug bug](https://www.pivotaltracker.com/story/show/121388173)

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -10,7 +10,8 @@ from .. import buyers, content_loader
 from ...helpers.buyers_helpers import (
     all_essentials_are_true, counts_for_failed_and_eligible_brief_responses,
     get_framework_and_lot, get_sorted_responses_for_brief, count_unanswered_questions,
-    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct, get_publishing_dates
+    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct, get_publishing_dates,
+    section_has_at_least_one_required_question
 )
 
 from dmapiclient import HTTPError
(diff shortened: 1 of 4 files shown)


app/buyers/views/buyers.py
```diff
@@ -115,7 +116,10 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
     completed_sections = {}
     for section in sections:
         required, optional = count_unanswered_questions([section])
-        completed_sections[section.slug] = True if required == 0 else False
+        if section_has_at_least_one_required_question(section):
+            completed_sections[section.slug] = True if required == 0 else False
+        else:
+            completed_sections[section.slug] = True if optional == 0 else False
```
TheDoubleK: Hmm... a thought.  Should be tick only if ALL optional are completed?
Or should it be `if optional < total_optional`?

At the moment it's a moot point, as the only section with all optional has just one question.  So let's leave it like this for now.

tests/unit/test_buyers_helpers.py
```diff
@@ -74,6 +74,29 @@ def test_brief_can_be_edited(self):
...
         assert helpers.buyers_helpers.brief_can_be_edited(api_stubs.brief(status='live')['briefs']) is False
 
+    def test_section_has_at_least_one_required_question(self):
+        brief = {
+            'status': 'draft',
+            'frameworkSlug': 'dos',
+            'lotSlug': 'digital-specialists',
+            'required1': True
+            }
+        content = content_loader.get_manifest('dos', 'edit_brief').filter(
```
TheDoubleK: It would be nice to use a test manifest for this rather than the actual DOS one... but that's probably a bigger job for another time.

TheDoubleK:
:+1:

app/buyers/views/buyers.py
pcraig3: Yeah, interesting question for another time.

tests/unit/test_buyers_helpers.py
pcraig3: Actually, that's a reasonable point.  This test doesn't actually need the summary bit for anything.  Can cut it out and save a few lines.

PR #341, opened 2016-06-13
Render all section links as lists.

pcraig3:
Previously, we would either render section links as either `li`s in a list (if there were multiple questions in a section) or as `a` elements (if there was just one question in a section).

Rendering things two different ways wasn't really giving us anything and it basically confused anyone who came into contact with it (including me), as well as our functional tests.

We're presenting all section links identically, whether they're the only question in a section or otherwise, so we should render them consistently.


TheDoubleK:
:+1:

PR #342, opened 2016-06-13
Update links in header

idavidmcdonald:
Remove "Create supplier account" link from the header
Logged in users will now see "View your account" and take them to
their respective dashboard page.
View your account link comes before the Log out link.
We changed from current_user.email_address to
current_user.is_authenticated() to check if a user is logged in or
not.

### DIFF
--- a/app/templates/_proposition_header.html
+++ b/app/templates/_proposition_header.html
@@ -4,12 +4,16 @@
     <nav id="proposition-menu">
       <a href="/" id="proposition-name">Digital Marketplace</a>
       <ul id="proposition-links">
-        {% if current_user.email_address %}
+        {% if current_user.is_authenticated() %}
+          {% if current_user.role == 'buyer' %}
+            <li><a href="{{ url_for('buyers.buyer_dashboard') }}">View your account</a></li>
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
Old:

image



New:
image

app/templates/_proposition_header.html
```diff
@@ -4,12 +4,16 @@
     <nav id="proposition-menu">
       <a href="/" id="proposition-name">Digital Marketplace</a>
       <ul id="proposition-links">
-        {% if current_user.email_address %}
+        {% if current_user.is_authenticated() %}
+          {% if current_user.role == 'buyer' %}
+            <li><a href="{{ url_for('buyers.buyer_dashboard') }}">View your account</a></li>
```
TheDoubleK: Looking good. Could maybe only make the url conditional, rather than the whole `li`, e.g.
```
 <li>
<a href="{{ url_for('buyers.buyer_dashboard') if  current_user.role == 'buyer' else '/suppliers' }}">
```

TheDoubleK:
Minor comment left but :+1: as it is if you don't like it.
We should merge/release this in step with the corresponding change to the supplier app header.

PR #344, opened 2016-06-14
Increase size of G-Cloud search button to fix bug

tombye:
Firefox's font rendering seems to need a bit more space. @andrewchong noticed the current styling fails at certain page zoom levels.

## Before (at 90% zoom)

90_percent_before

## After (at 90% zoom)

90_percent_after

## After (at 50% zoom)

50_percent_before

## After (at 50% zoom)

50_percent_after


pcraig3:
button

👍

PR #345, opened 2016-06-15
Reduce effect of padding changes to footer-meta

tombye:
As noticed by @pcraig3 on the supplier app, recent changes have reduced the `padding-bottom` on the global footer:

image

This fixes it:

image


pcraig3:
👉 😎 👉   --> 👍

PR #350, opened 2016-06-20
Fix email validation bug

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/121730503) bug on Pivotal.

The WTForms inbuilt validator for Emails uses a very loose Regex (`'^.+@[^.].*\.[a-z]{2,10}$'`) which would allow an email address with two `@`'s to be used. This meant an invalid email could be passed down to the API which would throw a value error and pass it back up to the front end.

This PR changes the validation of email addresses from WTForms Email validation to a Regex validation which is a lot tighter and solves this issue, and probably many more undiscovered issues.

Another PR will be made on the API to catch the invalid format properly there.

double 2

### DIFF
--- a/app/main/forms/auth_forms.py
+++ b/app/main/forms/auth_forms.py
@@ -10,7 +10,7 @@ class LoginForm(Form):
         'Email address', id="input_email_address",
         validators=[
             DataRequired(message="You must provide an email address"),
-            Regexp("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$",
+            Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
                    flags=re.IGNORECASE,
                    message="You must provide a valid email address")
         ]
@@ -28,7 +28,7 @@ class EmailAddressForm(Form):
         'Email address', id="input_email_address",
         validators=[
             DataRequired(message="You must provide an email address"),
-            Regexp("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$",
+            Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
                    flags=re.IGNORECASE,
                    message="You must provide a valid email address")
         ]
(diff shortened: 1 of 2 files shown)


app/main/forms/auth_forms.py
```diff
@@ -9,7 +10,9 @@ class LoginForm(Form):
         'Email address', id="input_email_address",
         validators=[
             DataRequired(message="You must provide an email address"),
-            Email(message="You must provide a valid email address")
+            Regexp("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$",
```
TheDoubleK: I think this regex is definitely going to be too restrictive.  What about extended unicode, which is perfectly valid in an email address, e.g. `ñoñó1234@gov.uk`.

On reflection (and after reading this gem: http://stackoverflow.com/questions/201323/using-a-regular-expression-to-validate-an-email-address) I'm tempted to stick with the WTForms validation out of the box, and just fix the API to not break in the case of multiple `@` signs.

No user account is set up until they click the link in their email, so that's really the proper test that the thing they typed in is a valid email address.  We should just make it so our app doesn't break if someone types the wrong thing in.

Wynndow: How about using the pattern that is used in the users json schema? 
`^[^@^\\s]+@[^@^\\.^\\s]+(\\.[^@^\\.^\\s]+)+$`

At least that way they can't use an email which will definitely fail later?

TheDoubleK: Good plan - makes perfect sense - do it!

app/main/forms/auth_forms.py
```diff
@@ -9,7 +10,9 @@ class LoginForm(Form):
         'Email address', id="input_email_address",
         validators=[
             DataRequired(message="You must provide an email address"),
-            Email(message="You must provide a valid email address")
+            Regexp("^[^@^\s]+@[^@^\.^\s]+(\.[^@^\.^\s]+)+$",
+                   flags=re.IGNORECASE,
```
TheDoubleK: Don't think we need this flag now, as the regex doesn't specify any "cased" characters, just `@` and `.`.

TheDoubleK:
minor comment, but should be good to go after a rebase.

app/main/forms/auth_forms.py
Wynndow: Yep, good call.

TheDoubleK:
Rebased :+1: - will merge once Travis is happy.

PR #353, opened 2016-06-22
[#121961157] Re-add format_links filter now that it has been fixed

TheDoubleK:
Work done with @carolinegreen 

Fixes this bug: https://www.pivotaltracker.com/story/show/121961157

Pulls in latest utils with fixed filter, and re-adds the filter to templates where it should be added.

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -158,6 +158,10 @@ $path: "/static/images/";
   }
 }
 
+.break-link {
+  word-break: break-all;
+}
+
 .summary-item-body {
   a {
     word-break: break-all;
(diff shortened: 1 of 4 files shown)


TheDoubleK:
:+1:

carolinegreen:
👍

app/assets/scss/application.scss
```diff
@@ -158,6 +158,10 @@ $path: "/static/images/";
   }
 }
 
+.break-link {
+  word-break: break-all;
```
pcraig3: Too late now, but I have a suspicion that `break-word` with a fallback to `break-all` might be more elegant.

carolinegreen: I went for break all as some of the links were absolutely horrible and didn't really have words in them, so breaking everything worked better here.

PR #354, opened 2016-06-24
Update toolkit to 16.0.0

pcraig3:
**Note: needs https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/269 to be merged first.**

Changes a class name to match the new one in the updated `instruction-list` pattern, used on the brief overview page. No other breaking changes.

### DIFF
--- a/app/assets/scss/_brief_submission.scss
+++ b/app/assets/scss/_brief_submission.scss
@@ -1,3 +1,20 @@
+.instruction-list-item {
+
+  h2 {
+    @include bold-27;
+    margin-bottom: 10px;
+  }
+
+  .instruction-list-item-top {
+    @include core-19;
+    display: block;
+  }
+
+  a {
+    text-decoration: none;
+  }
+}
+
 .last-edited {
     margin: 10px 0 15px 0;
 }
@@ -29,8 +46,8 @@
   }
 }
 
-ul.instruction-list-item-extra-information li,
-ol.instruction-list-item-extra-information li {
+ul.instruction-list-item-top li,
+ol.instruction-list-item-top li {
   margin-bottom: 0;
 }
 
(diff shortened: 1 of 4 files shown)


app/assets/scss/_brief_submission.scss
```diff
@@ -1,3 +1,20 @@
+.instruction-list-item {
```
TheDoubleK: Shouldn't this all get pushed back to the instruction list in the toolkit, rather than being in just this app?

TheDoubleK: Having different instruction list styling in different apps seems completely the opposite point of having a frontend toolkit?

pcraig3: @idavidmcdonald and I talked to @tombye about this and we agreed that (a) yes, your point is a good one and you are right and also that (b) we want to merge this now and fix it later.

Currently, this pattern is being used only twice (soon to be three times).  
- It's being used correctly on the download agreement page (ie, by using the toolkit macro)
- It _will be_ used correctly on the new "signing your agreement" flow
- And it's also used incorrectly on the brief overview/grandparent page where we're using instruction-list classnames but then overriding CSS rules.

As you've pointed out, overriding the general styles in our apps is contrary to what the toolkit is meant to be for and we shouldn't be doing this.

However, having acknowledged that the pattern isn't really working in a general sense, the three times we're using it are all currently working fine and we're not doing anything to make it worse in this pull request.  Breaking the pattern into something more reusable and modular is definitely the going-forward solution, but at this point it just doesn't seem that important.  It's also not entirely clear what breaking down the pattern means in terms of actual code, so it seems like it would require a lot of time to fix properly something that's, frankly, just not a priority right now.

As a side note, the [story](https://www.pivotaltracker.com/story/show/121577345) that David and I are working on has been going for about 2 weeks and has meant writing (and rewriting) a bunch of changes to 4 different repos so far.  Just as totally flawed and selfish humans, we really don't want to be hung up on this particular issue.

TheDoubleK:
:joy:

:+1:

PR #355, opened 2016-06-24
[WIP] Remove Digital Services Framework page & route

tombye:
NOT TO BE DEPLOYED UNTIL JULY 5th.

Remove the page for the Digital Services Framework:

https://www.pivotaltracker.com/story/show/122197827


app/main/views/digital_services_framework.py
```diff
@@ -1,12 +0,0 @@
-# -*- coding: utf-8 -*-
-from __future__ import unicode_literals
-
-from flask import render_template
-from ...main import main
-
-
-@main.route('/digital-services/framework')
```
TheDoubleK: We should return a 410 (GONE) here I think, rather than deleting the route completely:
`The 410 (Gone) status code SHOULD be used if the server knows, through some internally configurable mechanism, that an old resource is permanently unavailable and has no forwarding address.`
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

Wynndow:
Looks good to me! :+1:

TheDoubleK:
Cheers @Wynndow - will merge tomorrow when we need to ship it.

PR #356, opened 2016-06-24
Remove dss analytics

tombye:
We no longer link to https://digitalservicesstore.service.gov.uk/ so the analytics code should be removed.

https://www.pivotaltracker.com/story/show/122197843

For reference, here are the pull requests that added and fixed the cross-domain tracking:
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/121
- https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/124


pcraig3:
Forgot about this earlier, sorry.

I can confirm that there's some obscure-looking variable about [autolinking](https://support.google.com/analytics/answer/1034342?hl=en) that's now gone, and that the three tests that were removed fail if you put them back in.

Seems legit.

👍

PR #359, opened 2016-07-04
Fix break words

carolinegreen:
Fix break words in firefox.  Break-all only works in table cells in firefox if it is displayed as an inline block.


TheDoubleK:
:+1:

TheDoubleK:
I'm going to merge this now to ship it along with the regex fix this morning.

PR #360, opened 2016-07-04
[#122036279] create_your_account_complete: don't unnecessarily clear session, simplify handling of email_sent_to

risicle:
clearing the session had the inadvertent effect of resetting the user's csrf token, causing a users
cached forms (accessed through the back button) to no longer be accepted due to a "csrf error". confusing for users attempting to create multiple buyer accounts. this would appear to be a copypasta error resulting from the original code being copied from the supplier app which _does_ require an amount of session-clearing.

Story https://www.pivotaltracker.com/story/show/122036279

This is something that it might be nice to have a functional (regression) test for... (though this is reportedly a firefox-only issue and our functional tests only use webkit AFAIK)

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -324,12 +324,7 @@ def submit_create_user(encoded_token):
 
 @main.route('/create-your-account-complete', methods=['GET'])
 def create_your_account_complete():
-    if 'email_sent_to' in session:
-        email_address = session['email_sent_to']
-    else:
-        email_address = "the email address you supplied"
-    session.clear()
-    session['email_sent_to'] = email_address
+    email_address = session.setdefault("email_sent_to", "the email address you supplied")
     return render_template(
         "auth/create-your-account-complete.html",
         email_address=email_address), 200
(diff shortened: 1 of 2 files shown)


risicle:
Actually... I now question whether the session needs to be altered at all in this view. No other view appears to _read_ the `email_sent_to` session key...

risicle:
Ok, surveying the equivalent supplier frontend code has spawned new story https://www.pivotaltracker.com/story/show/125494661

allait:
I quite like the idea of clearing a session when the user logs in – we should't need anything that existed in the anonymous session and starting from a clean state each time keeps things a little simpler.

Re. the back button – I think @TheDoubleK and I were going back and forth on this, but I'm not sure if we want to allow logged in users to visit the registration page at all. It doesn't seem like going back to the registration form and submitting it again is a useful action. So I'd suggest keeping it as is (maybe moving session reset before `login_user`) or fixing the error page if users are running into this. But I'd like to hear what everyone else thinks first.

At any rate, I'm glad that we know what was going on with CSRF reset.

risicle:
The user isn't quite _logging in_ here, just having the invitation email sent to themselves.

The way we found out about this was through a user wanting to sign up several of their colleagues and hitting back to get back to the form repeatedly, which is a reasonable-enough thing to do.

allait:
Yeah, you're right. Let's do this then 👍

risicle:
Decided to add a regression test before I merged this...

PR #361, opened 2016-07-06
Fix responses CSV for briefs with an empty list in nice to have requirements

allait:
It looks like at the moment our logic for removing empty items saves `[]` to optional list questions when the user submits an empty page, which breaks the CSV download, since the `niceToHaveRequirements` key is present in the brief data but not in any of the responses (since the forms aren't rendered).

What happens with the brief data right now:

1. User clicks "Save" without entering any of the nice to have requirements
2. Buyer frontend recieves form data with two `""` form fields
3. Buyer frontend submits `{"niceToHaveRequirements": ["", ""]}` to the API
4. API runs `utils.strip_whitespace_from_data` and drops empty strings from `niceToHaveRequirements`, replacing `["", ""]` with `[]`
5. `{"niceToHaveRequirements": []}` is stored to the database

I'm not sure if that's a result of a recent change in content loader or if it's caused by users submitting nice-to-have requirements when they submit essentials since they're now on one page.

It feels like we don't want a key to exist when there are no values, but that requires some more digging and we'd need to run a migration to update existing brief, so for now I'm fixing the broken behaviour in the buyer frontend to support both missing key an an empty list.

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -295,7 +295,7 @@ def download_brief_responses(framework_slug, lot_slug, brief_id):
     # Build header row from manifest and add it to the list of rows
     for question in section.questions:
         question_key_sequence.append(question.id)
-        if question['type'] == 'boolean_list' and question.id in brief:
+        if question['type'] == 'boolean_list' and brief.get(question.id):
             column_headings.extend(brief[question.id])
             boolean_list_questions.append(question.id)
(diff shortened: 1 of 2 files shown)


TheDoubleK:
Ah, damn, yes I suspect it's moving both things to the same page that has lead to this, but not 100% sure.

Fix looks legit :+1:

PR #365, opened 2016-07-25
Specialist briefs can be open for 1 or two weeks

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

Will need separate PR's on [utils](https://github.com/alphagov/digitalmarketplace-utils/pull/275), [apiclient](https://github.com/alphagov/digitalmarketplace-apiclient/pull/30), [api](https://github.com/alphagov/digitalmarketplace-api/pull/422) and [frameworks](https://github.com/alphagov/digitalmarketplace-frameworks/pull/294) to be accepted before this one.

The logic for calculating dates has been moved to dmutils so that it can be used by other apps, specifically the api app.

The dates are shown on the publishing page when the buyer is creating a brief and are dependent on the brief length chosen during creation of the brief. If that question hasn't been answered yet a link to the question is shown instead.

The catalogue of briefs is ordered slightly differently now; by live then closed, and by publishing date within that. The published date is also displayed for briefs to give more context.

The dmspeak styles have been imported from the frontend toolkit for use on the publishing page, which was using prototype styles which had leaked into the app.

Screenshots
---------------

Requirements length question in overview.

screen shot 2016-07-25 at 13 26 53
---------------
Selecting length

screen shot 2016-07-25 at 13 27 22
---------------
Publishing page with no length set

screen shot 2016-07-25 at 13 28 58
---------------
Publishing page with length set

screen shot 2016-07-25 at 13 27 45
----------------
Publishing page with length set and no outstanding required questions.

screen shot 2016-07-25 at 13 30 42
----------------
Dates post publishing

screen shot 2016-07-25 at 13 31 07
---------------
Briefs catalogue
screen shot 2016-07-25 at 13 31 20
screen shot 2016-07-25 at 13 31 37

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -10,11 +10,12 @@ from .. import buyers, content_loader
 from ...helpers.buyers_helpers import (
     all_essentials_are_true, counts_for_failed_and_eligible_brief_responses,
     get_framework_and_lot, get_sorted_responses_for_brief, count_unanswered_questions,
-    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct, get_publishing_dates,
+    brief_can_be_edited, add_unanswered_counts_to_briefs, is_brief_correct,
     section_has_at_least_one_required_question
 )
 
 from dmapiclient import HTTPError
+from dmutils.dates import get_publishing_dates
 
 
 @buyers.route('/buyers')
@@ -362,6 +363,11 @@ def publish_brief(framework_slug, lot_slug, brief_id):
             question_and_answers['slug'] = section['id']
 
     unanswered_required, unanswered_optional = count_unanswered_questions(sections)
+
+    if sections.get_section('set-how-long-your-requirements-will-be-live-for') and \
+            sections.get_section('set-how-long-your-requirements-will-be-live-for').questions[0].answer_required:
+            unanswered_required -= 1
+
     if request.method == 'POST':
         if unanswered_required > 0:
             abort(400, 'There are still unanswered required questions')
@@ -372,7 +378,7 @@ def publish_brief(framework_slug, lot_slug, brief_id):
                     brief_id=brief['id'], published='true'))
     else:
         email_address = brief_users['emailAddress']
-        dates = get_publishing_dates()
+        dates = get_publishing_dates(brief)
 
         return render_template(
             "buyers/brief_publish_confirmation.html",
@@ -389,7 +395,6 @@ def publish_brief(framework_slug, lot_slug, brief_id):
 def view_brief_timeline(framework_slug, lot_slug, brief_id):
     get_framework_and_lot(framework_slug, lot_slug, data_api_client, status='live', must_allow_brief=True)
     brief = data_api_client.get_brief(brief_id)["briefs"]
-
     if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id) or brief.get('status') != 'live':
         abort(404)
 
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -373,7 +373,6 @@ def publish_brief(framework_slug, lot_slug, brief_id):
             url_for('.view_brief_overview', framework_slug=brief['frameworkSlug'], lot_slug=brief['lotSlug'],
                     brief_id=brief['id'], published='true'))
     else:
-
         #  requirements length is a required question but is handled separately to other
         #  required questions on the publish page if it's unanswered.
         if sections.get_section('set-how-long-your-requirements-will-be-live-for') and \
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -364,10 +364,6 @@ def publish_brief(framework_slug, lot_slug, brief_id):
 
     unanswered_required, unanswered_optional = count_unanswered_questions(sections)
 
-    if sections.get_section('set-how-long-your-requirements-will-be-live-for') and \
-            sections.get_section('set-how-long-your-requirements-will-be-live-for').questions[0].answer_required:
-            unanswered_required -= 1
-
     if request.method == 'POST':
         if unanswered_required > 0:
             abort(400, 'There are still unanswered required questions')
@@ -377,6 +373,13 @@ def publish_brief(framework_slug, lot_slug, brief_id):
             url_for('.view_brief_overview', framework_slug=brief['frameworkSlug'], lot_slug=brief['lotSlug'],
                     brief_id=brief['id'], published='true'))
     else:
+
+        #  requirements length is a required question but is handled separately to other
+        #  required questions on the publish page if it's unanswered.
+        if sections.get_section('set-how-long-your-requirements-will-be-live-for') and \
+                sections.get_section('set-how-long-your-requirements-will-be-live-for').questions[0].answer_required:
+                unanswered_required -= 1
+
         email_address = brief_users['emailAddress']
         dates = get_publishing_dates(brief)
 
--- a/app/templates/buyers/brief_publish_confirmation.html
+++ b/app/templates/buyers/brief_publish_confirmation.html
@@ -45,7 +45,7 @@
             <p>This will show you what the supplier application deadline will be if you publish your requirements today.</p>
           </div>
         {% else %}
-          <p>Your requirements will be open for {{ dates.application_open_weeks }} {{ pluralize(dates.application_open_weeks, "week", "weeks") }}.</p>
+          <p>Your requirements will be open for {{ dates.application_open_weeks }}.</p>
           <div class="dmspeak">
             <p>If you publish your requirements today ({{ dates.published_date | shortdateformat }}), suppliers will be able to apply until {{ dates.closing_time }}, {{ dates.closing_date | shortdateformat}}.</p>
           </div>
--- a/app/templates/buyers/brief_publish_confirmation.html
+++ b/app/templates/buyers/brief_publish_confirmation.html
@@ -34,20 +34,40 @@
     %}
         {% include "toolkit/page-heading.html" %}
     {% endwith %}
+
     {% if not published %}
-      <p class="padding-bottom-small">All requirements are published on the Digital Marketplace where anyone can see them.</p>
-      <p class="padding-bottom-small">All requirements are open for {{ dates.application_open_weeks }} weeks.</p>
-      <p class="padding-bottom-small">If you publish your requirements today ({{ dates.today | shortdateformat }}), suppliers will be able to apply until {{ dates.closing_time }}, {{ dates.closing_date | shortdateformat}}.</p>
+      <div class="dmspeak">
+        <p>All requirements are published on the Digital Marketplace where anyone can see them.</p>
+      </div>
+        {% if brief.lotSlug == 'digital-specialists' and not brief.requirementsLength %}
+          <p><strong>You must set how long your requirements will be open for.</strong></p>
+          <p>You must set how long your requirements will be open for so you can see the dates you need to be aware of if you publish today.</p>
+          <div class="dmspeak">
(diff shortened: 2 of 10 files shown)


requirements.txt
```diff
@@ -5,9 +5,9 @@ inflection==0.2.1
 unicodecsv==0.14.1
 werkzeug==0.10.4
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@21.2.3#egg=digitalmarketplace-utils==21.2.3
+git+https://github.com/alphagov/digitalmarketplace-utils.git@21.4.0#egg=digitalmarketplace-utils==21.4.0
 git+https://github.com/alphagov/digitalmarketplace-content-loader.git@1.1.0#egg=digitalmarketplace-content-loader==1.1.0
```
idavidmcdonald: Would be good to also bump the content loader to 1.2.0. This is because the frameworks repo introduced a breaking change for v2.0.0 and requires the latest version of the content loader (1.2.0) to support this. (I think nothing would actually break as the buyer app doesn't call that new functionality but no harm being preemptive)

app/buyers/views/buyers.py
```diff
@@ -362,6 +363,11 @@ def publish_brief(framework_slug, lot_slug, brief_id):
             question_and_answers['slug'] = section['id']
 
     unanswered_required, unanswered_optional = count_unanswered_questions(sections)
+
+    if sections.get_section('set-how-long-your-requirements-will-be-live-for') and \
+            sections.get_section('set-how-long-your-requirements-will-be-live-for').questions[0].answer_required:
+            unanswered_required -= 1
```
idavidmcdonald: So it's clear that we are decreasing the unanswered_required question count by one, but it might be good to put a one line comment _why_ we are doing this

app/buyers/views/buyers.py
```diff
@@ -362,6 +363,11 @@ def publish_brief(framework_slug, lot_slug, brief_id):
...
+
     if request.method == 'POST':
         if unanswered_required > 0:
```
idavidmcdonald: So we don't want to abort if they have answered all questions apart from choosing how long their requirements will be live for?

idavidmcdonald:
General comment to change camel case in test name from `requirementLength` to `requirement_length`

requirements.txt
Wynndow: Cool, it's done.

app/buyers/views/buyers.py
Wynndow: @idavidmcdonald Do we not normally try and avoid commenting the code? Or do you think it's unobvious enough to merit it?

app/buyers/views/buyers.py
Wynndow: @idavidmcdonald The template won't render the publish button if they haven't answered the requirements length question so in the normal flow they wouldn't be able to get to the post route. There may be security implications if someone were to use postman or edit the html on the page to render the form. I could add another conditional to line 386 to check if it's a get request?

app/buyers/views/buyers.py
allait: Yeah, I think it's not clear why we're not counting `requirementsLength` question as unanswered. Apart from the code comment, commit message might be a good place to explain this.

app/buyers/views/buyers.py
allait: Would moving `unanswered_required -= 1` after the `POST` check fix the issue?

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -56,7 +75,7 @@
           {% endif %}
         {% endfor %}
       {% endfor %}
-    {% else %}
+    {% elif not brief.lotSlug == 'digital-specialists' or brief.requirementsLength %}
```
allait: It's a minor point, but I'd go with `{% elif brief.lotSlug != 'digital-specialists' or brief.requirementsLength %}` to avoid the ambiguous `not` operator precedence

app/buyers/views/buyers.py
Wynndow: Cool, will add a comment. There's a bit of an explanation is the comment for this 1993e437d74d50f2f18dc6718db864eb2afc7398  commit which will help too

app/buyers/views/buyers.py
Wynndow: I think it needs to be checked for get requests too due to the edge case that requirements length is the only unanswered required question. In this case the heading for the list of unanswered questions would be rendered, but there would be nothing in the list.

Wynndow: Sorry, just realised you meant after the entire `if`, not just into it.

app/templates/buyers/brief_publish_confirmation.html
Wynndow: Cool, done.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -34,20 +34,39 @@
...
+      <div class="dmspeak">
+        <p>All requirements are published on the Digital Marketplace where anyone can see them.</p>
+      </div>
+        {% if brief.lotSlug == 'digital-specialists' and not brief.requirementsLength %}
+          <p><a href="{{ url_for('.edit_brief_question', framework_slug=brief.frameworkSlug, lot_slug=brief.lotSlug, brief_id=brief.id, section_slug='set-how-long-your-requirements-will-be-live-for', question_id='requirementsLength') }}">Set how long your requirements will be open for</a></p>
+          <div class="dmspeak">
+            <p>This will show you what the supplier application deadline will be if you publish your requirements today.</p>
+          </div>
+        {% else %}
+          <p>Your requirements will be open for {{ dates.application_open_weeks }} {{ pluralize(dates.application_open_weeks, "week", "weeks") }}.</p>
```
pcraig3: Made a comment about this [in the dm-utils PR](https://github.com/alphagov/digitalmarketplace-utils/pull/275/files#r72600783). Basically, it seems like we're recreating something we already have.

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -94,12 +113,9 @@
           {{ summary.text("The last day suppliers can ask questions.") }}
         {% endcall %}
         {% call summary.row(bold_border=True) %}
-          {{ summary.field_heading_date(dates.answers_close | shortdateformat, rowspan='2', scope='row') }}
+          {{ summary.field_heading_date(dates.answers_close | shortdateformat, scope='row') }}
           {{ summary.text("You must have published answers to all suppliers’ questions.") }}
         {% endcall %}
-        {% call summary.row(no_border=True) %}
-          {{ summary.text("If {} is a Saturday or Sunday, publish all questions and answers by the Friday before.".format(dates.answers_close | shortdateformat), border=True) }}
```
pcraig3: 👌

tests/app/views/test_buyers.py
```diff
@@ -1018,6 +1024,126 @@ def test_publish_button_unavailable_if_questions_not_answered(self, data_api_cli
...
+        res = self.client.get("/buyers/frameworks/digital-outcomes-and-specialists/requirements/"
+                              "digital-specialists/1234/publish")
+        page_html = res.get_data(as_text=True)
+
+        assert res.status_code == 200
+        assert 'Your requirements will be open for 2 weeks.' in page_html
+        assert 'This will show you what the supplier application deadline will be' not in page_html
+        assert 'Your requirements will be open for 1 week' not in page_html
+
+    def test_correct_content_is_displayed_if_requirementLength_is_not_set(self, data_api_client):
```
pcraig3: What's the difference between this test and `test_correct_content_is_displayed_if_no_requirementLength_is_set`?  Do we expect there to be briefs with `requirementsLength: ''`?

app/templates/buyers/brief_publish_confirmation.html
Wynndow: :shipit:

tests/app/views/test_buyers.py
Wynndow: Yeah, this test should be `'requirementsLength': None` and not an empty string. It's the case when there is a `requirementsLength` key, but it has no value. This can come from the API. The other test has no `requirementsLength` key. Have updated the test, everything still passes.

requirements.txt
```diff
@@ -5,9 +5,9 @@ inflection==0.2.1
 unicodecsv==0.14.1
 werkzeug==0.10.4
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@21.2.3#egg=digitalmarketplace-utils==21.2.3
-git+https://github.com/alphagov/digitalmarketplace-content-loader.git@1.1.0#egg=digitalmarketplace-content-loader==1.1.0
-git+https://github.com/alphagov/digitalmarketplace-apiclient.git@3.7.0#egg=digitalmarketplace-apiclient==3.7.0
+git+https://github.com/alphagov/digitalmarketplace-utils.git@21.4.0#egg=digitalmarketplace-utils==21.4.0
+git+https://github.com/alphagov/digitalmarketplace-content-loader.git@1.2.0#egg=digitalmarketplace-content-loader==1.2.0
```
allait: This will roll out the new specialists description for brief responses. @idavidmcdonald is there anything else that needs to be done for that to go live?

allait:
Everything looks good. Make sure to check with @idavidmcdonald if there's anything else needed for the content-loader upgrade and once that's sorted this should be good to go 👍 

Excellent work Chris! 🌟

app/templates/buyers/brief_publish_confirmation.html
```diff
@@ -34,20 +34,39 @@
...
 
     {% if not published and unanswered_required > 0 %}
-      <p class="padding-bottom-small"><strong>You still need to complete the following questions before your requirements can be published:</strong></p>
+      <div class="dmspeak">
+        <p><strong>You still need to complete the following questions before your requirements can be published:</strong></p>
+      </div>
       {% for section in sections %}
         {% for question in section.questions %}
-          {% if question.answer_required %}
+          {% if question.answer_required and not question.id == 'requirementsLength' %}
```
pcraig3: Just confirming that this is what we want.

### Requirements Length + 1 other question to answer
screen shot 2016-07-29 at 08 45 51

### Requirements length is the only question left to answer
screen shot 2016-07-29 at 08 46 20

***

We think this looks dumb, right?

### Requirements length is the only question left to answer
screen shot 2016-07-29 at 08 53 49

tests/app/views/test_buyers.py
pcraig3: So when does it come as a `None` out of the API?

I kind of thought it would either:
- come back with a string value
- not be a key in the `brief` data

requirements.txt
idavidmcdonald: Nope, all good!

tests/app/views/test_buyers.py
Wynndow: From [here](https://github.com/alphagov/digitalmarketplace-api/blob/1_week_or_2/app/models.py#L1016). An object is built to pass into get_publishing_dates function. If `requirementsLength` isn't set in the data blob of a brief, the `requirementsLength` is the new object is `None`

app/templates/buyers/brief_publish_confirmation.html
Wynndow: Yeah, that's what we're after. We decided repeating the message that they need to set the length looked odd. That's why in the view for this page there's a new bit of code about decreasing the unanswered questions count by one if requirementsLength isn't set.

tests/app/views/test_buyers.py
pcraig3: Ah, right.  Cheers.

app/templates/buyers/brief_publish_confirmation.html
pcraig3: Yeah, looks kind of ugly but I guess it works. 

As @risicle would say: "F*ck it, the tests pass."

Wynndow: Ugly as sin, but yep, what risicle said...

PR #366, opened 2016-07-29
[#119728323] Make sure withdrawn briefs can not be accessed

idavidmcdonald:
With a new status being introduced that allows a brief's status to be "withdrawn", we wish to make sure that withdrawn briefs are no longer accessible to users. Behaviour we want to happen across our app's is described in [this comment](https://www.pivotaltracker.com/story/show/119728323/comments/145270949) in the relevant pivotal story.

Most pages at the moment would not show the user a withdrawn brief and would instead 404, however there are a couple of routes that would show the brief and not 404:

`view_brief_overview `
`view_brief_responses`

Instead of adding a specific check to each view, we will add the check to the widely used `is_brief_correct` function. This should remove the responsibility that a developer must remember to check if a brief is withdrawn or not.

Note, `is_brief_correct` is used for when users related to the brief are interacting with the brief i.e. viewing their brief's responses. This function is not used for parts of the app such as the marketplace showing all briefs but these parts of the app already have sufficient brief status checks and so have remained unchanged.

### DIFF
--- a/app/helpers/buyers_helpers.py
+++ b/app/helpers/buyers_helpers.py
@@ -18,9 +18,12 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, status=None
 
 
 def is_brief_correct(brief, framework_slug, lot_slug, current_user_id):
-    return brief['frameworkSlug'] == framework_slug and \
-        brief['lotSlug'] == lot_slug and \
-        is_brief_associated_with_user(brief, current_user_id)
+    return (
+        brief['frameworkSlug'] == framework_slug
+        and brief['lotSlug'] == lot_slug
+        and is_brief_associated_with_user(brief, current_user_id)
+        and not brief_is_withdrawn(brief)
+    )
 
 
 def is_brief_associated_with_user(brief, current_user_id):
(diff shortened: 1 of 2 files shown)


app/helpers/buyers_helpers.py
```diff
@@ -20,7 +20,8 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, status=None
 def is_brief_correct(brief, framework_slug, lot_slug, current_user_id):
     return brief['frameworkSlug'] == framework_slug and \
         brief['lotSlug'] == lot_slug and \
-        is_brief_associated_with_user(brief, current_user_id)
+        is_brief_associated_with_user(brief, current_user_id) and \
```
allait: If you wrap the entire expression in a pair of parenthesis then there's no need for escaping the newlines. This is the [advice PEP8 gives](https://www.python.org/dev/peps/pep-0008/#id19) (although it looks like the linter doesn't catch this):

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

```python
return (
     ...
     and ...
)
```

allait:
Small note about line wrapping, but looks good to merge either way 👍

app/helpers/buyers_helpers.py
idavidmcdonald: Cheers, didn't know about this!

PR #367, opened 2016-08-01
Always use https:// protocol for bower dependencies

allait:
git:// is unauthenticated (and we don't pin with commit SHA) so https:// should be preferred.

Since we're downloading from GitHub we need to add `.git` to access the Git endpoint over HTTPS.

/cc @quis


tombye:
No brainer: 👍

PR #368, opened 2016-08-01
[#125881031]Brief filtering

risicle:
https://www.pivotaltracker.com/story/show/125881031

Still left to do for this:

- [x] tests
- [ ] ~~analytics hook-in~~ (this has been split into story https://www.pivotaltracker.com/story/show/127171827)

but I'm posting this now so people can take a bit of a look at it.

Firstly, one on my commits makes a bunch of comments on the gcloud search view. I _really_ think we need more comments in the code - it took me quite a while to decode what this view was doing and I don't _really_ want to have to do that again.

Looking at the various views that implement this sort of thing, I found most views using different patterns of their own, and none of them particularly extensible to what I needed to do. So following agreement from @pcraig3 I brought in a pattern I've used very successfully in previous projects. WTForms are very powerful and it doesn't look like we're actually using them that much. So what I've done is implement quite a bit of the logic of this view in the _form_ itself. This is quite powerful, because it not only means that the logic could easily be re-used from multiple views if necessary, but the logic can usually be extended through class inheritance.

(aside: How we actually used to use them was with a `.filter()` method where `get_briefs()` is which would return a django queryset (similar to a sqlalchemy query, basically a chain-filter()-able query representation) which could be further filtered in superclasses overriding and building on the `.filter()` method. This is less doable without the ORM at our fingertips, but perhaps the closer analogy would be `get_briefs()` actually being a method returning the kwargs to get passed to `find_briefs()` - I don't know... )

Now this may look like "overkill" for a single view, but the idea is the approach could probably be extended to cope with the service search and others too, giving us a path to possibly merging a lot of their "boring" code together.

Also note the slight oddness of `BriefSearchForm` taking a `data_api_client` as an constructor kwarg - this is to make the `mock.patch`ing of the tests work as-is. The current way the tests are `mock.patch`ing the `data_api_client`, we can only ever use the `data_api_client` from within the relevant view module. I know how we can fix this (make users of the `data_api_client` use it through the `current_app`'s reference to it and just pass a mock api client in to it at app construction time when we're doing tests...) but that's out of scope for this story.


app/main/views/marketplace.py
```diff
@@ -90,21 +92,40 @@ def get_brief_by_id(framework_slug, brief_id):
...
-    page = request.args.get('page', default=1, type=int)
     framework = data_api_client.get_framework(framework_slug)['frameworks']
-
     if not framework:
         abort(404, "No framework {}".format(framework_slug))
 
-    api_result = data_api_client.find_briefs(status='live,closed', framework=framework_slug,
-                                             page=page, human=True)
+    # disabling csrf protection as this should only ever be a GET request
+    form = BriefSearchForm(request.args, framework=framework, data_api_client=data_api_client, csrf_enabled=False)
```
TheDoubleK: I think csrf protection is only enabled in the global config for POST requests to the app, so it might not be necessary to disable it here.

TheDoubleK:
I agree that some comments here are useful, but if we are going to have comments in our code it would be nice for them to be consistently written, and as concise as possible.

Some sentences in the comments here get full stops, others don't.  Some get capital letters, others don't... if we're committing comments then I think they need as much :heart: love :heart: as the code itself.  I don't think phrases like `we have to work backwards a bit here` add to the usefulness of the comments, they just make them break over more lines than necessary.

app/main/views/marketplace.py
risicle: Yeah I think the issue is though that the form doesn't know that it doesn't need to look for a `csrf_token` field because all it knows is it's getting a `MultiDict`, not whether it was a `GET` or `POST`.

risicle:
(bleedin' pep8)

app/main/forms/brief_forms.py
```diff
@@ -0,0 +1,91 @@
+from flask_wtf import Form
+from wtforms import IntegerField, SelectMultipleField
+from wtforms.validators import NumberRange
+
+
+class BriefSearchForm(Form):
```
allait: I've probably missed the discussions about this, but I'm curious why aren't we reusing the way we're doing filters for G-Cloud. We have a way to define forms using manifests and the logic to group options based on question type and it feels like it would be a more extensible approach, especially if we start filtering by brief fields in the future.

app/main/forms/brief_forms.py
```diff
@@ -0,0 +1,91 @@
...
+    status = SelectMultipleField("Status", choices=(
+        ("live", "Live",),
+        ("closed", "Closed",)
+    ))
+    # lot choices expected to be set at runtime
+    lot = SelectMultipleField("Category")
+
+    def __init__(self, *args, **kwargs):
+        """
+            Accepts extra keyword arguments:
```
TheDoubleK: Should this be `requires` rather than `accepts`?  Neither of these args is optional.

app/main/forms/brief_forms.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        super(BriefSearchForm, self).__init__(*args, **kwargs)
+        try:
+            # popping this kwarg so we don't risk it getting fed to wtforms default implementation which might use it
+            # as a data field if there were a name collision
+            framework = kwargs.pop("framework")
+            self._framework_slug = framework["slug"]
+            self.lot.choices = tuple((lot["slug"], lot["name"],) for lot in framework["lots"] if lot["allowsBrief"])
+        except KeyError:
+            raise TypeError("Expected keyword argument 'framework' with framework information")
```
TheDoubleK: Why a TypeError rather than the original KeyError? This doesn't seem to be a problem with object types?

TheDoubleK:
The checkboxes here don't look like the ones we have for G-Cloud service filtering:
G-Cloud:
screen shot 2016-08-02 at 12 17 29

Briefs:
screen shot 2016-08-02 at 12 17 39

app/main/forms/brief_forms.py
risicle: I was talking about this with @pcraig3.

The first point is that there's not really... anything... to fetch from the content loader for this story. The human-readable lot names come from the `get_framework` call and the status strings are `Live` and `Closed`. So I thought dragging pointlessly tiny things from the content loader would get me slapped with the "overkill" label.

On the other hand, the gcloud search method doesn't _really_ have a way of dealing with fields etc. that _aren't_ sourced from the content loader. So I would have had to extend the gcloud mechanisms to cope with these - however, if we're possibly going to take the next step in filtering these and throw the briefs into the search api along with other filterable parameters, it would then make that addition redundant. This (the way I've done it), well you _could_ just chuck this away at that point if you decided.

Thirdly, if we're going to choose one approach to be able to "extend" to cover all cases, from a maintainability point of view I would *strongly* argue in favour of basing it on top of a proven, object-oriented, heavily tested & highly documented forms implementation like wtforms rather than a home-baked solution which is not really object-oriented and already a little hard for a newbie to decode. I'm confident wtforms would be able to build its fields from the content loader if necessary.

And the lovely thing about object-oriented forms implementations like wtforms is they have extensibility in their very nature - you can use inheritance to plug in & out bits of functionality as you like. You have to get you hands a lot dirtier with procedural code...

app/main/forms/brief_forms.py
risicle: Yeah, I know, I'd like them to be written-out too but python won't let you do an in-place kwarg default after an `*args`. So it's either do it like this or don't accept positional parameters. And people like passing in positional parameters for the form data.

app/main/forms/brief_forms.py
risicle: _Technically_ I'm considering it a call-signature error, which is supposed to be a `TypeError`. But... it's not hugely important, just might be a bit surprising for the caller to get a `KeyError` out of this.

risicle:
Removed `we have to work backwards a bit here` but comment still line-spills, take anything else away and it makes less sense.

app/main/forms/brief_forms.py
risicle: My argument is basically: google how to get wtforms to do xyz and you get about 100 results - you can't get the same thing from homebrew solutions.

app/main/forms/brief_forms.py
```diff
@@ -0,0 +1,91 @@
+from flask_wtf import Form
+from wtforms import IntegerField, SelectMultipleField
+from wtforms.validators import NumberRange
+
+
+class BriefSearchForm(Form):
+    page = IntegerField(default=1, validators=(NumberRange(min=1),))
+    status = SelectMultipleField("Status", choices=(
+        ("live", "Live",),
```
TheDoubleK: Somewhat confusingly, I think the public-facing label for `live` briefs should now be `Open` ¯\_(ツ)_/¯

Whether or not this is something we want to change in the API model to stop ourselves from going mad, I don't know... not for this story though.

TheDoubleK:
Tested and reviewed - generally looks good and works nicely, definitely worth putting in front of Ralph and/or Sarah if you haven't already.

I think I'm generally won over by the argument that we aren't looking into the data blob at all for these filters, which makes the content loader less useful in this case, but would be good to hear what @allait thinks.

risicle:
Changed "Live" status -> "Open", checked for more stuff in tests.

app/main/forms/brief_forms.py
allait: >Thirdly, if we're going to choose one approach to be able to "extend" to cover all cases, from a maintainability point of view I would strongly argue in favour of basing it on top of a proven, object-oriented, heavily tested & highly documented forms implementation like wtforms rather than a home-baked solution which is not really object-oriented and already a little hard for a newbie to decode. I'm confident wtforms would be able to build its fields from the content loader if necessary.

I think it's important to understand that you're making a choice in the context of an existing project. This is not the first form we're building and it's not the first filter we're building, so the question is not "what technology would I personally prefer" it's "what makes sense for this project and this team".

And I think that the important part is that we have dozens of forms driven by content-loader and frontend-toolkit, including one working implementation of filters. It is proven, it is tested, and the team as a whole has more experience with it than with wtforms.

Setting that aside, I'd rather we have one approach to filter implementation than two. It feels like adding lot and      status questions to the frameworks is not really a blocker, since we've done this for G-Cloud before. And the argument for extending it to filter on other brief fields is less abstract than it is with wtforms – we have a working example.

I'd love to hear what everyone else thinks (esp @tombye as he knows the G-Cloud filters much better than I do), but my position is: unless there's a lot more work involved in going with the content-loader route (or if we think it'd be a good idea to use wtforms for G-Cloud filters) I think we should make two implementations as similar as possible.

tests/app/views/test_marketplace.py
```diff
@@ -477,25 +479,123 @@ def setup(self):
...
+
+        normalize_qs = lambda qs: {k: set(v) for k, v in iteritems(parse_qs(qs)) if k != "page"}
+        assert normalize_qs(parsed_original_url.query) == \
+            normalize_qs(parsed_next_url.query) == \
+            normalize_qs(parsed_prev_url.query)
+
+        ss_elem = document.xpath("//p[@class='search-summary']")[0]
+        assert self._normalize_whitespace(self._squashed_element_text(ss_elem)) == "2 results"
+
+    def test_catalogue_of_briefs_400_if_invalid_status(self):
```
TheDoubleK: This should be a 404 rather than 400 - there's nothing wrong with the HTTP request itself - we can understand what's being asked for - it's just that we don't currently have anything with that status.

app/main/forms/brief_forms.py
risicle: Hold up a bit - I'm not even touching on the idea of replacing any of the 1001 uses of content loaded forms in the system - I'm just talking about the search/filtering forms, which _from what I could tell_ are actually implemented in a fairly separate way.

When it comes to this particular use case, neither of the existing methods _quite_ covered this use case, so whatever I did I would have to "break new ground" with this code. And if in two months it's decided to extend the brief search to be search-api based, sure, if it's decided the other way's better (or just lower immediate effort), this can be binned. At least this being here would be a strawman example of how such a form could operate.

And it's not really about _personal preference_, it's more that if I write code and commit it, I'm putting my personal seal-of-approval on it, and having seen so many subtle mistakes being made (and having made _so many_ mistakes myself) doing manual form processing, I'd find it a lot trickier to do that if I just mangled the existing methods.

("Personal seal-of-approval" sounds more pompous than I mean it - it's just an "I think this works ok" which I'd like people to be able to trust to a degree)

Unfortunately @tombye has just gone on hols...

risicle: > Setting that aside, I'd rather we have one approach to filter implementation than two. It feels like adding lot and status questions to the frameworks is not really a blocker, since we've done this for G-Cloud before. And the argument for extending it to filter on other brief fields is less abstract than it is with wtforms – we have a working example.

(looking further into this now...)

risicle: It strikes me that the _biggest_ difference (that would require the most mangling) between g-cloud's search and the way we want to do dos's filtering is g-cloud's desire to treat lots quite separately, affecting the other options offered, while dos's treatment of lots is quite straightforward. At some point if we want to bring the two into line together this will have to be resolved, and that seemed a bit much for a two-point story.

allait: > I'm just talking about the search/filtering forms, which from what I could tell are actually implemented in a fairly separate way.

This is kind of my point, I'd like to have fewer separate implementations for similar things.  Building up an HTML form, validating, and joining parameters based on question type (checkboxes or radio buttons) feel like enough similarities to justify a shared implementation.

> At least this being here would be a strawman example of how such a form could operate.

We've done the different approach before with frontend validation for supplier declarations and we still haven't managed to make it fit with the rest of the code.

Anyway, I'm 10000 miles away, so I'll leave the decision to you and the team.

risicle: > Building up an HTML form, validating, and joining parameters based on question type (checkboxes or radio buttons) feel like enough similarities to justify a shared implementation.

I think I would agree if it weren't for the way service search treats the lot as a mode-switch parameter, which adds a lot of special-case-ness to that code, which I thought interfering with was beyond the 2-point, just-get-this-out nature of this story.

I actually didn't realize that we only currently use wtforms for login forms - thought it was used for some of the admin forms too, but no.

allait: I don't really understand the `lot` problem to be honest. G-Cloud filters use content-loader to only show filters that apply to current lot, it feels like we should be able to easily ignore this if none of DOS filters specify lot dependencies.

risicle: > I don't really understand the lot problem to be honest

I did a bit of investigation and I see a way I could work around the lot treatment differences. I would however need to screw around with the service search code a little bit to make a shared implementation work (it would need modernization - it still uses hard-coded lot names), but the feeling is to just get this in now and look at unification in a future story.

TheDoubleK:
Needs a rebase again, but let's do it :+1:

app/main/forms/brief_forms.py
pcraig3: Yeah, I'm with @risicle on this one. 

We did a bit of digging through the `/g-cloud/search` endpoint in the search API this morning and there's a bunch of hard-coded `g-cloud-6` stuff in there where it's looking for a manifest of search filters.   Seems like writing this in a more robust way should also mean rethinking where we keep our manifest for which search filters to use, and that's getting a bit beyond what this story really needs to be.

It would be neat to seed the WTForm with content that lives in `digitalmarketplace-frameworks` (instead of hardcoding it into the form), but it seems like if we're going with Rob's stuff right now on the argument that the `g-cloud` search is dissimilar, we shouldn't really be trying to pull the stuff from the content loader into here until we're more aligned on how to merge everything.  At the moment, it all looks pretty straightforward to me.

As a final point, I remember that when David and I were working on the supplier submissions framework submission, we also weren't clear on whether or not to use WTF forms.  In our case, we ended up doing [WTF forms for entering the signer names and the final "I have the authority" form](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/main/forms/frameworks.py) but not for the file uploading bit because there was a bunch of file uploading code that we reused.  So, while we've got the submissions bit pretty much nailed down, I think there's still a bunch of ambiguity around building other forms that don't (or don't have to) involve the content loader.

PR #369, opened 2016-08-01
Quick fix: remove framework link if there is no framework page

TheDoubleK:
First version fix for this story: https://www.pivotaltracker.com/story/show/126767693

There is no point in linking to a generic list of CCS frameworks that doesn't even include the framework just clicked on.

This is a quick fix for G8 going live today.  The proper fix will move the URLs out to the frameworks repo.

### DIFF
--- a/app/assets/scss/_service-page.scss
+++ b/app/assets/scss/_service-page.scss
@@ -52,7 +52,7 @@
 
 #meta {
 
-  .meta-item {
+  %meta-item {
     margin-bottom: $gutter;
   }
   
@@ -64,12 +64,16 @@
     display: none;
   }
 
+  .framework-name {
+    @extend %meta-item;
+  }
+  
   .service-id {
-    @extend .meta-item;
+    @extend %meta-item;
   }
 
   .external-link-default {
-    @extend .meta-item;
+    @extend %meta-item;
   }
 }
 
--- a/app/assets/scss/_service-page.scss
+++ b/app/assets/scss/_service-page.scss
@@ -51,6 +51,11 @@
 }
 
 #meta {
+
+  .meta-item {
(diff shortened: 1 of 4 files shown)


pcraig3:
Seems okay to me, but maybe @tombye has an opinion.

TheDoubleK:
I worked out that the `_service_page` styles are in this app, not the toolkit at all, so fixed the `meta-item` extensions to make more sense.

tombye:
I have two:

1. isn't the `<div>` a paragraph? Seems you want a block-level bit of text to me but it's unclear looking at these changes
2. it looks like you're using `meta-item` as a placeholder as you already have classes using it to apply that block of CSS to a descriptive class name. Shouldn't we be following this and using a class name that describes the thing in the HTML? Could it be `service-id`?

TheDoubleK:
1. I followed the existing pattern on the page - the linked version of the framework name, service IDs and contact details are all in `<div>`s.  Seems like this should follow the same pattern?  Or do you mean there should be a `<p>` inside the enclosing `<div>`?
2. Do you mean there should be a new style that extends meta-item for the new unlinked framework name?  Seems like overkill but I can do if you think it helps?

tombye:
What I failed to explain was I think the existing pattern was implemented badly.

Really we want all children of `<div class="meta">` to be block-level elements. Some, like `<div class="contact-details">` should be `<div>`'s because they're a collection of `<p>`'s already but others, like `<div class="service-id">` and `<div class="external-link-default">` just contain inline elements so should really be `<p>`'s.

TheDoubleK:
I've fixed the new element here to be a <p>, but the rest come from frontend toolkit elements that we should change in future, but not right now.

tombye:
OK by me 👍

PR #371, opened 2016-08-02
PROD BUGFIX: show all service ID digits for 15-digit IDs

TheDoubleK:
Fix for this bug: https://www.pivotaltracker.com/story/show/127547401

Previous G-Cloud iterations had 16-digit service IDs, which are displayed on the service page as four chunks of four digits.
For G-Cloud 8 we have moved to 15-digit IDs to help people using spreadsheets, but the method used to "chunk" the IDs for display on the service page dropped the final three digits as they did not match the regular expression that was being used to split the ID (`r'....'`).

This introduces a new method `chunk_string` that returns a generator for chunks of a specified length.  Unlike the old regular expression, this generator also returns smaller chunks at the end of the string, so that 15-digit IDs display as three chunks of four plus three digits at the end (e.g. 1111 2222 3333 444).

BEFORE
----------
screen shot 2016-08-02 at 17 11 38

AFTER
--------
screen shot 2016-08-02 at 17 10 25

### DIFF
--- a/app/helpers/shared_helpers.py
+++ b/app/helpers/shared_helpers.py
@@ -25,3 +25,7 @@ def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_o
         for framework in frameworks:
             if framework.get('status') == status:
                 return framework
+
+
+def chunk_string(string, chunk_length):
+    return (string[0+i:chunk_length+i] for i in range(0, len(string), chunk_length))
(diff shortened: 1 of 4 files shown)


idavidmcdonald:
Looks good. My only question would be should `chunk_string` return it's result in a list instead of having us call `list` every time it's being run?

I'll leave you to decide, either way is fine with me. 👍

TheDoubleK:
I did think about doing that, but it would just be calling the same `list()` every time at a slightly earlier point.  In terms of future reusability (despite the name the method can potentially be used to chunk any iterable at all, not just strings) I think having it return the generator is a bit nicer.

PR #372, opened 2016-08-02
Get framework URLs from frameworks content instead of hard-coded in app

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/126767693

Depends on:
 - [x] https://github.com/alphagov/digitalmarketplace-frameworks/pull/297

### DIFF
--- a/app/presenters/service_presenters.py
+++ b/app/presenters/service_presenters.py
@@ -1,9 +1,10 @@
-from ..helpers.shared_helpers import chunk_string
 import os
 import re
-from dmutils.service_attribute import Attribute
-from dmcontent.formats import format_service_price
 
+from app import content_loader
+from dmutils.service_attribute import Attribute
+from dmcontent.errors import ContentNotFoundError
+from dmcontent.formats import format_service_price
 try:
     from urllib.parse import urlparse
 except ImportError:
@@ -12,6 +13,7 @@ try:
     from urllib import unquote
 except ImportError:
     from urllib.parse import unquote
+from ..helpers.shared_helpers import chunk_string
 
 
 class Service(object):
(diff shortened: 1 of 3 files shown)


app/presenters/service_presenters.py
```diff
@@ -93,14 +96,17 @@ def get_service_id(self, service_data):
...
-            framework_url = None
-        return framework_url
+        try:
+            content_loader.load_messages(service_data['frameworkSlug'], ['urls'])
+            framework_url = content_loader.get_message(
+                service_data['frameworkSlug'],
+                'urls',
+                'framework_url'
+            )
+            return framework_url if framework_url else None
```
risicle: This can just be done as:

```
return content_loader.get_message(
    service_data['frameworkSlug'],
    'urls',
    'framework_url'
) or None
```

but the difference is mostly stylistic.

risicle:
This gets a soft-thumbs-up from me but I don't feel I know enough about the content loader yet to pull the big lever...

app/presenters/service_presenters.py
TheDoubleK: Yep that's a bit nicer.  Needs a rebase anyway so will fix up.

pcraig3:
So we're not linking to the G-Cloud 6 framework agreement anymore?  

Is there a chance that someone could still need to see the contract from the service page even if they can't buy the service?

(I realise this comment should be in the other repo, but here we are.)

TheDoubleK:
The G-Cloud 6 link, like the G-5 one, has been made `FORBIDDEN` by CCS, so no point showing it any more.

pcraig3:
`FORBIDDEN` sounds serious.

👍  when you're ready.

TheDoubleK:
Travis really struggling with NPM today.

TheDoubleK:
Cheers Paul, going to park this here until the CCS page is live.

PR #374, opened 2016-08-03
[#127062871] Track account creation using GA virtual page views

idavidmcdonald:
Currently, we are not able to tell using GA how many user accounts have been created. We could try to use a page view of the URL after an account is created to indicate this, however the URL is also accessible from other user flows and will not therefore be an accurate count of how many user accounts have been created.

Therefore, as part of this pivotal story, we wish to send a GA virtual page view every time a buyer account is successfully created. By embedding

`data-analytics="trackPageView" data-url="/vpv/?account-created=true"`

into our template, we will send the required virtual page view on the success page. This behaviour [already exists](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/assets/javascripts/analytics/_virtualPageViews.js) in the supplier front end but did not yet exist in the buyer front end. I have added it here but this may be something that should be moved into the front end toolkit as there is some code duplication?

This tracking tag will only be embedded if they have come through the account creation process (indicated by setting a flash message).

Note, a similar process [has been created](https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/517) for the creation of supplier accounts.

### DIFF
--- a/app/assets/javascripts/_analytics.js
+++ b/app/assets/javascripts/_analytics.js
@@ -1,6 +1,5 @@
 //= include ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/google-analytics-universal-tracker.js
 //= include ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/analytics.js
-//= include analytics/_dm-analytics.js
 //= include analytics/_register.js
 //= include analytics/_pageViews.js
 //= include analytics/_events.js
--- a/app/assets/javascripts/_analytics.js
+++ b/app/assets/javascripts/_analytics.js
@@ -4,4 +4,5 @@
 //= include analytics/_register.js
 //= include analytics/_pageViews.js
 //= include analytics/_events.js
+//= include analytics/_virtualPageViews.js
 //= include analytics/_init.js
--- a/app/assets/javascripts/analytics/_init.js
+++ b/app/assets/javascripts/analytics/_init.js
@@ -5,5 +5,6 @@
     this.register();
     this.pageViews.init();
     this.events.init();
+    this.virtualPageViews();
   };
 })(window);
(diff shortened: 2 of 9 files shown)


pcraig3:
I've been testing this by printing out arguments to the `window.ga` object into the console,(example code below)

```js
// app/assets/javascripts/_analytics.js

window.ga = function() { console.log(arguments); };
//= include ... 

```
but it wasn't appearing to send the right thing back, which was odd, because the HTML was definitely correct and the JavaScript tests were working as expected.

Problem was that the `virtualPageViews()` function was never getting called, [which it is in the supplier app](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/ddeb9b26f2893aaf3df3c0427011c297e7baf0d3/app/assets/javascripts/analytics/_init.js#L7).
The JS tests call it directly, so that's why they were working even though I couldn't duplicate it in the app.

There's also a couple of imports looking for a deleted file (here are the [two](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/d2ed6f338044f57629e843ac292c5b7f3630fbba) [commits](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/5dafe20e23d412390484294679b484dde8078e2b) we have in the supplier app remove them, so you should probably do that here.  (You could also do them in your supplier app pull request if you wanted to anticipate the change.)

Code looks good though, as well as the javascript tests you've written.

pcraig3:
Yeah, awesome. Looks 👍  to me.

pcraig3:
Worth adding (just saw this out now in the supplier app and it's happening here too), that we seem to be sending two pageviews back to google when we do supplier account creation.

screen shot 2016-08-04 at 18 02 22

In order, those calls:
1. create a tracker 
2. anonymize the users ip address
3. send the normal pageview that we always send when we hit the homepage
4. send a virtual pageview for the first time a new buyer is logged in

If Ash is fine with it, then it should be fine.  Otherwise, maybe intercept the normal pageview if a virtualpageview is triggered somewhow?

I dunno.  It seems like it's probably fine.

idavidmcdonald:
Just spoke to Ash and he is happy with the two page views

pcraig3:
👍

PR #378, opened 2016-08-05
Fix bug on publish page with required questions

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/127797513) bug on Pivotal.

When the buyer frontend was updated to introduce one or two week
specialist briefs, the question was referring to briefs as being 'live'
and not 'open'. This was changed to 'open' instead of 'live' after the app was released which broke
the code which removed that particular question from the count of
unanswered questions. This update is needed to fix it.


TheDoubleK:
It might be nice to have a test for when this message gets shown or not.

Wynndow:
Cheers Kev, have added that test in.

idavidmcdonald:
👍  after rebase

PR #379, opened 2016-08-08
[#125881031]brief filtering template: put page heading, marketplace-paragraph in a 2/3 column as it should be

risicle:
as per ralph's request.

Story https://www.pivotaltracker.com/story/show/125881031


ralph-hawkins:
👍

PR #380, opened 2016-08-08
[#127171827] Set Google Analytic custom dimensions for filtering opportunities

idavidmcdonald:
We added basic filtering to DOS opportunities in [this story](https://www.pivotaltracker.com/story/show/125881031). We wish to add some basic analytics to this new functionality, details can be found [in this analytics story](https://www.pivotaltracker.com/story/show/127171827).

We add custom dimensions for
- The number of opportunities found
- Which filter groups were used e.g. the lot type filter group
- When filtering by lot, which exact filters were used e.g. digital-specialists
- When filtering by status, which exact filters were used e.g. open

The custom dimension must be sent before the page view. ID's for each custom dimension have been set up by Ash.

For the filter and filter group dimensions, we wish to return the values in an alphabetically sorted list, separated by "|", e.g. "digital-outcomes|digital-specialists". This means Ash will have more consistent results.

### DIFF
--- a/app/assets/javascripts/analytics/_pageViews.js
+++ b/app/assets/javascripts/analytics/_pageViews.js
@@ -1,7 +1,18 @@
 (function (GOVUK) {
   GOVUK.GDM.analytics.pageViews = {
     'init': function () {
+      this.setCustomDimensions();
       GOVUK.analytics.trackPageview();
+    },
+
+    'setCustomDimensions': function() {
+      // check that we're on the catalogue page for opportunites
+      if(GOVUK.GDM.analytics.location.pathname() === "/digital-outcomes-and-specialists/opportunities") {
+
+        // if it does, we want to send the current number of opportunites back to GA
+        numberOfOpportunites = $('.search-summary-count').text();
+        GOVUK.analytics.setDimension(21, numberOfOpportunites);
+      }
     }
   };
 })(window.GOVUK);
--- a/app/assets/javascripts/analytics/_pageViews.js
+++ b/app/assets/javascripts/analytics/_pageViews.js
@@ -7,8 +7,7 @@
 
     'setCustomDimensions': function() {
       var search = GOVUK.GDM.analytics.location.search(),
-          categoryParams = [],
-          statusParams = [],
+          filterGroupDimension = [],
           numberOfOpportunites,
           dimensions,
           pairs,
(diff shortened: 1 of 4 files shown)


risicle:
Would be nice to rebase this to check the tests still pass against new master

idavidmcdonald:
@risicle tests all green after the rebase

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,75 @@
...
+          filterGroupDimension = [],
+          numberOfOpportunites,
+          dimensions,
+          pairs,
+          pair,
+          i,
+          j;
+
+      // check that we're on the catalogue page for opportunites
+      if (GOVUK.GDM.analytics.location.pathname() === "/digital-outcomes-and-specialists/opportunities") {
```
risicle: I know this is a broader thing and partly built into the _way we do things_, but hardcoding the exact url path here isn't tremendously DRY (don't repeat yourself) - changing an url route will need to be reflected here to be picked up properly (and will lead to a silent failure if someone forgets to do it).

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,75 @@
...
+              paramArray:  [],
+            },
+            {
+              dimensionId: 24,
+              paramType:  'status',
+              paramArray:  [],
+            },
+          ]
+
+          j = dimensions.length; while (j--) {
```
pcraig3: Wow! This is a really cool way to iterate through stuff.

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -1,7 +1,75 @@
...
+              filterGroupDimension.push(dimensions[j]['paramType']);
+            }
+          }
+
+          if (filterGroupDimension.length) {
+            filterGroupDimension.sort();
+            GOVUK.analytics.setDimension(22, filterGroupDimension.join('|'));
+          }
+
+          function setFilterDimension(pairs, dimension) {
```
pcraig3: Probably leaving a comment in front of this function so we know what it's doing.  I was there and I still find it a bit cryptic.

app/assets/javascripts/analytics/_pageViews.js
pcraig3: Sure thing.  Let's ask @tombye when he gets back about hardcoding urls in JS, [as it seems to be what we've gone with before](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/0f3613a0f2cfdd67d78bc91b8e2ae3f456dd49c1/app/assets/javascripts/analytics/_virtualPageViews.js#L16).

pcraig3:
Good to merge, I think, after describing what that other function does. 

<sub>preemptive</sub> 👍

PR #381, opened 2016-08-09
[#125881031]Brief filtering: fix count noun

risicle:
Story https://www.pivotaltracker.com/story/show/125881031

Result count noun should be "results" if all options are selected.

Two changesets - the first adds a test that can verify the behaviour, the second changes the behaviour and expectation of the test.


pcraig3:
👍  2 good ~~commits~~ results here.

PR #382, opened 2016-08-10
Fix wrong url on publish page

pcraig3:
After changing the text from 
`set how long your requirements will be live for` 
to 
`set how long your requirements will be open for`, 
the url we used to get to the requirementsLength page changed with it.

Since we've been treating the requirementsLength question in a special way where we don't use the rest of the content-loaded stuff, we're actually hard-coding the url string into the publish page. Unfortunately, when we updated the content, this broke the link because we didn't update the link along with it.

This pull request fixes the broken link, but it doesn't really address the problem of hardcoding links into our template.

Not sure there is a solution for that, incidentally.


idavidmcdonald:
👍 after travis.

Also very good clear pull request message 💯

pcraig3:
Thanks! 👍

PR #383, opened 2016-08-11
Remove conditional links on home page

TheDoubleK:
DOS is now live so we don't need the conditional sections that had to magically appear when we first launched DOS.

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -16,35 +16,6 @@
 {% block main_content %}
 
 {% set dos_slug = 'digital-outcomes-and-specialists' %}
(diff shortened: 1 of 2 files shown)


idavidmcdonald:
This won't take into account when DOS becomes 'expired' but I very much assume at the moment we don't care about that? (nor do I know when DOS is actually due to expire?) So 👍

TheDoubleK:
At the point of DOS expiring there will be DOS-2 stuff going live, so probably a new conditional on the `dos_slug` variable.

TheDoubleK:
And if that's not happening then we'll need a complete re-design of the homepage. (Which might happen before then anyway.)

PR #384, opened 2016-08-18
[#128509823] Suppliers should be able to view their response for any brief state

TheDoubleK:
Part 1 bug fix for this: https://www.pivotaltracker.com/story/show/128509823

The "View your response" link was only being shown on the opportunity page while the opportunity is live.  Suppliers need to revisit their responses after the opportunity has closed too.

This will need to be released AFTER the supplier app fix (coming soon!), as otherwise the shiny new link here will 404.

### DIFF
--- a/tests/app/views/test_marketplace.py
+++ b/tests/app/views/test_marketplace.py
@@ -456,13 +456,29 @@ class TestBriefPage(BaseApplicationTest):
 
         self._assert_start_application(document, brief_id)
 
-    def test_supplier_applied_view_application(self):
+    def test_supplier_applied_view_application_for_live_opportunity(self):
         self.login_as_supplier()
         # mocking that we have applied
         self._data_api_client.find_brief_responses.return_value = {
             "briefResponses": [{"lazy": "mock"}],
         }
         brief_id = self.brief['briefs']['id']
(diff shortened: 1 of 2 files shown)


tests/app/views/test_marketplace.py
```diff
@@ -469,6 +469,24 @@ def test_supplier_applied_view_application(self):
...
 
+    def test_supplier_applied_view_application_for_closed_opportunity(self):
+        self.login_as_supplier()
+        # mocking that we have applied
+        self._data_api_client.find_brief_responses.return_value = {
+            "briefResponses": [{"lazy": "mock"}],
+        }
+        brief = self.brief.copy()
+        brief['briefs']['status'] = "closed"
+        brief['briefs']['publishedAt'] = "2000-01-25T12:00:00.000000Z"
```
idavidmcdonald: "publishedAt" and "applicationsClosedAt" lines appear to be unnecessary (tests pass without them). The "status" should be enough to indicate a closed brief.

idavidmcdonald:
👍 
2 lines can be removed but apart from that it looks good

tests/app/views/test_marketplace.py
TheDoubleK: Yep that was some lazy copy-pasting from a different test for closed briefs.  Have removed them now.

PR #389, opened 2016-08-30
Update apiclient to latest

pcraig3:
Updating the apiclient lets us get rid of a bunch of dead code in the API.

The jump from 5.x to 6.x gets rid of the `update_brief_status` method in favour of `publish_brief`.

[See here for more info](https://github.com/alphagov/digitalmarketplace-apiclient/pull/31).


idavidmcdonald:
👍 

Commit message could be more descriptive instead of using the word "latest" but apart from that it's good after a rebase

PR #390, opened 2016-08-31
Update frameworks to 2.3.4

pcraig3:
Minor change: makes a few h1 headings into h2s so that our page is more semantic and that our functional tests run.

✨  tested locally and it works ✨


ralph-hawkins:
once it turns green 👍

PR #394, opened 2016-10-21
Feature flag for new supplier flow

Wynndow:
For [this story on Pivotal](https://www.pivotaltracker.com/story/show/129838641).

This feature flag will render a new message and button before a supplier starts applying for a brief when turned on. It's currently only set in the base config and can be turned on in different environments when needed.

More feature flags will be needed in the supplier app, and the api when the actual new flow is being developed. This flag is in the buyer app, and clicking the button to start an application takes you to the supplier app. Therefore this flag can only be used for this message, which is a bit weird.

Feature flag not activated:
screen shot 2016-10-21 at 14 38 21

Feature flag activated:
screen shot 2016-10-21 at 14 37 34

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -70,16 +70,36 @@
     </div>
   </div>
 {% elif brief.status == 'live' %}
-  <div class="grid-row">
-    <div class="column-one-third">
-      {% with
-         url = "/suppliers/opportunities/{}/responses/create".format(brief.id),
-         label = "Start application"
-      %}
-        {% include "toolkit/link-button.html" %}
-      {% endwith %}
+  {% if 'NEW_SUPPLIER_FLOW' is active_feature %}
+    <div class="grid-row">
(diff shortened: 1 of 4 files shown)


app/templates/brief.html
```diff
@@ -70,16 +70,36 @@
...
+        <p class="dmspeak">To apply, you must give evidence for all the essential and nice-to-have skills and experience you have.</p>
+      </div>
     </div>
-  </div>
+    <div class="grid-row">
+      <div class="column-one-third">
+        {% with
+           url = "/suppliers/opportunities/{}/responses/create".format(brief.id),
+           label = "Apply",
+           advice = True
```
idavidmcdonald: Optional improvement but could maybe do with a small comment here to say what advice is doing (group this button with the above advice).

idavidmcdonald:
Nice work 👍

PR #395, opened 2016-10-24
Change two flows feature flag to true for QA

Wynndow:
This is turning on the feature flag from a recent PR [here](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/394).

This is so it can be QA'd by Andrew. It will need switching off again with another PR before the buyer app is released to staging/prod.

### DIFF
--- a/config.py
+++ b/config.py
@@ -63,7 +63,7 @@ class Config(object):
 
     # Feature Flags
     RAISE_ERROR_ON_MISSING_FEATURES = True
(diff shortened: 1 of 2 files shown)


andrewchong:
Ok for now but consider changing the test to account for feature being TRUE as well as FALSE, rather than having to change the test each time the flag is changed?

👍

andrewchong:
Looks good to me. 👍

PR #396, opened 2016-10-24
Change feature flag to false

Wynndow:
This feature flag was only turned on for testing on preview. It's needs to be turned off again.


andrewchong:
👍

TheDoubleK:
You can set feature flags on and off per environment, e.g. ON in Preview AND Staging but OFF in Production - see the individual environments towards the bottom of the file, where default values can be overridden per environment.

PR #397, opened 2016-10-26
Turn feature flag on for preview and staging

Wynndow:
For [this story on pivotal](https://www.pivotaltracker.com/story/show/129838641).

Turns the flag on in preview and staging so it can be signed off before it goes live on production.

### DIFF
--- a/config.py
+++ b/config.py
@@ -119,11 +119,11 @@ class Live(Config):
 
 
 class Preview(Live):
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
👍 nice

TheDoubleK:
Might also be worth it for Development?

PR #398, opened 2016-10-27
[#131733667] Content loader to support jinja templating

allait:
This pull request pulls in the latest content loader, frontend toolkit, and content (frameworks repo).
These are all breaking changes, so some fixes were needed.

- always calls `.filter` before displaying section data (this isn't strictly necessary anymore, but it's not a bad idea)
- stop using the `|markdown` filter in the frontend toolkit or the buyer templates
- fix test which expected the old markdown formatting

Everything should work as before, except now jinja templating (processed in the content-loader) will be supported by the buyer frontend.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/131733667)

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -26,7 +26,6 @@ $path: "/static/images/";
 @import "toolkit/_buttons.scss";
 @import "toolkit/_contact-details.scss";
 @import "toolkit/_document.scss";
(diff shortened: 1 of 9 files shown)


TheDoubleK:
All looks good to me, unless there are outstanding issues you know about?
:+1:

PR #399, opened 2016-11-08
Replace in-app beta banner with new toolkit version

pcraig3:
Replaces the in-app beta banner with the new toolkit version.

Also removes some of the classnames for the subheadings on the homepage. 
@tombye might have opinions about that.


tombye:
👍

PR #400, opened 2016-11-09
Fix initial project setup

danielknell:
- added a quickstart section to README
- added a bootstrap.sh for consistency with other projects
- added npm_install to the .PHONY directive in Makefile
- added package.json as dependency of npm_install in Makefile
- made run_app use the virtualenv python
- used DM_ENVIRONMENT to infer a GULP_ENVIRONMENT in order to build development assets on development environments and production assets in all other environments

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ requirements: virtualenv requirements.txt
 requirements_for_test: virtualenv requirements_for_test.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements_for_test.txt
 
-npm_install:
+npm_install: package.json
 	npm install
 
 frontend_build:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -15,11 +22,11 @@ requirements: virtualenv requirements.txt
 requirements_for_test: virtualenv requirements_for_test.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements_for_test.txt
 
-npm_install:
+npm_install: package.json
```
allait: What's the benefit of the dependency? As far as I understand `package.json` is always there and since `npm_install` is a phony step there are no timestamps to compare.

allait: Yeah, good point about the error message, let's keep it.

allait:
👍

PR #403, opened 2016-11-21
Content

fearlesssteven:
Nathan, as well as these copy changes we need to:
1) Resurface the section from V2 which was the list of 'Useful documents' please. (Somewhere down the end before th AusTender question)
2) Add a link to the Deed in the buying process to the copy where it says 'Deed of Standing Offer'.

Thanks!


TheDoubleK:
Pretty sure this was meant as a PR against the AusDTO fork.

PR #404, opened 2016-11-24
remove email from reset password token

leohemsted:
~~with dmutils 22.1.0 (https://github.com/alphagov/digitalmarketplace-utils/pull/287), it's no longer required - removing it means that we no longer store user email addresses unencrypted in logs and such. Yay!~~

encrypt all tokens.

depends on dmutils 23.0.0

- [ ] https://github.com/alphagov/digitalmarketplace-utils/pull/288

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -88,8 +88,7 @@ def send_reset_password_email():
 
             token = generate_token(
                 {
-                    "user": user.id,
-                    "email": user.email_address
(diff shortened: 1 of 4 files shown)


TheDoubleK:
We should do this in supplier (inviting collaborators email) and admin (sending invitations) apps too.

:+1:

pcraig3:
👍  I think.

PR #409, opened 2016-12-02
Add new watcher for frameworks to make framework dev nicer

galund:
- new watcher for frameworks to make framework dev nicer
 - bower now copies only the frameworks YAML directory, not the entire frameworks
   repo, for consistency with the admin frontend


pcraig3:
Works for me!  

Maybe be a bit more specific in the README if you think it'll be helpful, but if you don't think so, it should be good to go as is.

pcraig3:
👍  after a rebase.

PR #412, opened 2016-12-06
Fix rendering of button and message for new supplier brief response flow

Wynndow:
Part of this story: [https://www.pivotaltracker.com/story/show/129843611](https://www.pivotaltracker.com/story/show/129843611)

There was an issue with the logic around deciding whether to render a link button that linked to the `/start` route or the `/create` route when viewing a brief and applying.

Previously, the page would have rendered the link-button pointing at the new `/start` view if the new supplier flow feature flag was active regardless of brief status. This is incorrect. If the flag is active, but a brief was published before it was made active, it should render the link-button pointing at the old `/create` view.

This will only need to be maintained for a couple of weeks after the new flow go's live, then it can be remove.

### DIFF
--- a/tests/app/views/test_marketplace.py
+++ b/tests/app/views/test_marketplace.py
@@ -407,10 +407,14 @@ class TestBriefPage(BaseApplicationTest):
 
     @staticmethod
     def _assert_start_application(document, brief_id):
+        message = document.xpath("//p[@class='dmspeak']/text()")[0]
+
+        assert message == "To apply, you must give evidence for all the essential and nice-to-have " \
+                  "skills and experience you have."
         assert len(document.xpath(
             '//a[@href="{0}"][contains(normalize-space(text()), normalize-space("{1}"))]'.format(
-                "/suppliers/opportunities/{}/responses/create".format(brief_id),
-                "Start application",
+                "/suppliers/opportunities/{}/responses/start".format(brief_id),
+                "Apply",
             )
         )) == 1
 
--- a/tests/app/views/test_marketplace.py
+++ b/tests/app/views/test_marketplace.py
@@ -377,16 +377,6 @@ class TestBriefPage(BaseApplicationTest):
 
         assert_equal(qa_link_text.strip(), "Ask a question")
 
-    def test_can_apply_to_live_brief_with_legacy_supplier_flow(self):
-        self.app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'] = False
-        brief_id = self.brief['briefs']['id']
-        res = self.client.get('/digital-outcomes-and-specialists/opportunities/{}'.format(brief_id))
-        assert_equal(200, res.status_code)
-        document = html.fromstring(res.get_data(as_text=True))
-
-        apply_links = document.xpath('//a[@href="/suppliers/opportunities/{}/responses/create"]'.format(brief_id))
-        assert len(apply_links) == 1
-
     def test_can_apply_to_live_brief(self):
         brief_id = self.brief['briefs']['id']
         res = self.client.get('/digital-outcomes-and-specialists/opportunities/{}'.format(brief_id))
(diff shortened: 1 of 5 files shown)


idavidmcdonald:
Similar to supplier front end, might be good to test the correct button is shown when the feature flag is False.

tombye:
I've a few points, all related to code in `tests/app/views/test_marketplace.py`. They're all associated with code related to this change but not actually changed so I can't comment in-place (sorry).

I think the tests that mention the old flow should be renamed "legacy" and those that mention the new flow to just be the current flow.

For example, `test_can_apply_to_live_brief_with_old_supplier_flow` would be `test_can_apply_to_live_brief_with_legacy_supplier_flow` and `test_can_apply_to_live_brief_with_new_supplier_flow` would be `test_can_apply_to_live_brief`.

This would match how we describe the difference in the supplier app.

`test_cannot_apply_to_closed_brief` tests that the opportunity page is in a closed state when the brief is closed for responses. Seeing as these changes are to introduce the new flow, we should really be checking the bits of the page you use to apply in the new flow are absent. This currently looks for the old ones.

These are a bit like reckons though so interested to hear your and @idavidmcdonald 's opinion.

No problem with the rest of it, all makes sense.

Wynndow:
@tombye Yep, that makes sense about the renaming. Could only find the examples you mentioned though, not sure it needs changing elsewhere? Also, have updated that test.

@idavidmcdonald I've chucked in an extra test for the flag being `False`. It feels a little redundant, as when the template renders one of the first things it does is check the status of the brief. If it's `closed`, it just renders the brief and if you've applied, skipping anything to do with the new supplier flag or anything around dates. What do you think?

tests/app/views/test_marketplace.py
```diff
@@ -377,7 +377,7 @@ def test_dos_brief_has_different_link_text_for_logged_in_supplier(self):
 
         assert_equal(qa_link_text.strip(), "Ask a question")
 
-    def test_can_apply_to_live_brief_with_old_supplier_flow(self):
+    def test_can_apply_to_live_brief_with_legacy_supplier_flow(self):
         self.app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'] = False
```
idavidmcdonald: Should this test exist twice? Once with the feature flag false and once with the brief published at date before the supplier feature flag?

tests/app/views/test_marketplace.py
```diff
@@ -399,8 +400,39 @@ def test_can_apply_to_live_brief_with_new_supplier_flow(self):
     def test_cannot_apply_to_closed_brief(self):
         brief = self.brief.copy()
         brief['briefs']['status'] = "closed"
-        brief['briefs']['publishedAt'] = "2000-01-25T12:00:00.000000Z"
-        brief['briefs']['applicationsClosedAt'] = "2000-02-25T12:00:00.000000Z"
+        brief['briefs']['publishedAt'] = "2016-12-01T12:00:00.000000Z"
```
idavidmcdonald: Maybe same here on the comment about published at after the feature flag. It does seem like a bit silly to do so but it took me quite a while to get my head around the tests going on and I've spent plenty of time on the context so hopefully it will be worth it at a later point.

tests/app/views/test_marketplace.py
```diff
@@ -399,8 +400,39 @@ def test_can_apply_to_live_brief_with_new_supplier_flow(self):
...
+        document = html.fromstring(res.get_data(as_text=True))
+
+        apply_links = document.xpath('//a[@href="/suppliers/opportunities/{}/responses/start"]'.format(brief_id))
+        assert len(apply_links) == 0
+        assert '2 December 2016' in document.xpath('//p[@class="banner-message"]')[0].text_content()
+
+    def test_cannot_apply_to_closed_legacy_brief(self):
+        brief = self.brief.copy()
+        brief['briefs']['status'] = "closed"
+        brief['briefs']['publishedAt'] = "2016-11-20T12:00:00.000000Z"
```
idavidmcdonald: Same as above regarding comment.

tests/app/views/test_marketplace.py
```diff
@@ -399,8 +400,39 @@ def test_can_apply_to_live_brief_with_new_supplier_flow(self):
...
+        brief_id = brief['briefs']['id']
+        res = self.client.get('/digital-outcomes-and-specialists/opportunities/{}'.format(brief_id))
+        assert_equal(200, res.status_code)
+        document = html.fromstring(res.get_data(as_text=True))
+
+        apply_links = document.xpath('//a[@href="/suppliers/opportunities/{}/responses/create"]'.format(brief_id))
+        assert len(apply_links) == 0
+        assert '21 November 2016' in document.xpath('//p[@class="banner-message"]')[0].text_content()
+
+    def test_cannot_apply_to_closed_brief_with_legacy_supplier_flow(self):
```
idavidmcdonald: I'm confused by this test name. What is the difference between this and `test_cannot_apply_to_closed_legacy_brief` above it? If it is that the feature flag is off in this test then it might be worth including that in the test name

idavidmcdonald: Oh and if the two tests are so similar but it's just the feature flag date that differs then you might be able to have them both run in one test by using something along the lines of `for feature_flag_date in (False, 'your date here'):`

Wynndow:
@idavidmcdonald @tombye 
I've restructured the tests and removed a fair bit of duplication I hadn't realised was in there. Hopefully stuff is a bit clearer now, and it should be pretty trivial to strip out these extra tests once they're no longer needed. Have a look and let me know what you think!

idavidmcdonald:
I think this looks good. Good comments on those child classes, nice and clear what they are doing and then will be easy to remove later.

Should be good after a rebase and tests green. 👍

PR #415, opened 2017-01-04
ODS support

danielknell:
- generate CSV or ODS based on brief published time.
- generate ODS from new section in manifest.

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -1,8 +1,11 @@
 # coding: utf-8
 from __future__ import unicode_literals
 import unicodecsv
+import inflection
+import sys
 
-from flask import abort, render_template, request, redirect, url_for, flash, Response
+from flask import abort, render_template, request, redirect, url_for, flash, Response, current_app
+from flask.views import View
 from flask_login import current_user
 
 from app import data_api_client
@@ -14,9 +17,17 @@ from ..helpers.buyers_helpers import (
     section_has_at_least_one_required_question
 )
 
+from ..helpers.ods import SpreadSheet
+
 from dmapiclient import HTTPError
 from dmutils.dates import get_publishing_dates
+from dmutils.formats import DATETIME_FORMAT
 from dmutils import csv_generator
+from datetime import datetime
+
+from odf.style import TextProperties, TableRowProperties, TableColumnProperties, TableCellProperties, FontFace
+
+from io import BytesIO
 
 
 @buyers.route('')
@@ -270,59 +281,228 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
     ), 200
 
 
-@buyers.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses/download',
-              methods=['GET'])
-def download_brief_responses(framework_slug, lot_slug, brief_id):
-    get_framework_and_lot(framework_slug, lot_slug, data_api_client, status='live', must_allow_brief=True)
-    brief = data_api_client.get_brief(brief_id)["briefs"]
+class DownloadBriefResponsesView(View):
+    def __init__(self, **kwargs):
+        self.data_api_client = kwargs.pop('data_api_client', data_api_client)
+        self.content_loader = kwargs.pop('content_loader', content_loader)
 
-    if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
-        abort(404)
+        super(View, self).__init__(**kwargs)
 
-    if brief['status'] != "closed":
-        abort(404)
+    def get_responses(self, brief):
+        return get_sorted_responses_for_brief(brief, self.data_api_client)
 
-    sorted_brief_responses = get_sorted_responses_for_brief(brief, data_api_client)
+    def get_context_data(self, **kwargs):
+        get_framework_and_lot(kwargs['framework_slug'], kwargs['lot_slug'],
+                              self.data_api_client, status='live',
+                              must_allow_brief=True)
 
-    content = content_loader.get_manifest(brief['frameworkSlug'], 'legacy_output_brief_response').filter(
-        {'lot': brief['lotSlug']}
-    )
-    section = content.get_section('view-response-to-requirements')
+        brief = self.data_api_client.get_brief(kwargs['brief_id'])["briefs"]
 
-    column_headings = []
-    question_key_sequence = []
-    boolean_list_questions = []
-    csv_rows = []
+        if not is_brief_correct(brief, kwargs['framework_slug'],
+                                kwargs['lot_slug'], current_user.id):
+            abort(404)
 
-    # Build header row from manifest and add it to the list of rows
-    for question in section.questions:
-        question_key_sequence.append(question.id)
-        if question['type'] == 'boolean_list' and brief.get(question.id):
-            column_headings.extend(brief[question.id])
-            boolean_list_questions.append(question.id)
-        else:
-            column_headings.append(question.name)
-    csv_rows.append(column_headings)
+        if brief['status'] != "closed":
+            abort(404)
 
-    # Add a row for each eligible response received
-    for brief_response in sorted_brief_responses:
-        if all_essentials_are_true(brief_response):
-            row = []
-            for key in question_key_sequence:
-                if key in boolean_list_questions:
-                    row.extend(brief_response.get(key))
+        text_type = str if sys.version_info[0] == 3 else unicode
+        filename = inflection.parameterize(text_type(brief['title']))
+
+        kwargs.update({
+            'brief': brief,
+            'responses': self.get_responses(brief),
+            'filename': 'supplier-responses-{0}'.format(filename),
+        })
+
+        return kwargs
+
+    def get_questions(self, framework_slug, lot_slug, manifest):
+        section = 'view-response-to-requirements'
+
+        result = self.content_loader.get_manifest(framework_slug, manifest)\
+                                    .filter({'lot': lot_slug})\
+                                    .get_section(section)
+
+        return result.questions if result else []
+
+    def create_csv_response(self, context=None):
+        column_headings = []
+        question_key_sequence = []
+        boolean_list_questions = []
+        csv_rows = []
+        brief = context['brief']
+
+        questions = self.get_questions(context['framework_slug'],
+                                       context['lot_slug'],
+                                       'legacy_output_brief_response')
+
+        # Build header row from manifest and add it to the list of rows
+        for question in questions:
+            question_key_sequence.append(question.id)
+            if question['type'] == 'boolean_list' and brief.get(question.id):
+                column_headings.extend(brief[question.id])
+                boolean_list_questions.append(question.id)
+            else:
+                column_headings.append(question.name)
+        csv_rows.append(column_headings)
+
+        # Add a row for each eligible response received
+        for brief_response in context['responses']:
+            if all_essentials_are_true(brief_response):
+                row = []
+                for key in question_key_sequence:
+                    if key in boolean_list_questions:
+                        row.extend(brief_response.get(key))
+                    else:
+                        row.append(brief_response.get(key))
+                csv_rows.append(row)
+
+        return Response(
+            csv_generator.iter_csv(csv_rows),
+            mimetype='text/csv',
+            headers={
+                "Content-Disposition": (
+                    "attachment;filename=responses-to-requirements-{}.csv"
+                ).format(brief['id']),
+                "Content-Type": "text/csv; header=present"
+            }
+        ), 200
+
+    def generate_ods(self, brief, responses):
+        doc = SpreadSheet()
+
+        doc.add_font(FontFace(name="Arial", fontfamily="Arial"))
+
+        doc.add_style("ce1", "table-cell", (
+            TableCellProperties(wrapoption="wrap", verticalalign="top"),
+            TextProperties(fontfamily="Arial", fontnameasian="Arial",
+                           fontnamecomplex="Arial", fontsize="11pt"),
+        ), parentstylename="Default")
+
+        doc.add_style("ce2", "table-cell", (
+            TableCellProperties(wrapoption="wrap", verticalalign="top",
+                                backgroundcolor="#f3f3f3"),
+            TextProperties(fontfamily="Arial", fontnameasian="Arial",
+                           fontnamecomplex="Arial", fontsize="11pt",
+                           fontweight="bold"),
+        ), parentstylename="Default")
+
+        doc.add_style("ce3", "table-cell", (
+            TableCellProperties(wrapoption="wrap", verticalalign="top",
+                                backgroundcolor="#f3f3f3"),
+            TextProperties(fontfamily="Arial", fontnameasian="Arial",
+                           fontnamecomplex="Arial", fontsize="11pt"),
+        ))
+
+        doc.add_style("co1", "table-column", (
+            TableColumnProperties(columnwidth="150pt", breakbefore="auto"),
+        ))
+
+        doc.add_style("co2", "table-column", (
+            TableColumnProperties(columnwidth="300pt", breakbefore="auto"),
+        ))
+
+        doc.add_style("ro1", "table-row", (
+            TableRowProperties(rowheight="30pt", breakbefore="auto",
+                               useoptimalrowheight="false"),
+        ))
+
+        doc.add_style("ro2", "table-row", (
+            TableRowProperties(rowheight="30pt", breakbefore="auto",
+                               useoptimalrowheight="true"),
+        ))
+
+        sheet = doc.sheet("Supplier evidence")
+
+        questions = self.get_questions(brief['frameworkSlug'],
+                                       brief['lotSlug'],
+                                       'output_brief_response')
+
+        # two intro columns for boolean and dynamic lists
+        sheet.create_column(stylename="co1", defaultcellstylename="ce1")
+        sheet.create_column(stylename="co1", defaultcellstylename="ce1")
+
+        # HEADER
+        row = sheet.create_row("header", stylename="ro1")
+        row.write_cell(brief['title'], stylename="ce2",
+                       numbercolumnsspanned=str(len(responses) + 2))
+
+        # QUESTIONS
+        for question in questions:
+            if question.type in ('boolean_list', 'dynamic_list'):
+                length = len(brief[question.id])
+
+                for i, requirement in enumerate(brief[question.id]):
+                    row = sheet.create_row("{0}[{1}]".format(question.id, i))
+                    if i == 0:
+                        row.write_cell(question.name, stylename="ce2",
+                                       numberrowsspanned=str(length))
+                    else:
+                        row.write_covered_cell()
+                    row.write_cell(requirement, stylename="ce3")
+            else:
+                row = sheet.create_row(question.id, stylename="ro2")
+                row.write_cell(question.name, stylename="ce2",
+                               numbercolumnsspanned="2")
+                row.write_covered_cell()
+
+        # RESPONSES
+        for response in responses:
+            sheet.create_column(stylename="co2", defaultcellstylename="ce1")
+
+            for question in questions:
+                if question.type == 'dynamic_list':
+                    for i, item in enumerate(response[question.id]):
+                        row = sheet.get_row("{0}[{1}]".format(question.id, i))
+                        # TODO this is stupid, fix it (key should not be hard coded)
+                        row.write_cell(item.get('evidence') or '',
+                                       stylename="ce1")
+
+                elif question.type == 'boolean_list':
+                    for i, item in enumerate(response[question.id]):
+                        row = sheet.get_row("{0}[{1}]".format(question.id, i))
+                        row.write_cell(str(bool(item)).lower(),
+                                       stylename="ce1")
+
                 else:
-                    row.append(brief_response.get(key))
-            csv_rows.append(row)
+                    sheet.get_row(question.id).write_cell(response[question.id],
+                                                          stylename="ce1")
 
-    return Response(
-        csv_generator.iter_csv(csv_rows),
-        mimetype='text/csv',
-        headers={
-            "Content-Disposition": "attachment;filename=responses-to-requirements-{}.csv".format(brief['id']),
-            "Content-Type": "text/csv; header=present"
-        }
-    ), 200
+        return doc
+
+    def create_ods_response(self, context=None):
+        buf = BytesIO()
+
+        self.generate_ods(context['brief'], context['responses']).save(buf)
+
+        return Response(
+            buf.getvalue(),
+            mimetype='application/vnd.oasis.opendocument.spreadsheet',
+            headers={
+                "Content-Disposition": (
+                    "attachment;filename={0}.ods"
+                ).format(context['filename']),
+                "Content-Type": "application/vnd.oasis.opendocument.spreadsheet"
+            }
+        ), 200
+
+    def create_response(self, context=None):
+        responses = context['responses']
+
+        if responses and 'essentialRequirementsMet' in responses[0]:
+            return self.create_ods_response(context)
+
+        return self.create_csv_response(context)
+
+    def dispatch_request(self, **kwargs):
+        context = self.get_context_data(**kwargs)
+
+        return self.create_response(context)
+
+
+buyers.add_url_rule('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses/download',
+                    view_func=DownloadBriefResponsesView.as_view(str('download_brief_responses')),
+                    methods=['GET'])
 
 
 @buyers.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/publish', methods=['GET', 'POST'])
--- a/tests/buyers/views/test_buyers.py
+++ b/tests/buyers/views/test_buyers.py
@@ -4,9 +4,22 @@ from __future__ import unicode_literals
 from ...helpers import BaseApplicationTest
 from dmapiclient import api_stubs, HTTPError
 from dmcontent.content_loader import ContentLoader
+from dmcontent.questions import Question
 import mock
 from lxml import html
 import pytest
+
+from app.buyers.views import buyers
+from dmcontent.content_loader import ContentLoader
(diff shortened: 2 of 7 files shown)


app/helpers/ods.py
```diff
@@ -0,0 +1,76 @@
...
+        
+    def sheet(self, name):
+        if name not in self._sheets:
+            self._sheets[name] = Sheet(name)
+            self._document.spreadsheet.addElement(self._sheets[name]._table)
+            
+        return self._sheets[name]
+    
+    def add_style(self, name, family, styles, **kwargs):
+        kwargs['name'] = name
```
galund: Would it be neater to just write this:` Style(name=name, family=family, **kwargs`

danielknell: probably, will make a note to fix when i tidy up.

app/buyers/views/buyers.py
```diff
@@ -270,8 +278,225 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+
+        content = content_loader.get_manifest(
+            brief['frameworkSlug'], 
+            'output_brief_response'
+        ).filter({'lot': brief['lotSlug']})
+
+        section = content.get_section('view-ods')
+
+        sheet.create_column(stylename="intro", defaultcellstylename="default")
+        sheet.create_column(stylename="intro", defaultcellstylename="default")
```
pcraig3: It might be helpful to add a comment what data these columns contain (unless it's specified elsewhere that's really obvious).  I was just glancing through and I thought this might have been a copy+paste error.

app/buyers/views/buyers.py
```diff
@@ -270,8 +279,226 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+
+        # HEADER
+        row = sheet.create_row("header")
+        row.write_cell(brief['title'], stylename="header",
+                       numbercolumnsspanned=str(len(responses) + 2))
+
+        # QUESTIONS
+        for question in section.questions:
+            if question.type in ('boolean_list', 'dynamic_list'):
```
idavidmcdonald: For the new flow and thus the ODS we are no longer asking any boolean list questions. (We currently look like we are from the current frameworks repo as the nice to have requirements question is a boolean list but this about to be replaced by a dynamic list question). Therefore I think we can lose the boolean list requirement.

app/buyers/views/buyers.py
```diff
@@ -270,8 +279,226 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+        section = content.get_section('view-ods')
+
+        # two intro columns for boolean and dynamic lists
+        sheet.create_column(stylename="intro", defaultcellstylename="default")
+        sheet.create_column(stylename="intro", defaultcellstylename="default")
+
+        # HEADER
+        row = sheet.create_row("header")
+        row.write_cell(brief['title'], stylename="header",
+                       numbercolumnsspanned=str(len(responses) + 2))
```
idavidmcdonald: Might be nice to give some context about what the + 2 is doing

idavidmcdonald:
I'm liking the class based view, it gives a really good separation of logic and makes it easily readable what is going on.

app/buyers/views/buyers.py
danielknell: who knows how that will change, no reason not to support them considering its a few lines

idavidmcdonald: Good point (although we are planning on deprecating boolean list questions so I don't know if that changes your mind). Happy either way

TheDoubleK: "*deprecating boolean list questions*"?!? 
Me and @pcraig3 sweated blood over those mothers.

pcraig3: I cannot begin to express how upset I am. It's taken me a full 24 hours before I could even calm down enough to respond to this.

tests/features.py
```diff
@@ -0,0 +1,42 @@
+from flask import current_app
+
+
+def enable_feature(name, value=True):
```
allait: For the Slack discussion, could this be replaced with
```python
def enable_flag(name, value=True):
    PATCH = {'FEATURE_FLAGS_{}'.format(name.upper()): value}
    return patch.dict(current_app.config, PATCH)
```

galund: FWIW, the name of the function doesn't work for me, because you might end up writing `enable_flag('WOBBLE', False)`, and that reads quite badly.

Maybe `set_flag` would be better?

idavidmcdonald: would `feature_flag` rather than `flag` be better? (happy to be told otherwise)

galund: The nice thing about 'set' is that on assumes it's positive (i.e. True), but it still makes sense for False. `feature_flag` doesn't make it clear what the default is, IMO

idavidmcdonald: Sorry, I meant `enable_feature_flag` as opposed to `enable_flag`. Or if we went with your suggestion then `set_feature_flag` rather than `set_flag`.

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+        self.brief['essentialRequirements'] = [
+            "Good nose for tea",
+            "Good eye for biscuits",
+            "Knowledgable about tea"
+        ]
+        self.brief['niceToHaveRequirements'] = [
+            "Able to bake",
+            "Able to perform the tea ceremony"
+        ]
+        self.brief['blah'] = ['Affirmative', 'Negative']
```
idavidmcdonald: What is this for? If it is to signify an unrelated key on the brief then it might be good to name it as such `self.brief['unrelatedKey']` or `self.brief['nonImportantKey']`

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+
+        self.responses = [
+            {
+                "supplierName": "Prof. T. Maker",
+                "respondToEmailAddress": "t.maker@example.com",
+                "niceToHaveRequirements": [
+                    {
+                        "yesNo": False,
+                        "evidence": ""
```
idavidmcdonald: If this data has come from the API then if `yesNo` is `False` then the `evidence` key will not exist

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+                    }
+                ],
+                "availability": "2017-12-25",
+                "essentialRequirementsMet": True,
+                "essentialRequirements": [
+                    {
+                        "evidence": "From Assan to Yixing I've got you covered."
```
idavidmcdonald: Great tea based examples

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+                ).format(context['brief']['id']),
+                "Content-Type": "application/vnd.oasis.opendocument.spreadsheet"
+            }
+        )
+
+    def test_create_response(self):
+        self.instance.create_ods_response = mock.Mock()
+        self.instance.create_csv_response = mock.Mock()
+
+        context = {'responses': [{'essentialRequirementsMet': True }]}
```
idavidmcdonald: Additional whitespace failing pep8 here

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+                        row.append(brief_response.get(key))
+                csv_rows.append(row)
+
+        return Response(
+            csv_generator.iter_csv(csv_rows),
+            mimetype='text/csv',
+            headers={
+                "Content-Disposition": (
+                    "attachment;filename=responses-to-requirements-{}.csv"
+                ).format(brief['id']),
```
idavidmcdonald: You haven't introduced this behaviour but was suprised to see we use the brief id in the filename as that is unlikely to be helpful for buyers to identify the files

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+                    # TODO this is stupid too, fix it (name needs to come from question)
+                    row = sheet.get_row(question.id)
+                    row.write_cell("Evidence", stylename="subtitle")
+
+                    for i, item in enumerate(response[question.id]):
+                        row2 = sheet.get_row("{0}[{1}]".format(question.id, i))
+                        # TODO omg its everywhere, fix it (key needs to come from question)
+                        row2.write_cell(item.get('evidence') or '')
+
+                elif question.type == 'boolean_list':
```
idavidmcdonald: We aren't expecting any boolean list questions to be in the ODS

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+
+    def dispatch_request(self, **kwargs):
+        context = self.get_context_data(**kwargs)
+
+        return self.create_response(context)
+
+
+buyers.add_url_rule('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses/download',
+                    view_func=DownloadView.as_view(str('download_brief_responses')),
+                    methods=['GET'])
```
idavidmcdonald: Sorry this might be my lack of experience with class based views but how come 'download_brief_responses' is wrapped in `str`?

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
     ), 200
 
 
-@buyers.route('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses/download',
-              methods=['GET'])
-def download_brief_responses(framework_slug, lot_slug, brief_id):
-    get_framework_and_lot(framework_slug, lot_slug, data_api_client, status='live', must_allow_brief=True)
-    brief = data_api_client.get_brief(brief_id)["briefs"]
+class DownloadView(View):
```
idavidmcdonald: `DownloadBriefResponsesView` could be more descriptive

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
         assert res.status_code == 404
 
 
+class TestDownloadView(BaseApplicationTest):
+    def setup_method(self, method):
+        super(TestDownloadView, self).setup_method(method)
+
+        self.data_api_client = mock.MagicMock(spec_set=DataAPIClient)
```
idavidmcdonald: Hadn't seen spec_set before. Good idea to include

app/buyers/views/buyers.py
danielknell: because unicode_litterals is stupid and breaks flask in places, this is one...

that value gets assigned as the name of the function so routing treats it the same `def download_brief_responses` but you cant have unicode function names in python2

app/buyers/views/buyers.py
danielknell: didn't take much to support them, it should support any kind of question type tbh as we dont know what will happen in the manifests going forward.

app/buyers/views/buyers.py
danielknell: it does however mean the file names dont conflict when downloading from multiple places.

tests/app/views/test_buyers.py
danielknell: urgh, i thought i got all those...

tests/app/views/test_buyers.py
danielknell: why are there no tea terms starting with Z :(

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1882,396 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+
+        self.instance.data_api_client.get_brief\
+            .assert_called_once_with(kwargs['brief_id'])
+
+        self.instance.get_responses\
+            .assert_called_once_with(brief['briefs'])
+
+        assert result == expected
+
+    def test_get_context_data_incorrect_brief(self):
```
idavidmcdonald: Super small but naming your test based on the situation + expected result would make this more readable e.g. `test_get_context_data_with_incorrect_brief_will_abort`, same as below with `test_get_context_data_with_open_brief_will_abort`.

tests/app/views/test_buyers.py
danielknell: its used to test boolean questions

idavidmcdonald: Ah OK. Maybe `booleanListQuestion` would be an appropriate key then

tests/app/views/test_buyers.py
```diff
@@ -4,9 +4,17 @@
 from ...helpers import BaseApplicationTest
 from dmapiclient import api_stubs, HTTPError
 from dmcontent.content_loader import ContentLoader
+from dmcontent.questions import Question
 import mock
 from lxml import html
 import pytest
+from ...features import disable_feature
```
idavidmcdonald: Is this being used?

tests/app/views/test_buyers.py
```diff
@@ -4,9 +4,17 @@
...
+
+from app.buyers.views import buyers
+from dmcontent.content_loader import ContentLoader
+from dmapiclient import DataAPIClient
+import functools
+from string import ascii_uppercase
```
idavidmcdonald: Is this being used?

idavidmcdonald:
Is this the best place for the ODS helper to live or should it live in a more shareable place?

app/buyers/views/buyers.py
idavidmcdonald: True, consistency would be nice. I'd suggest checking with @tracy-content regarding file names. It looks like nothing has yet been confirmed (or at least written on the pivotal) so we should check if it should remain as the CSV or we would prefer something else: https://www.pivotaltracker.com/story/show/133916871

tests/app/views/test_buyers.py
danielknell: tis full of awesome, its like spec's better looking cousin.

tests/app/views/test_buyers.py
danielknell: ah, i think i removed the bit using that, this is why we should be using flake8 instead of pep8, it complains about unused imports

danielknell:
i am reluctant to move ods stuff elsewhere at the moment, as its not used elsewhere, when that comes up we can find a home for it, or hopefully ODS libraries will be less full of fail by that point...

app/buyers/views/buyers.py
danielknell: that feels excessively verbose, yay java... but what the hell.

tests/app/views/test_buyers.py
danielknell: i dont think it matters.

idavidmcdonald: I disagree. Coming at this fresh it took me a while to understand what this was for (at first I thought it wasn't even a needed key) and a more descriptive key would have cut that time dramatically. It feels like there is no harm in changing it and should only improve readability

idavidmcdonald:
Fair enough about the location of the ODS helpers.

Only things remaining from me are 
- the small key change from using 'blah' to something descriptive (I think we may disagree on whether this excessive but if you do feel strongly then get a second opinion to help decide)
- Confirm desired filename with @tracy-content and add to the pivotal story

Once those are addressed then I'm happy.

app/helpers/ods.py
```diff
@@ -0,0 +1,87 @@
...
+
+class Row(object):
+    def __init__(self, **kwargs):
+        self._row = TableRow(**kwargs)
+
+    def write_cell(self, value, **kwargs):
+        if "numbercolumnsspanned" in kwargs:
+            kwargs.setdefault("numberrowsspanned", "1")
+        if "numberrowsspanned" in kwargs:
+            kwargs.setdefault("numbercolumnsspanned", "1")
```
pcraig3: Can you put in both?  Will this work anyway or will that cause this method to break?

app/helpers/ods.py
```diff
@@ -0,0 +1,87 @@
...
+
+        cell = TableCell(**kwargs)
+        cell.setAttrNS(OFFICENS, "value-type", "string")
```
pcraig3: What does this do?  I've had a go at looking it up but I'm not sure I understood what the output is.

tests/features.py
```diff
@@ -0,0 +1,41 @@
...
+            with patch.dict(current_app.config, PATCH):
+                return f(*args, **kwargs)
+
+        return wrap
+
+    return decorator
+
+
+def disable_feature(name):
+    return enable_feature(name, False)
```
pcraig3: Has this gotten the go-ahead?  There was a discussion last time in #dm-development about overriding the config variable if possible.  Did that idea get torpedoed?

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+            TableCellProperties(wrapoption="wrap", verticalalign="top"),
+            TextProperties(fontfamily="Arial", fontnameasian="Arial",
+                           fontnamecomplex="Arial", fontsize="16pt",
+                           fontweight="bold"),
+        ))
+
+        doc.add_style("subtitle", "table-cell", (
+            TextProperties(fontfamily="Arial", fontnameasian="Arial",
+                           fontnamecomplex="Arial", fontsize="16pt"),
+        ))
```
pcraig3: Looks like tons of `fontfamily="Arial"`s, etc.  Any way of grouping common things in a way that makes it shorter (but doesn't make it way more complicated to read) or is this the best we can do?

pcraig3:
Yeah, I agree with what @idavidmcdonald said 20 days ago.  I hadn't seen this idea of using a `View` object before ([docs](http://flask.pocoo.org/docs/0.12/views/)) but it looks wicked good.

app/buyers/views/buyers.py
galund: If we're going there, shouldn't it be "new transport" anyway :-D

allait: @galund It's unlikely that anyone has New Transport installed locally and I'm not sure if GDS has the license to distribute the font.

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1880,394 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+        }
+
+        expected = dict(**kwargs)
+        expected['brief'] = brief['briefs']
+        expected['responses'] = self.instance.get_responses.return_value
+
+        self.instance.data_api_client.get_brief.return_value = brief
+
+        with po(buyers, 'get_framework_and_lot') as get_framework_and_lot:
+            with po(buyers, 'is_brief_correct') as is_brief_correct:
```
allait: I like this approach in general. It's nice that we don't need to mock the data for this, although I'd rather we do something about the 4-level nested patches.

But one problem is that autospec isn't doing much apart from checking the number of arguments. If their order or types ever change, the test will keep passing.

So if we're doing this, I think we need at least one test to test the integration.

It also feels like we need to set the return value. It's great that mock objects are truthy, but I don't think there's a good reason to obscure the actual return type.

tests/app/views/test_buyers.py
```diff
@@ -1874,6 +1880,394 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
         assert res.status_code == 404
 
 
+class TestDownloadBriefResponsesView(BaseApplicationTest):
```
allait: While testing the callable directly seems like a useful approach, I think we're missing some things that we usually test for the view.

We no longer have a test checking that URL maps correctly to the view arguments. And we no longer have a test to check that user authentication is required. I think both are useful for the view, so I think we should add them.

tests/features.py
```diff
@@ -0,0 +1,41 @@
+from flask import current_app
+from mock import patch
+
+
+def enable_feature(name, value=True):
```
allait: I think there was some reason why we couldn't go with 
```python
def enable_flag(name, value=True):
    PATCH = {'FEATURE_FLAGS_{}'.format(name.upper()): value}
    return patch.dict(current_app.config, PATCH)
```
but I don't think it's captured in this PR. Do you remember what it was?

tests/unit/test_ods.py
```diff
@@ -0,0 +1,214 @@
+import mock
+import unittest
+import functools
+
+import app.helpers.ods as ods
+
+from hypothesis import strategies as st
+from hypothesis import given, example, assume
+
+po = functools.partial(mock.patch.object, autospec=True)
```
allait: This feels like a better approach than repeating it in every test method. Is there a reason view tests aren't doing the same?

I'd also go a step further and move this to the test helpers, but with a more descriptive name.

app/buyers/views/buyers.py
```diff
@@ -270,59 +279,229 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
...
+        return self.create_csv_response(context)
+
+    def dispatch_request(self, **kwargs):
+        context = self.get_context_data(**kwargs)
+
+        return self.create_response(context)
+
+
+buyers.add_url_rule('/buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/responses/download',
+                    view_func=DownloadBriefResponsesView.as_view(str('download_brief_responses')),
```
allait: I might be missing something, but I think this is missing the kwargs we need. As far as I understand `.as_view` will forward the arguments to the class constructor, so we need to pass `data_api_client` and `content_loader` here.

I think this is a strong example for why we need to have some tests using the test client for each view.

tests/features.py
allait: Oh, and as David said, this seems to be unused. I say we drop this if that's the case. While there's a possibility that this would be useful, I'd rather not add dead code. Chances are, with no existing usage examples, whoever's going to need this functionality will likely end up implementing it somewhere else anyway.

app/helpers/ods.py
danielknell: aye, its both or nothing, so this just ensures if one is set the other will default to 1 if not set, could rewrite as:

```python
if "numbercolumnsspanned" in kwargs or  "numberrowsspanned" in kwargs:
    kwargs.setdefault("numberrowsspanned", "1")
    kwargs.setdefault("numbercolumnsspanned", "1")
```

app/helpers/ods.py
danielknell: it sets an attribute on the cell with a namespace, without it nothing shows in excel... i have no idea why it fixes things but found it by digging into working files.

tests/features.py
danielknell: overriding at class levels upsets flask as theres no app context...

app/buyers/views/buyers.py
danielknell: its a mess but short of building a proper spreadsheet library (which would probably take much longer) i would rather put up with the lower level stuff

tests/unit/test_ods.py
danielknell: its short because patch.object along with the keys ends up being stupid long, and i didn't want to add a global randomly to the other test file.

app/buyers/views/buyers.py
danielknell: those kwargs default to the globals

tests/features.py
danielknell: the above code will throw an error about flasks application context being accessed out of scope...

danielknell: ill just delete.

app/buyers/views/buyers.py
allait: Ah, of course. I've missed the defaults.

tests/unit/test_ods.py
allait: I say give it a better name and put it in helpers. Repeating this in every test method doesn't add much. And pyflakes will tell you if any of the local variables clash with your imported name.

allait:
The blueprint split PR (https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/419) got merged, so some of the files have to be moved during rebase. `tests/app`, `tests/unit` and `app/helpers` folders no longer exist, they are replaced with`tests/buyers/views`, `tests/buyers/helpers` and `app/buyers/helpers`.

tests/unit/test_ods.py
danielknell: thats the problem, a better name makes it less useful, its a shorthand...

danielknell:
@pcraig3 view objects are amazing, I hadn't used them in flask till an old colleague showed me how much cleaner they were compared to our function based views in that project,  I was sold,  and I've used them near exclusively since.

app/buyers/views/buyers.py
galund: Not important, but I believe the Crown owns the font, and embedding fonts in ODF is definitely a thing :) https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/embed-fonts-in-odf-documents

idavidmcdonald:
No one else mentioned the 'blah' key Dan so I will concede on that. However we still need confirmation on the filename from @tracy-content. 

You can either wait for one to be confirmed, fix this PR and merge or you can merge this PR but create a new pivotal story to fix the filename so don't forget to do it. Let me know what you decide.

danielknell:
tracy confirmed the filename already.

tests/buyers/views/test_buyers.py
```diff
@@ -1874,6 +1887,406 @@ def test_404_if_lot_does_not_allow_brief(self, data_api_client):
...
+                mock.patch.object(buyers, 'current_user') as current_user:
+
+            is_brief_correct.return_value = False
+            with pytest.raises(NotFound):
+                self.instance.get_context_data(**kwargs)
+
+    def test_get_context_data_with_open_brief(self):
+        self.instance.get_responses = mock.Mock()
+
+        brief = api_stubs.brief(status='open')
```
idavidmcdonald: Should the status be 'live' rather than 'open' - https://github.com/alphagov/digitalmarketplace-api/blob/master/app/models.py#L1197

idavidmcdonald:
👍 wooooo well done!

PR #417, opened 2017-01-10
Don't call filter() unnecessarily

TheDoubleK:
Clarification questions don't have any `depends` dependencies in content, so the call to filter here doesn't achieve anything.


TheDoubleK:
Closing this as apparently `filter()` does more than just 
```
        """Return a new :class:`ContentManifest` filtered by service data

        Only includes the questions that should be shown for the provided
        service data. This is calculated by resolving the dependencies
        described by the `depends` section."""
```

PR #419, opened 2017-01-20
[#136176723] Briefs blueprint split

allait:
Splitting brief-related modules into a separate blueprint to make moving them to a new app easier.

### What's done

Most of the application and test code is now split between `buyers/` and `main/`. `buyers` will become the base for the separate briefs app.

There should be no direct imports between the packages, but there are still indirect dependencies that would need to be copied (eg application error handlers are defined for the entire app in `main.errors`)

#### Split helpers and move brief start pages to buyers/

Brief start pages for each lot have a `/buyers` prefix, so in order to allow routing between different apps they'd have to be part of the briefs app.

Unlike every other route in buyers they don't require a login so we create an additional blueprint (`dos` to match the view name, but this can be changed later) in `app.buyers` that doesn't require authorization.

Split the rest of the helpers and associated tests.

#### Move /buyers URL prefix to the blueprints

This makes sure that all brief URLs start with /buyers.

The only remaining views that use the same prefix outside the buyers blueprint are `/buyers/create` GET and POST views. Moving them from main requires duplicating a lot of the forms and helper functions, and it feels like the right solution for them is a separate users app.

The simplest thing we can do right now would be to change URLs to eg `/create/buyer`. But I don't want to introduce any functional changes with this split, so I'm leaving this for now.

#### Split tests between blueprints

Move test files from tests/app and tests/unit to tests/main and tests/buyers, with the nested directory structure mirroring the structure in app/

### What's not done

* Templates - it's possible to define blueprint `template_folder`, but it adds directories to the search path, so it's still hard to verify that templates aren't using anything outside their own namespace
* Static files - we'd have to split, copy or move to frontend toolkit any js/css files that are currently part of the buyer frontend
* `/buyers/create` views to create buyer users are still part of `app.main`. Those URLs would probably need to change before the split. We could also consider moving user management to a separate application
* Some of the application setup code is done in `app.main`. Things like Flask error handlers would have to be copied when the apps are split
* We probably should consider moving things like the status blueprint and `tests/helpers.py` to shared libraries. They're duplicated between the apps, and `/_status` feels like it's definitely part of the contract all apps need to follow.


pcraig3:
I think it all looks good, honestly.

The stuff that changed is pretty straightforward and there's nothing in here that seems surprising or unexpected.  

Can you just merge it if I give it a 👍 , or do you want anyone else to look at it?

PR #421, opened 2017-01-24
 Enable old and new 'shortlist suppliers' page to have different content

danielknell:
updates content and icon based on whether its the new flow or old.

### DIFF
--- a/app/assets/scss/_brief_submission.scss
+++ b/app/assets/scss/_brief_submission.scss
@@ -52,6 +52,13 @@ ol.instruction-list-item-top li {
 }
 
 ol.steps {
+  @include bold-27;
+  padding-left: 1em; // 27px
+
+  @include media( $min-width: $site-width + (2 * (27px + $gutter) ) ) {
+    padding-left: 0;
+  }
+
   list-style: none;
   padding-left: 0;
 
--- a/app/templates/buyers/brief_responses.html
+++ b/app/templates/buyers/brief_responses.html
@@ -39,6 +39,53 @@
     <div class="column-two-thirds">
         <div class="dmspeak">
           {% if response_counts['eligible'] > 0 %}
+            {% if ods %}
+            <p>
+              <span class='visual-emphasis'>{{ response_counts['eligible'] }} {{ pluralize(response_counts['eligible'], "supplier", "suppliers") }}</span>
+              responded to your requirements and {{ pluralize(response_counts['eligible'], "meets", "meet") }} all your essential skills and experience. Suppliers that didn't meet all your essential requirements have already been told they were unsuccessful.
+            </p>
+            <p>
+              You said you’d take <span class='visual-emphasis'>{{ brief.numberOfSuppliers }} {{ pluralize(brief.numberOfSuppliers, "supplier", "suppliers") }}</span> through to the evaluation stage. To do this, you need to:
+            </p>
+            {% with items = [
+                {
+                    "body": 'Read the <a href="https://www.gov.uk/guidance/how-to-shortlist-digital-outcomes-and-specialists-suppliers">guidance on how to shortlist.</a>'|safe,
+                },
+                {
+                    "body": 'Review and score the evidence suppliers have given.',
+                },
+                {
+                    "body": 'Tell the highest scoring suppliers that they’re through to the next stage.',
+                },
+                {
+                    "body": 'Tell any unsuccessful suppliers who don\'t make it through to the next stage.',
+                }
+            ]
+            %}
+                {% include "toolkit/instruction-list.html" %}
+            {% endwith %}
+
+            {% if brief.get("status") == "closed" %}
+                {%
+                with
+                items = [
+                    {
+                        "title": "Download supplier responses to this requirement",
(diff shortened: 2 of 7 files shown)


idavidmcdonald:
I've had a quick glance at this. It looks like it isn't going to be too controversial. Once you've sorted the conflicts I'll take a proper look.

app/assets/scss/_brief_submission.scss
```diff
@@ -52,6 +52,13 @@ ol.instruction-list-item-top li {
 }
 
 ol.steps {
+  @include bold-27;
+  padding-left: 1em; // 27px
+
+  @include media( $min-width: $site-width + (2 * (27px + $gutter) ) ) {
+    padding-left: 0;
+  }
+
```
danielknell: someone carelessly extended it, we pegged their changes on the additional class they added.

tombye: We had a chat about this a while ago.

History of it is that the styles Dan took out of `application.scss` we added when there was only one use of the instruction list pattern in this app, on the brief overview page (which uses the `steps` class for the list of things you need to do to publish the brief). This meant all instruction lists in the app had a 27pt font.

Dan's change puts those styles in the block tied to the `steps` class so tying it to the overview page. This mean our use of the list, in this page, reverted to those defined in [the original pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/instruction-list.html).

In other words, sensible stuff 👍

app/templates/buyers/brief_responses.html
```diff
@@ -39,6 +39,58 @@
...
           {% if response_counts['eligible'] > 0 %}
+            {% if ods %}
+            <p>
+              <span class='visual-emphasis'>{{ response_counts['eligible'] }} {{ pluralize(response_counts['eligible'], "supplier", "suppliers") }}</span>
+              responded to your requirements and {{ pluralize(response_counts['eligible'], "meets", "meet") }} all your essential skills and experience. Suppliers that didn't meet all your essential requirements have already been told they were unsuccessful.
+            </p>
+            <p>
+              You said you’d take <span class='visual-emphasis'>{{ brief.numberOfSuppliers }} {{ pluralize(brief.numberOfSuppliers, "supplier", "suppliers") }}</span> through to the evaluation stage. To do this, you need to:
+            </p>
+            <ol class="instruction-list">
```
tombye: Is there a reason for using the raw HTML for the instruction list pattern rather than the jinja macro?

I'd be better to use the latter if possible.

tests/buyers/views/test_buyers.py
```diff
@@ -1812,79 +1801,129 @@ def test_page_shows_csv_download_link_if_brief_closed(self, data_api_client):
...
+            "CSVdocument:Downloadsupplierresponsesto‘Ineedathingtodoathing’"
 
-    def test_404_if_lot_does_not_allow_brief(self, data_api_client):
+
+@mock.patch("app.buyers.views.buyers.data_api_client")
+class TestViewBriefResponsesPage2(AbstractViewBriefResponsesPage):
+    two_good_three_bad_responses = {
+        "briefResponses": [
+            {"essentialRequirementsMet": True, "essentialRequirements": [{"evidence": "blah"}]},
+            {"essentialRequirementsMet": False, "essentialRequirements": []},
```
idavidmcdonald: As spotted in your other pull request, this isn't actually expected data. You can't have a 'bad' response for the new flow because `essentialRequirementsMet` will either not be set (and therefore not returned by the API) or will be `True`.

Sorry! I was about to thumbs this up and all...

tombye:
For the record, I'm happy with the frontend parts of this 👍 

In future it would be good for this kind of thing to be broken up into more than one commit so discussions like the ones we've had on this can reference changes in isolation through their URLs.

Wynndow:
If I run this branch locally, it looks a bit odd. The `<ol>` has a font size of 20px compared to 16px on the rest of the page. Is this right?
screen shot 2017-02-01 at 11 54 08

tombye:
After seeing the comment by @Wynndow I checked this locally. It renders fine for me with none of the problems with the font size. At a guess I'd say it might be the result of some CSS that needs re-compiling.

I did notice the vertical spacing is a bit big compared to the [instruction-list pattern](http://alphagov.github.io/digitalmarketplace-frontend-toolkit/instruction-list.html).

It's because it's inside a `dmspeak` div which adds `padding-bottom` to all paragraphs in the list. The instruction list already gives them a `margin-bottom` so it's doubling up. If you take it out of the div it sorts this.

### In the dmspeak div

instruction_list_inside_of_dmspeak

### Outside of the dmspeak div

instruction_list_outside_of_dmspeak

Wynndow:
Sorted my problem, Tom was right!

idavidmcdonald:
👍

Nice one Dan. It will need a rebase on top of master however **please hold off merging** as we have lots of things on the buyer frontend in our deployment pipeline already and we don't want to further block that work getting released as this will need to be QA'd.

We will need to write the stories for the bug fix and the content for the no responses.

PR #423, opened 2017-01-30
Pull in content including G-Cloud 9

TheDoubleK:
This will allow the "G-Cloud 9 is coming" message to show when we set the framework up in the DB.


idavidmcdonald:
👍

PR #424, opened 2017-01-30
[WIP] changes required to launch of DOS2

danielknell:
Pull request to track the changes we need to deploy as DOS2 launches.


danielknell:
as we are changing how the opertunities page works this is not relevant

PR #426, opened 2017-02-01
filter failed responses from the ODS

danielknell:
bugfix to earlier work to filter out responses that do not meet the essential requirements

### DIFF
--- a/tests/buyers/views/test_buyers.py
+++ b/tests/buyers/views/test_buyers.py
@@ -1969,6 +1969,14 @@ class TestDownloadBriefResponsesView(BaseApplicationTest):
                 "blah": [False, True]
             }
         ]
+        self.bad_response = {
+            "supplierName": "Tea Boy Ltd.",
+            "respondToEmailAddress": "teaboy@example.com",
+            "availability": "Tomorrow",
+            "essentialRequirementsMet": False,
+            "dayRate": "1000",
+            "blah": [False, True]
+        }
 
     def teardown_method(self, method):
         self.instance = None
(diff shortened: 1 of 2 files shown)


tests/buyers/views/test_buyers.py
```diff
@@ -1969,6 +1969,14 @@ def setup_method(self, method):
                 "blah": [False, True]
             }
         ]
+        self.bad_response = {
+            "supplierName": "Tea Boy Ltd.",
+            "respondToEmailAddress": "teaboy@example.com",
+            "availability": "Tomorrow",
+            "essentialRequirementsMet": False,
```
idavidmcdonald: In our flow, if you try to answer no to essential requirements met, we don't actually save your answer to the brief response. So we would actually never expect essential requirements met to be set and be False.

idavidmcdonald: After some further digging, because of this the user could never 'submit' a brief response that had essential requirements met as False. Therefore because the brief response is not submitted it will not be returned when you request all brief responses from the API (unless you specified otherwise).

idavidmcdonald: If what I've said above is all correct (do follow through and check the logic yourself) I believe it means this pull request is not needed because we would never expect data in the format you are fixing.

danielknell: @idavidmcdonald Hrm,  is this desired?  The current term system lets you submit a useless response that the buyer can never access,  is there a reason for this?  Probably worth investigating? 

I am all in favour of not needing this but maybe we need to check there's not some reasoning we are not aware of for the old functionality?

idavidmcdonald: I believe we don't want to save the information because it's an invalid form input. Therefore they have never actually applied and so therefore the buyer would never need to access it.

Sorry what do you mean by 'term system'

In terms of the old flow, I believe you could always submit the brief response and all the data would be saved but you could submit one that had 'failed' i.e. you had no answer Yes to all the essential requirements.

PR #428, opened 2017-02-01
support in ODS for missing keys in brief response

danielknell:
add support for missing keys in the brief response

will now skip over complex questions with no content and insert empty strings for other questions


idavidmcdonald:
👍 nice

PR #429, opened 2017-02-03
Add dos2 manifests and other fixes

danielknell:
some minor tweaks around dos2 content


danielknell:
not relevant any more

PR #430, opened 2017-02-03
key opportunities page of framework.framework instead of framework.slug

danielknell:
this will switch the opportunities endpoints to use framework.framework instead of framework.slug ~, it cant be merged yet though, it needs a data migration and API changes to switch the value of framework from `dos` to `digital-outcomes-and-specialists` first~.

both endpoints now use the framework.framework slug instead of the framework.slug slug

### DIFF
--- a/app/main/forms/brief_forms.py
+++ b/app/main/forms/brief_forms.py
@@ -22,11 +22,14 @@ class BriefSearchForm(Form):
         try:
             # popping this kwarg so we don't risk it getting fed to wtforms default implementation which might use it
             # as a data field if there were a name collision
-            framework = kwargs.pop("framework")
-            self._framework_slug = framework["slug"]
-            self.lot.choices = tuple((lot["slug"], lot["name"],) for lot in framework["lots"] if lot["allowsBrief"])
+            frameworks = kwargs.pop("frameworks")
+            self._framework_slug = ",".join(v["slug"] for v in frameworks)
+            seen = set()
+            self.lot.choices = tuple((v['slug'], v['name']) for f in frameworks
+                                     for v in f['lots'] if v['allowsBrief'] and
+                                     not (v['slug'] in seen or seen.add(v['slug'])))
         except KeyError:
-            raise TypeError("Expected keyword argument 'framework' with framework information")
+            raise TypeError("Expected keyword argument 'frameworks' with framework information")
         try:
             # data_api_client argument only needed so we can fit in with the current way the tests mock.patch the
             # the data_api_client directly on the view. would be nice to able to use the global reference to this
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -87,7 +87,7 @@ def terms_and_conditions():
 def get_brief_by_id(framework_slug, brief_id):
     briefs = data_api_client.get_brief(brief_id)
     brief = briefs.get('briefs')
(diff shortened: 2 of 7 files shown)


app/main/views/marketplace.py
```diff
@@ -68,7 +68,7 @@ def terms_and_conditions():
 def get_brief_by_id(framework_slug, brief_id):
```
idavidmcdonald: We should change `framework_slug` to `framework_framework` or similar, just so we don't get confused between a individual framework and a framework framework. Unfortunately this is all down to poor naming of framework_framework but that's a bigger problem...

tests/main/views/test_marketplace.py
```diff
@@ -526,16 +526,41 @@ def setup_method(self, method):
...
+                'framework': "digital-outcomes-and-specialists",
+                'lots': [
+                    {'name': 'Lot 1', 'slug': 'lot-one', 'allowsBrief': True},
+                    {'name': 'Lot 2', 'slug': 'lot-two', 'allowsBrief': False},
+                    {'name': 'Lot 3', 'slug': 'lot-three', 'allowsBrief': True},
+                    {'name': 'Lot 4', 'slug': 'lot-four', 'allowsBrief': True},
+                ]
+            },
+            {
+                'name': "Foobar",
```
idavidmcdonald: Good test setup, covers all the bases nicely

app/main/views/marketplace.py
```diff
@@ -148,7 +151,7 @@ def list_opportunities(framework_slug):
         next_link_args.setlist("page", api_next_link_args.get("page") or ())
 
     return render_template('search/briefs.html',
-                           framework=framework,
+                           framework=frameworks[-1],
```
idavidmcdonald: Can we rely on the last one being the 'first' framework? Obviously your unit tests are fine as we have mocked the API client but can we always be sure what order will come from the API?

When hitting PREVIEW api, the frameworks seem to come back in an order I can't quite make out...

API side at least, this is what is happening
https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/frameworks.py#L22

If we are able to rely on this behaviour, I wonder if it's worth commenting to say so...

app/main/forms/brief_forms.py
```diff
@@ -22,11 +22,14 @@ def __init__(self, *args, **kwargs):
         try:
             # popping this kwarg so we don't risk it getting fed to wtforms default implementation which might use it
             # as a data field if there were a name collision
-            framework = kwargs.pop("framework")
-            self._framework_slug = framework["slug"]
-            self.lot.choices = tuple((lot["slug"], lot["name"],) for lot in framework["lots"] if lot["allowsBrief"])
+            frameworks = kwargs.pop("frameworks")
+            self._framework_slug = ",".join(v["slug"] for v in frameworks)
```
idavidmcdonald: Worth pluralising this?

danielknell: would rather leave it as is, it maps the api param.

idavidmcdonald: Fair point, keep as is

app/main/views/marketplace.py
danielknell: ordering by id now

idavidmcdonald:
Nice work 👍 should be good when tests pass

PR #435, opened 2017-02-07
Load DOS2 display brief manifest

idavidmcdonald:
Will need this for DOS2

Welcome thoughts about how to do test this (i.e. how our tests would know if we were missing this manifest).


idavidmcdonald:
Have added to the tech debt board

PR #438, opened 2017-02-13
Pass slug of newest live dos framework to index

Wynndow:
For this story on Pivotal: [https://www.pivotaltracker.com/story/show/139283877](https://www.pivotaltracker.com/story/show/139283877)

Currently the home page has links to create DOS opportunities for the different lots. The framework slug used in the url is hardcoded to `digital-outcomes-and-specialists`.

When DOS2 goes live this slug needs to be `digital-outcomes-and-specialists-2`.

This PR updates the buyer app to check for all live DOS frameworks and passes the slug of the most recent (judged by the greatest ID) to the index template to correctly render the links. This handles the edge case of when we have two DOS frameworks live (the short period when transitioning that DOS1 is live at the same time as DOS 2).

FUNCTIONAL TESTS:

Those on the master branch will break when we transition, but this is due to briefs that are created as part of the flow being on the DOS1 framework. Can easily be updated.

Those on the jenkins branch also break a bit. I'm investigating why...

UPDATE:

Functional tests will break on master and live only when DOS 1 is expired. They'll be fine of DOS1 and DOS2 are live. They break when DOS1 is expired due to a mix of being unable to create briefs on an expired framework, and hardcoded references to dos1 links (especially in the admin app). Should be fairly straightforward to fix, I've got a list on my machine of where/why they're breaking.

### DIFF
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -186,7 +186,6 @@ class TestHomepageSidebarMessage(BaseApplicationTest):
         _frameworks = []
 
         for index, framework_slug_and_status in enumerate(framework_slugs_and_statuses):
-
             framework_slug, framework_status = framework_slug_and_status
             _frameworks.append({
                 'framework': 'framework',
@@ -258,7 +257,6 @@ class TestHomepageSidebarMessage(BaseApplicationTest):
 
         framework_slugs_and_statuses = [
             ('g-cloud-7', 'pending'),
-            ('digital-outcomes-and-specialists', 'live')
         ]
         framework_messages = [
             u"G‑Cloud 7 is closed for applications",
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -26,9 +26,19 @@ class TestApplication(BaseApplicationTest):
             'Find out more about cookies</a></p>' in res.get_data(as_text=True)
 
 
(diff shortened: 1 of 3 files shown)


tests/main/views/test_marketplace.py
```diff
@@ -146,7 +257,8 @@ def test_homepage_sidebar_message_exists_gcloud_8_open(self):
     def test_homepage_sidebar_message_exists_g_cloud_7_pending(self):
 
         framework_slugs_and_statuses = [
-            ('g-cloud-7', 'pending')
+            ('g-cloud-7', 'pending'),
+            ('digital-outcomes-and-specialists', 'live')
```
idavidmcdonald: Sorry, why has this been added?

Wynndow: Should have been removed, a hangover from a previous approach.

tests/main/views/test_marketplace.py
```diff
@@ -75,6 +186,7 @@ def _find_frameworks(framework_slugs_and_statuses):
         _frameworks = []
 
         for index, framework_slug_and_status in enumerate(framework_slugs_and_statuses):
+
```
idavidmcdonald: Should there be a new line here?

idavidmcdonald:
👍

tests/main/views/test_marketplace.py
Wynndow: Nope! Taken care of.

PR #439, opened 2017-02-15
Link to correct call off contract and framework agreement on requirements page

Wynndow:
For this story on Pivotal [https://www.pivotaltracker.com/story/show/139281415](https://www.pivotaltracker.com/story/show/139281415)

Will need a [PR on the apiclient](https://github.com/alphagov/digitalmarketplace-apiclient/pull/68) and a [PR on the frameworks repo](https://github.com/alphagov/digitalmarketplace-frameworks/pull/377) merging before tests will go green.

This PR adds meta information to give a buyer context when looking at one of their requirements page. It's modelled roughly on a similar thing we have when viewing a G-Cloud service, see the one third column [here](https://www.digitalmarketplace.service.gov.uk/g-cloud/services/7959974558158858) for an example.

This meta information is only displayed if the brief is published and includes the name of the framework the brief is on. This will help the buyer understand what framework they've published their link under. It also links them to the framework agreement for that framework for further context.

The link to the call off contract is also updated depending on the framework. The link text is also changed to provide further context.

URL's have been added to the frameworks repo messages for both DOS frameworks and are pulled in and used from there.

Screen shots!

screen shot 2017-02-15 at 11 09 18

screen shot 2017-02-15 at 11 09 22

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -127,6 +127,9 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
     sections = content.summary(brief)
     delete_requested = True if request.args.get('delete_requested') else False
 
+    content_loader.load_messages(brief['frameworkSlug'], ['urls'])
+    call_off_contract_url = content_loader.get_message(brief['frameworkSlug'], 'urls', 'call_off_contract_url')
+
     completed_sections = {}
     for section in sections:
         required, optional = count_unanswered_questions([section])
@@ -149,6 +152,7 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
         completed_sections=completed_sections,
         step_sections=[section.step for section in sections if hasattr(section, 'step')],
         delete_requested=delete_requested,
+        call_off_contract_url=call_off_contract_url,
     ), 200
 
 
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -129,6 +129,7 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
 
     content_loader.load_messages(brief['frameworkSlug'], ['urls'])
     call_off_contract_url = content_loader.get_message(brief['frameworkSlug'], 'urls', 'call_off_contract_url')
+    framework_agreement_url = content_loader.get_message(brief['frameworkSlug'], 'urls', 'framework_agreement_url')
 
     completed_sections = {}
     for section in sections:
@@ -153,6 +154,7 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
         step_sections=[section.step for section in sections if hasattr(section, 'step')],
         delete_requested=delete_requested,
         call_off_contract_url=call_off_contract_url,
+        framework_agreement_url=framework_agreement_url,
     ), 200
 
 
--- a/app/templates/buyers/_requirements_meta.html
+++ b/app/templates/buyers/_requirements_meta.html
@@ -1,0 +1,8 @@
+<div id="requirements-meta">
+  <h2 class="sidebar-heading">Published</h2>
+  <p class="sidebar-content">{{ brief.publishedAt|dateformat }}</p>
+  <h2 class="sidebar-heading">Closing</h2>
+  <p class="sidebar-content">Tuesday 27 October 2017</p>
+  <h2 class="sidebar-heading">Framework</h2>
+  <p class="framework-name">{{ brief.frameworkName }}</p>
+</div>
--- a/app/templates/buyers/_requirements_meta.html
+++ b/app/templates/buyers/_requirements_meta.html
@@ -9,5 +9,5 @@
     <p class="sidebar-content">{{ brief.applicationsClosedAt|dateformat }}</p>
   {% endif %}
   <h2 class="sidebar-heading">Framework</h2>
-  <p class="framework-name">{{ brief.frameworkName }}</p>
+  <p class="framework-name"><a href="{{ framework_agreement_url }}">{{ brief.frameworkName }}</a></p>
 </div>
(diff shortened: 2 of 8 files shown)


app/templates/buyers/_requirements_meta.html
```diff
@@ -0,0 +1,13 @@
...
+  <p class="sidebar-content">{{ brief.publishedAt|dateformat }}</p>
+  {% if brief.status == 'closed' %}
+    <h2 class="sidebar-heading">Closed</h2>
+    <p class="sidebar-content">{{ brief.applicationsClosedAt|dateformat }}</p>
+  {% else %}
+    <h2 class="sidebar-heading">Closing</h2>
+    <p class="sidebar-content">{{ brief.applicationsClosedAt|dateformat }}</p>
+  {% endif %}
+  <h2 class="sidebar-heading">Framework</h2>
+  <p class="framework-name"><a href="{{ framework_agreement_url }}">{{ brief.frameworkName }}</a></p>
```
idavidmcdonald: It does at least make your tests easier though...

app/buyers/views/buyers.py
```diff
@@ -127,6 +127,10 @@ def view_brief_overview(framework_slug, lot_slug, brief_id):
     sections = content.summary(brief)
     delete_requested = True if request.args.get('delete_requested') else False
 
+    content_loader.load_messages(brief['frameworkSlug'], ['urls'])
+    call_off_contract_url = content_loader.get_message(brief['frameworkSlug'], 'urls', 'call_off_contract_url')
```
idavidmcdonald: I forgot to answer my question before submitting the review....

idavidmcdonald: But maybe a quick check to see if the url doesn't exist then does `get_message` return None or does the app fall over?

Will only really make sure we can't put a new framework live without having the correct messages set up....

Wynndow: It will return `None` if it can't find the key it's after and the app stays up. You get a weird thing on the requirements page though. If for example the contract url wasn't set and so `call_off_contract_url` was set to None, the link would point at `http://localhost/buyers/frameworks/digital-outcomes-and-specialists/requirements/digital-specialists/None`.

I think this is ok? It just means that new frameworks need url's set?

app/templates/buyers/_requirements_meta.html
```diff
@@ -0,0 +1,13 @@
+<div id="requirements-meta">
+  <h2 class="sidebar-heading">Published</h2>
+  <p class="sidebar-content">{{ brief.publishedAt|dateformat }}</p>
+  {% if brief.status == 'closed' %}
+    <h2 class="sidebar-heading">Closed</h2>
```
Wynndow: Nice...

app/buyers/views/buyers.py
idavidmcdonald: See if you can find another place where we rely on urls (maybe signing framework agreements) and follow their convention. I'm happy either way as we aren't being less protective then we are in other places

app/templates/buyers/_requirements_meta.html
Wynndow: I was mostly cribbing off the meta stuff on the g-cloud page, and it's done like this there.

I think it would actually make my tests easier if the classes were the same.

Change or keep...your call?

app/templates/buyers/brief_overview.html
```diff
@@ -211,7 +214,11 @@ <h2 class="instruction-list-item-body">
         {% endwith %}
       </ol>
     </div>
-
+    <div class="column-one-third">
+      {% if brief.publishedAt %}
```
Wynndow: Yep, had that conversation with Ralph. It's not confirmed until it's publishes so we don't show it.

app/templates/buyers/_requirements_meta.html
idavidmcdonald: I guess follow convention...

I'm not too fussed either way so ... your call?

app/buyers/views/buyers.py
Wynndow: Other places (supplier framework agreement page, framework dashboard, the homepage) don't do any error handling. They just set a variable and pass it to the template.

idavidmcdonald: Cool, all fine then

tests/buyers/views/test_buyers.py
```diff
@@ -1528,6 +1528,91 @@ def test_links_to_sections_go_to_the_correct_pages_whether_they_be_sections_or_q
...
+
+                    sidebar_headings = document.xpath('//*[@class="sidebar-heading"]/text()')
+                    sidebar_content = document.xpath('//*[@class="sidebar-content"]/text()')
+                    framework_name_content = document.xpath('//*[@class="framework-name"]/a/text()')[0]
+                    contract_link_text = document.xpath('//main[@id="content"]//ul/li/a/text()')[-1]
+
+                    assert sidebar_headings == ['Published', heading, 'Framework']
+                    assert sidebar_content == ['Tuesday 29 March 2016', 'Thursday 7 April 2016']
+                    assert framework_name_content == framework_name
+                    assert contract_link_text == "View the {} contract".format(framework_name)
```
Wynndow: My thinking was keep content in one test, and link destinations in another. Something about the contracts may move but the link text should stay the same so separate tests? Not sure. 

I'll move it.

tests/buyers/views/test_buyers.py
```diff
@@ -1528,6 +1528,91 @@ def test_links_to_sections_go_to_the_correct_pages_whether_they_be_sections_or_q
             assert section_4_link[0].get('href').strip() == \
                 '/buyers/frameworks/digital-outcomes-and-specialists/requirements/digital-specialists/1234/section-4'
 
+    def test_shows_correct_content_for_open_and_closed_dos_1_and_2_briefs(self, data_api_client):
```
Wynndow: Stuck one in

app/templates/buyers/_requirements_meta.html
Wynndow: I'm gonna go keep. It means if we wanted to do something different with the framework name in future it makes it easier.

idavidmcdonald:
👍

PR #440, opened 2017-02-16
Dos1 functional when expired

Wynndow:
Part of this story: [https://www.pivotaltracker.com/story/show/139283847](https://www.pivotaltracker.com/story/show/139283847)

When DOS1 expires and DOS2 comes live, there are some key areas for the DOS1 framework that it is important still work. Expiring the framework broke some of them. The ones identified that this PR fixes are listed below.

- Viewing a briefs overview
- Viewing the brief responses page
- Downloading an ods/csv of the responses.
- Viewing a live briefs timeline
- Viewing supplier questions
- Adding/answering supplier questions

They were all failing due to a helper called `get_framework_and_lot` which took in the required framework status as a parameter, and was restricted to only one status. That helper has been updated to be able to take multiple statuses. Previously we were only allowing the above actions for a `live` framework. We now allow them for `live` and `expired` frameworks. We do NOT allow things like creating/publishing/editing briefs for `expired` frameworks however.

There is a thought that we may not even need to check the framework status for some of these things. For example, viewing a briefs overview is restricted to briefs on a framework that is `live` or `expired`. Why? A brief cannot be created unless the framework is live, so to explicitly disallow the viewing of a brief that is on a framework thats status is `coming`, `open`, `pending`, or  `standstill` seems overkill. No harm, but why bother?

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -117,7 +117,12 @@ def create_new_brief(framework_slug, lot_slug):
 @buyers.route('/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>', methods=['GET'])
 def view_brief_overview(framework_slug, lot_slug, brief_id):
     framework, lot = get_framework_and_lot(
-        framework_slug, lot_slug, data_api_client, allowed_statuses='live', must_allow_brief=True)
+        framework_slug,
+        lot_slug,
+        data_api_client,
+        allowed_statuses=['live', 'expired'],
+        must_allow_brief=True
+    )
     brief = data_api_client.get_brief(brief_id)["briefs"]
 
     if not is_brief_correct(brief, framework_slug, lot_slug, current_user.id):
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -51,7 +51,7 @@ def buyer_dashboard():
 def start_new_brief(framework_slug, lot_slug):
 
     framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client,
-                                           allowed_statuses='live', must_allow_brief=True)
+                                           allowed_statuses=['live'], must_allow_brief=True)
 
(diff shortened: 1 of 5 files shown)


app/buyers/views/buyers.py
```diff
@@ -51,7 +51,7 @@ def buyer_dashboard():
 def start_new_brief(framework_slug, lot_slug):
 
     framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client,
-                                           status='live', must_allow_brief=True)
+                                           allowed_statuses='live', must_allow_brief=True)
```
danielknell: method expects a list or tuple and this is a string?

app/buyers/views/buyers.py
```diff
@@ -73,7 +73,7 @@ def start_new_brief(framework_slug, lot_slug):
 def create_new_brief(framework_slug, lot_slug):
 
     framework, lot = get_framework_and_lot(framework_slug, lot_slug, data_api_client,
-                                           status='live', must_allow_brief=True)
+                                           allowed_statuses='live', must_allow_brief=True)
```
danielknell: same as above

app/buyers/views/buyers.py
```diff
@@ -184,7 +195,7 @@ def view_brief_section_summary(framework_slug, lot_slug, brief_id, section_slug)
     '/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/edit/<section_slug>/<question_id>',
     methods=['GET'])
 def edit_brief_question(framework_slug, lot_slug, brief_id, section_slug, question_id):
-    get_framework_and_lot(framework_slug, lot_slug, data_api_client, status='live', must_allow_brief=True)
+    get_framework_and_lot(framework_slug, lot_slug, data_api_client, allowed_statuses='live', must_allow_brief=True)
```
danielknell: and again

Wynndow:
@danielknell Should all be fixed up now

PR #441, opened 2017-02-16
Don't override margin styles in homepage sidebar

TheDoubleK:
Styles in the app for `supplier-messages` were overriding the styles of things from the frontend toolkit.

Ralph doesn't like this.

Spot the difference in the temporary message:
## BEFORE
screen shot 2017-02-16 at 11 31 50

## AFTER
screen shot 2017-02-16 at 11 32 38


tombye:
Deleted code = 👍

PR #448, opened 2017-02-21
Update content on shortlisting page

Wynndow:
For these stories on Pivotal:
[https://www.pivotaltracker.com/story/show/139290411](https://www.pivotaltracker.com/story/show/139290411)
[https://www.pivotaltracker.com/story/show/138994639](https://www.pivotaltracker.com/story/show/138994639)

This PR updates content displayed when a buyer view the shortlist page for a brief. It also changes the way that page decides to render content for a legacy style brief or a new flow style brief. Details below.

When there are no responses to a brief the content displayed needs to be
different depending on if the brief is a legacy brief (no evidence
required for responses), or a new flow brief (evidence required in
responses).

Previously the split between new and old style briefs was being
determined by the `requirementsMet` key of the responses. If there are
no responses however this breaks down and we need another way to
determine what kind of brief we're dealing with.

This can be done by comparing the `publishedAt` date of the brief with
the feature flag set in the config.

This may not be a permanent solution. We may want to remove that flag at
some point in the future. We could possibly hard code the date the new
flow went live in the logic.

Screenshots:

Legacy brief, no responses

screen shot 2017-02-21 at 15 51 16

Legacy brief with valid responses

screen shot 2017-02-21 at 15 41 45

New flow style brief, no responses

screen shot 2017-02-21 at 16 02 08

New flow style brief, with responses

screen shot 2017-02-21 at 15 42 28

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -298,19 +298,18 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
 
     brief_responses = data_api_client.find_brief_responses(brief_id)['briefResponses']
 
-    met = None
+    new_flow_brief = (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
+                      <= datetime.strptime(brief['publishedAt'][0:10], "%Y-%m-%d"))
 
     counter = Counter()
 
     for response in brief_responses:
-        met = response.get('essentialRequirementsMet', None)
-
         counter[all(response['essentialRequirements'])] += 1
 
     return render_template(
         "buyers/brief_responses.html",
         response_counts={"failed": counter[False], "eligible": counter[True]},
-        ods=met is not None,  # will be None for legacy responses
+        new_flow_brief=new_flow_brief,
         brief=brief
     ), 200
 
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -298,8 +298,10 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
 
     brief_responses = data_api_client.find_brief_responses(brief_id)['briefResponses']
 
-    new_flow_brief = (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
-                      <= datetime.strptime(brief['publishedAt'][0:10], "%Y-%m-%d"))
+    brief_responses_require_evidence = (
+        datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
+        <= datetime.strptime(brief['publishedAt'][0:10], "%Y-%m-%d")
+    )
 
     counter = Counter()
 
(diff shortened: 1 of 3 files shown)


app/buyers/views/buyers.py
```diff
@@ -298,19 +298,18 @@ def view_brief_responses(framework_slug, lot_slug, brief_id):
 
     brief_responses = data_api_client.find_brief_responses(brief_id)['briefResponses']
 
-    met = None
+    new_flow_brief = (datetime.strptime(current_app.config['FEATURE_FLAGS_NEW_SUPPLIER_FLOW'], "%Y-%m-%d")
```
idavidmcdonald: I might suggest something like 'brief_response_has_evidence' so instead of using something like new/old/legacy we have something that instead describes what is different about the brief response and it's data.

Wynndow: yep, makes sense.

tests/buyers/views/test_buyers.py
```diff
@@ -1932,16 +1934,19 @@ def test_api_error(self, data_api_client):
...
             slug='digital-outcomes-and-specialists',
             status='live',
             lots=[
                 api_stubs.lot(slug='digital-outcomes', allows_brief=True),
             ]
         )
-        data_api_client.get_brief.return_value = api_stubs.brief(lot_slug="digital-outcomes")
+
+        brief_stub = api_stubs.brief(lot_slug="digital-outcomes")
+        brief_stub['briefs'].update({'publishedAt': '2017-01-21T12:00:00.000000Z'})
```
idavidmcdonald: This common test is shared between the new and old flow test classes but this date is just applicable for one. 

You could put something like `self.briefPublishedAt` value on both the `TestViewBriefResponsesPageForLegacyBrief` and `TestViewBriefResponsesPageForNewFlowBrief` and then in this test set `brief_stub['briefs'].update({'publishedAt': self.briefPublishedAt})`.

Maybe this date has just been put here because you need published at to be "something" but the date doesn't actually matter for the common tests, however I think that means your tests then rely on your implementation

tests/buyers/views/test_buyers.py
```diff
@@ -1964,7 +1973,10 @@ def test_page_does_not_pluralise_for_single_response(self, data_api_client):
...
+        data_api_client.get_brief.return_value = brief_stub
```
idavidmcdonald: It feels like all these tests have a lot of duplication in terms of their set up (not just the stuff you've added but generally). Lots of frameworks and briefs being set repeatedly. You could move some of these into a setup method

idavidmcdonald:
Oh, excellent PR message btw, the 4 screenshots much appreciated

idavidmcdonald:
Also spotted a bit of a weird inconsistency that might be nice to address/figure out if it is a problem
AbstractViewBriefResponsesPage - The first two tests are looking for response to brief ID 1 rather than the rest of the tests which look for responses to brief ID 1234

idavidmcdonald:
Oh and another tiny one from previous code, in the abstract class we have a test called `test_page_does_not_show_csv_download_link_if_brief_open`. However this test is abstract so ideally the class name wouldn't have the word `csv` in because this feels misleading as it could be a CSV or ODS.

Sorry for finding other random bits in the code....

Wynndow:
The ID thing is because we're mocking the response from the api client. You could actually chuck any old number in there and the tests still pass as you'll always get back the brief stub you've asked for. I've updated to 1234 to keep those tests consistent.

tests/buyers/views/test_buyers.py
Wynndow: Removed a bunch to a setup!

Wynndow:
Have fixed up the reference to `csv_link` in `AbstractViewBriefResponsesPage`, just referred to it as `download_link.`

Wynndow:
Cool. I'll leave the rebase till it's ready to go.

PR #450, opened 2017-02-22
Bring in new toolkit with new input fields

pcraig3:
This pull request

- updates the govuk_frontend toolkit to the latest
- updates the digitalmarketplace frontend toolkit to the latest
  - [brings in the new radios/checkboxes](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/302)
- updates the homepage beta header colours (blue text on a white background)
- treats links with a `role="button"` as buttons (ie, "Apply for this opportunity" link)
- removes some unused classnames
- brings the forgotten password link closer to the password input field on the login screen

💥  Boom 💥

PR #452, opened 2017-02-23
Revert "Bring in new toolkit with new input fields"

pcraig3:
Reverts alphagov/digitalmarketplace-buyer-frontend#450

The functional tests are breaking and I can't figure out how exactly to fix them, so we're reverting the changes until I can get them to pass locally.


idavidmcdonald:
👍

PR #453, opened 2017-02-24
Revert "Revert "Bring in new toolkit with new input fields""

pcraig3:
Reverts alphagov/digitalmarketplace-buyer-frontend#452

So yesterday, we merged in these changes (#450) and then our functional tests started to break.
Upon investigation, we found that it was not able to find checkbox/radio elements [when calling its default 'find_all' method and looking for input fields](https://github.com/alphagov/digitalmarketplace-functional-tests/blob/19f5fa5c7aa2db20dddb921e96869a06614864ac/features/support/form_helpers.rb#L52).  

Since the default input elements are really small, the new elements actually hide the inputs and draw a custom interface using the `::before` and `::after` psuedo-elements. @robinwhittleton talks about how [hiding the elements was an accessibility challenge](https://gdstechnology.blog.gov.uk/2016/12/13/improving-forms-with-new-radios-and-checkboxes/), but now they work with real-world accessibility software.

New inputs have an `opacity` of `0`.
screen shot 2017-02-24 at 10 55 22

Unfortunately, [selenium doesn't find hidden elements](http://stackoverflow.com/questions/22110282/how-to-click-on-hidden-element-in-selenium-webdriver), which means it can't find our inputs.

Since we can't make large interface changes while G-Cloud is open for suppliers, but we can make testing changes, the decision has been made to merge this now and fix the tests later.

### DIFF
--- a/app/assets/javascripts/_analytics.js
+++ b/app/assets/javascripts/_analytics.js
@@ -1,5 +1,8 @@
 //= include ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/google-analytics-universal-tracker.js
 //= include ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/analytics.js
+
+; // JavaScript in the govuk_frontend_toolkit doesn't have trailing semicolons
+
 //= include analytics/_register.js
 //= include analytics/_pageViews.js
 //= include analytics/_events.js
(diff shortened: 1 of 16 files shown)


idavidmcdonald:
Based on the fact this is a revert of a revert and the original PR was thumbs up by Tom, I think this should be good after a rebase 👍

PR #455, opened 2017-02-28
Remove legacy behaviour to link to old supplier flow

idavidmcdonald:
As all briefs published will be applied to using the old flow we now no longer need to point users to start a brief application using the old flow so we remove this functionality.

Note, tests now no longer need to cover three different cases with very slight differences but only cover the case for the new supplier flow, therefore have had a minor restructure and tidy up.


pcraig3:
Looks good after a rebase.  👍 

Nice to finally be able to get rid of old stuff.

PR #456, opened 2017-03-06
Latest g9 content homepage message

TheDoubleK:
This pulls in the latest content - net effect is hopefully just that the "G9 open" message changes from GMT to BST.

i.e it will now look like:

screen shot 2017-03-06 at 13 48 53

In doing this there was a major version change for both the content and content loader modules.

This created some issues for downloading responses to briefs (the ODS spreadsheet of responses).

I have fixed things so the unit tests pass, but once merged it will definitely need to be checked on Preview that downloading brief responses still works correctly (i.e. that the file is generated and downloads OK, and that the data is as expected).


idavidmcdonald:
This sucks but lets do it ;) 👍

PR #458, opened 2017-03-15
404 brief responses page if brief is not closed

idavidmcdonald:
For this pivotal story - https://www.pivotaltracker.com/story/show/138995271

Previously we were somewhat supporting this page for live briefs even though we didn't link to the page. We have removed this functionality because there were bugs on the page for live briefs and we had no evidence that it is worth the maintenance cost as the page is effectively hidden unless a user 'hacks' the url.


TheDoubleK:
Bugs?  In a template I made?  If you can't be specific about them then I don't believe you :stuck_out_tongue: 

Changes here look reasonable though :+1:

PR #460, opened 2017-03-21
Remove hard-coding to G-Cloud 6 of content (filters, service page).

galund:
A one-line fix that became a bit more than that...
 - work in preparation for G-Cloud 9 category filtering / go-live;
 - G-Cloud 4 and 5 services are to continue to use G-Cloud 6
   content and manifests (new config controls this);
 - refactor code to get 'latest' iteration of a framework that
   already existed for DOS frameworks;
 - tests are largely unaltered, we probably want to write new
   ones relating to G-Cloud 9 at some point.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -9,6 +9,7 @@ from dmutils.user import User
 
 from config import configs
 
+
 login_manager = LoginManager()
 data_api_client = dmapiclient.DataAPIClient()
 search_api_client = dmapiclient.SearchAPIClient()
@@ -16,10 +17,6 @@ feature_flags = flask_featureflags.FeatureFlag()
 csrf = CsrfProtect()
 
 content_loader = ContentLoader('app/content')
-content_loader.load_manifest('g-cloud-6', 'services', 'search_filters')
-content_loader.load_manifest('g-cloud-6', 'services', 'display_service')
-content_loader.load_manifest('digital-outcomes-and-specialists', 'briefs', 'display_brief')
-content_loader.load_manifest('digital-outcomes-and-specialists-2', 'briefs', 'display_brief')
 
 
 def create_app(config_name):
@@ -33,6 +30,16 @@ def create_app(config_name):
         login_manager=login_manager,
         search_api_client=search_api_client
     )
+
+    for framework_data in data_api_client.find_frameworks().get('frameworks'):
+        if not framework_data['slug'] in application.config.get('DM_FRAMEWORK_CONTENT_MAP', {}):
+            if framework_data['framework'] == 'g-cloud':
+                if framework_data['status'] != 'expired':
+                    content_loader.load_manifest(framework_data['slug'], 'services', 'search_filters')
+                # we need to be able to display old services, even on expired frameworks
+                content_loader.load_manifest(framework_data['slug'], 'services', 'display_service')
+            elif framework_data['framework'] == 'digital-outcomes-and-specialists':
+                content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
 
     from .main import main as main_blueprint
     from .status import status as status_blueprint
--- a/app/main/helpers/framework_helpers.py
+++ b/app/main/helpers/framework_helpers.py
@@ -1,0 +1,9 @@
+def get_latest_live_framework(all_frameworks, framework_type):
+    try:
+        latest = max(
+            (f for f in all_frameworks if f['status'] == 'live' and f['framework'] == framework_type),
+            key=lambda f: f['id'],
+        )
+        return latest
+    except ValueError:  # max of empty iterable
+        return None
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -68,6 +68,11 @@ def get_service_by_id(service_id):
 
         service_data = service['services']
         framework_slug = service_data['frameworkSlug']
+
+        # Some frameworks don't actually have framework content of their own (e.g. G-Cloud 4 and 5) - they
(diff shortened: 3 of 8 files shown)


config.py
```diff
@@ -71,6 +71,11 @@ class Config(object):
     DM_APP_NAME = 'buyer-frontend'
     DM_DOWNSTREAM_REQUEST_ID_HEADER = 'X-Amz-Cf-Id'
 
+    DM_FRAMEWORK_CONTENT_MAP = {
```
TheDoubleK: An alternative to this (not necessarily a better alternative) might be to try to load the relevant manifest and then default to G6 if it can't be loaded?

Or - and this might be really mad - create G4 and G5 folders in frameworks repo and copy the required G6 manifest there.

app/main/helpers/framework_helpers.py
```diff
@@ -0,0 +1,9 @@
+def get_latest_live_framework(all_frameworks, framework_type):
+    try:
+        latest = max(
+            (f for f in all_frameworks if f['status'] == 'live' and f['framework'] == framework_type),
+            key=lambda f: f['id'],
```
TheDoubleK: Using ID for this really stinks.  (Yes, I know that's how it was doing it before, and no, I don't have a better suggestion right now).

config.py
TheDoubleK: I like the way putting this into config gives the impression that we don't have hard-coded framework slugs in our frontend apps while still having hard-coded framework slugs in our apps :stuck_out_tongue: 

I think it's worth a short comment in the config file to explain what this is used for, because it's pretty esoteric otherwise.

app/__init__.py
```diff
@@ -34,6 +31,16 @@ def create_app(config_name):
         search_api_client=search_api_client
     )
 
+    for framework_data in data_api_client.find_frameworks().get('frameworks'):
```
TheDoubleK: Interesting issue here - we will need to trigger a re-start of the application server (probably by re-deploying the app without any code changes) to load any new frameworks that are created in future.

This isn't the end of the world: previously we needed to deploy a new version of the app with the manifests explicitly loaded.

But it could be a tricky gotcha at some point.

config.py
galund: Yeah - I needed something that could be shared between the loading up of the content (i.e. into the app on startup), and the getting of it (e.g. in the service page view). Hard-coding a bunch of stuff about G4/5/6 in both places seemed like I hadn't really made much progress. But I totally dig that this is really still the same thing.

On the second point, I don't know if there's a memory impact of loading up 5 copies of the same content (G4/5/6/7/8) but it sure does seem a bit wrong :-D

app/main/views/g_cloud.py
```diff
@@ -66,9 +67,14 @@ def get_service_by_id(service_id):
             abort(404, "Service ID '{}' can not be found".format(service_id))
 
         service_data = service['services']
+        framework_slug = service_data['frameworkSlug']
+        override_framework_slug = current_app.config.get('DM_FRAMEWORK_CONTENT_MAP').get(framework_slug)
```
TheDoubleK: Again, I think a short comment here to explain why it's overriding anything wouldn't hurt.

TheDoubleK:
I agree it's not the prettiest, but I don't have any obviously better alternatives at all.  It definitely feels like some kind of step forward to get rid of the explicit `'g-cloud-6'` load everywhere.

Cajole the tests into submission, add a couple of pertinent comments, and I can live with this.

We should definitely do something similar in the other frontend apps to dynamically pick on start-up which manifests to load.

app/__init__.py
galund: Yeah, hopefully mitigated by the fact that we're loading up 'open' and 'standstill' frameworks too - chances are we'll have had at least one restart between the new framework first appearing (e.g. becoming open) and it actually going into the 'live' state.

Dunno that applies to DOS frameworks, mind.

I'm still slightly worried about the long-term implications for memory usage seeing as each framework a megabyte or so of data. Maybe that's actually peanuts, on reflection :)

galund:
I won't be pushing the related ticket into test for this PR, so I'm wondering what the best way to get some QA love would be. Make a new ticket and push that through to @andrewchong, or just discuss here?

TheDoubleK:
I don't think it needs a new ticket - tell Andrew what the relevant things to check on preview are once it gets merged I reckon.

app/__init__.py
TheDoubleK: Yeah I think it's peanuts in this day and age.  Doesn't mean we shouldn't think about it though.

tests/helpers.py
```diff
@@ -14,6 +15,12 @@
 
 class BaseApplicationTest(object):
     def setup_method(self, method):
+        # We need to mock the API client in create_app, however we can't use patch the constructor,
+        # as the DataAPIClient instance has already been created; nor can we temporarily replace app.data_api_client
+        # with a mock, because then the shared instance won't have been configured (done in create_app). Instead,
+        # just mock the one function that would make an API call in this case.
+        data_api_client.find_frameworks = mock.Mock()
+        data_api_client.find_frameworks.return_value = BaseApplicationTest._get_fixture_data('frameworks.json')
```
TheDoubleK: Nice neat way to fix up all the tests 👏

TheDoubleK:
I think this looks good to go :+1:

In terms of QA just the functional tests all running should give us a decent level of confidence that nothing is broken, as all relevant content must be being loaded for them to pass.

But would be good to give Andrew a heads up - I guess what we need to check is that service pages for all old frameworks load, plus briefs can still be made and responded to?  That should cover it?

PR #463, opened 2017-03-23
Support for G9 services

samuelhwilliams:
## Summary
Adds compatibility support for G9 services, which have some slightly different question keys.

## Changes
* Load the `display_service` G9 manifest
* Pull 'serviceDescription' from G9 into framework['serviceSummary']; the key has changed for some reason. Maybe it would be better to normalise this part and bring it back to serviceSummary...
* Add a check for 'minimumContractPeriod' key and include the caveat only if present.
* Add a check for 'freeVersionTrialOption', which combines previous keys freeOptional and trialOption.

## Dependency
https://github.com/alphagov/digitalmarketplace-frameworks/pull/410
* Requires version 6.4.10 of the frameworks repo, which adds a G9 display_service manifest.

### DIFF
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -28,7 +28,9 @@ class Service(object):
         self.service_questions = service_questions
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
-        self.serviceSummary = service_data['serviceSummary']
+        self.serviceSummary = service_data['serviceSummary']\
+            if 'serviceSummary' in service_data else \
+            service_data['serviceDescription']
         self.lot = service_data['lot']
         self.frameworkName = service_data['frameworkName']
         # optional attributes directly mapped to service_data values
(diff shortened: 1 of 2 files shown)


app/main/helpers/search_helpers.py
```diff
@@ -218,3 +218,10 @@ def valid_page(page):
         except ValueError:
             return None
     return None
+
+
+def map_framework_slug_to_display_service_framework_slug(slug):
```
galund: This conflicts with the fix I just landed - sorry! Basically, we don't need it, cos it's covered by the map I've added into the config.

One difference is I didn't bother mapping G-Cloud 7 and 8 because we do actually have content for those (even though it wasn't used, and is as far as I could see identical the the G-Cloud 6 content).

samuelhwilliams: Nice, no matter, thanks!

app/main/presenters/service_presenters.py
```diff
@@ -28,7 +28,9 @@ def __init__(self, service_data, service_questions):
         self.service_questions = service_questions
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
-        self.serviceSummary = service_data['serviceSummary']
+        self.serviceSummary = service_data['serviceSummary']\
+            if 'serviceSummary' in service_data else \
+            service_data['serviceDescription']
```
galund: If 'description' is the new one, then maybe we should change the variable name.

PR #464, opened 2017-03-24
Remove search box from /g-cloud

samuelhwilliams:
## Ticket
https://trello.com/c/fEWlJym1/344-remove-search-on-g-cloud-page

## Summary
The lot will be the most important filter for users to select once G-Cloud 9 goes live. This PR removes the search box on the initial /g-cloud page so that users must choose which lot they are searching before further filtering/browsing services.

## Changes
* Adds a generic 'gcloud9' feature flag, as we expect to be deploying a number of changes that should only go active with G-Cloud 9.
* Enable the feature flag for local development and preview.
* Remove the search box from /g-cloud (gated by said feature flag). 

## Old
screen shot 2017-03-27 at 08 39 03

## New
screen shot 2017-03-27 at 08 38 36

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -2,7 +2,6 @@
 from __future__ import unicode_literals
 
 from flask import abort, render_template, request, redirect, current_app, url_for
-import flask_featureflags as feature
 
 from dmutils.formats import dateformat
 from dmapiclient import HTTPError
@@ -30,11 +29,12 @@ from app import search_api_client, data_api_client, content_loader
 
 @main.route('/g-cloud')
 def index_g_cloud():
-    show_search_box = not feature.is_active('GCLOUD9')
-
     # if there are multiple live g-cloud frameworks, assume they all have the same lots
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
     framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+
+    # TODO remove me after G-Cloud 9 goes live
+    show_search_box = not framework_helpers.is_g9_live(all_frameworks)
 
     lot_browse_list_items = list()
     for lot in framework['lots']:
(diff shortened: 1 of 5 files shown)


TheDoubleK:
Enabling this on Preview/Staging will break the functional tests, that have a whole bunch of scenarios around searching for services.

It would be good to have a corresponding PR up for the functional tests with changes that will allow them to pass once this gets merged.

app/main/views/g_cloud.py
```diff
@@ -29,7 +30,10 @@
 
 @main.route('/g-cloud')
 def index_g_cloud():
-    return render_template('index-g-cloud.html')
+    show_search_box = not feature.is_active('GCLOUD9')
+
+    return render_template('index-g-cloud.html',
```
TheDoubleK: All the lots and everything for G-Cloud 9 are wrong in that template.  Rather than building up a huge sequence of conditionals in that file, would it be better to have the feature flag switch to a new `index-g-cloud` template altogether (with a different name, but one that doesn't specifically include a `9` in it), with no search box and the right lots?

And then after G9 goes live we can remove the old template along with the feature flag.

samuelhwilliams: George has fixed the lots in his own work.

TheDoubleK:
I also wonder how the feature flag will play with turning everything off and on when the time comes (i.e. expiring G7, not expiring G8, making G9 live).  Updating the feature flag means a code release as well as flipping things via the API, and ideally we still want everything to work in the meantime.

I guess this is what a feature flag is meant for though, and having things in the frontend conditional on framework states would be pretty ugly. But this does feel like a thing that directly corresponds to framework state, rather than features behind a flag, so we could do that instead and tidy up once all the framework states are settled.  What do you reckon?

galund:
hmm yeah, would be nice if we could toggle configuration without having to do a code release!

Maybe having a custom feature flag checker would make this easier to rip out later - something along the lines of:

```
def check_g_cloud(feature):
    if feature == 'GCLOUD9':
        all_frameworks = data_api_client.find_frameworks().get('frameworks')
        framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
        return framework['slug'] != 'g-cloud-8'

feature_flags.add_handler(check_g_cloud)
```

Issue with doing this is that for any view that was going to load up the frameworks anyway (which is probably most of these now), we'll end up with two API calls, one for the feature flag check, and one for whatever else.

TheDoubleK:
:+1:

PR #465, opened 2017-03-27
Remove hardcoded lots

galund:


app/main/helpers/framework_helpers.py
```diff
@@ -7,3 +7,7 @@ def get_latest_live_framework(all_frameworks, framework_type):
         return latest
     except ValueError:  # max of empty iterable
         return None
+
+
+def get_lots_by_slug(framework_data):
+    return dict((lot['slug'], lot) for lot in framework_data['lots'])
```
samuelhwilliams: Slightly more optimal and clear to use a dict comprehension here, I think:
`{lot['slug']: lot for lot in framework_data['lots']}`

app/main/presenters/search_results.py
```diff
@@ -4,17 +4,26 @@
...
-                        ''.join(service['highlight']['serviceSummary'])
-                    )
+    def _annotate(self):
+        for service in self.search_results:
+            self._replace_lot(service)
+            self._add_highlighting(service)
+
+    def _replace_lot(self, service):
+        # replace lot slug with reference to dict containing all the relevant lot data
+        service['lot'] = self._lots.get(service['lot'])
```
samuelhwilliams: I think this is probably neater inline if there's only one reference to it, which it looks like there is.

galund: This was for symmetry with `_add_highlighting`, i.e. to make it clear we now perform two annotation operations on the dataset

app/main/helpers/framework_helpers.py
galund: thanks, fixed now, forgot about dict comprehensions :)

app/main/presenters/service_presenters.py
```diff
@@ -24,14 +24,15 @@ def _add_as_attribute_if_key_exists(self, key, service_data):
...
-    def __init__(self, service_data, service_questions):
+    def __init__(self, service_data, service_questions, lots_by_slug):
         self.service_questions = service_questions
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
         self.serviceSummary = service_data['serviceSummary']\
             if 'serviceSummary' in service_data else \
             service_data['serviceDescription']
-        self.lot = service_data['lot']
+        # according to our test data, lot field in service_data may not be correct slug i.e. it may be mixed case
```
TheDoubleK: What test data? I think all our production data had their lot slugs migrated to all-lowercase a long time ago, so I don't think this should be necessary.  Would be good to know if this is really a thing still.

The commit message here suggests it shouldn't be and that all mixed-case lot slugs can be done away with: https://github.com/alphagov/digitalmarketplace-frameworks/commit/b29692060e3c7f90bfb37f5c166ca92850b58d8d

galund: The fixture in question is `dm-buyer-frontend/tests/fixtures/g6_service_fixture.json` and if you take out the `.lower()` backup call then the following tests fail:

`test_lot_attribute_is_set`
`test_g5_service_page_url`
`test_g6_service_page_url`

In the last two, it's assertions relating to the breadcrumbs that fail.

galund: @TheDoubleK have fixed up the tests now so we aren't using mixed-case lot slugs.

galund: Yup, already pushed, it never happened, etc :)

app/main/views/g_cloud.py
```diff
@@ -29,7 +28,42 @@
...
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+
+    lot_browse_list_items = list()
+    for lot in framework['lots']:
+        lot_item = {
+            "link": url_for('.search_services', lot=lot['slug']),
+            "title": lot['name']
+        }
+
+        # TODO proper lot body/subtext for G9 - G7/G8 content moved here temporarily from template
```
galund: It's this ticket: https://trello.com/c/qF5KbUwX/377-update-content-on-homepage-for-g-cloud

Although right now there are discussions about dropping that page entirely, and having links direct from the DM home page. Not sure if that will come to anything before go-live though.

app/main/views/g_cloud.py
```diff
@@ -129,21 +162,31 @@ def search_services():
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
     framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
 
+    current_lot_slug = get_lot_from_request(request)
+    current_lot = None
+    lots = framework['lots']
+    for lot in lots:
```
galund: Impossible, at the moment anyway. There's currently an 'all categories' option on the search page, but that's effectively 'no lot selected' rather than several. (And that's going away for G-Cloud 9 anyway.)

tests/fixtures/g5_service_fixture.json
```diff
@@ -13,7 +13,7 @@
     "frameworkSlug": "g-cloud-5",
     "frameworkStatus": "live",
     "status": "published",
-    "lot": "IaaS",
+    "lot": "iaas",
```
galund: woot, thanks 👍

PR #466, opened 2017-03-28
Don't break for non-numeric `page` URL param

TheDoubleK:
We've seen some 500s in production for someone trying `page=test` in the URL.  This should fix that.


samuelhwilliams:
Would've been nice to do a better fix for this but OK, I guess this solves half the problem.

TheDoubleK:
What's the other half?

PR #468, opened 2017-04-04
Fix G-Cloud filters lot slug.

galund:
- should have gone in with 65e24 :-(

I had fixed this in my followup PR, but not in the first one, woops.


TheDoubleK:
Ah, not obvious at all. :+1: :crossed_fingers:

PR #470, opened 2017-04-04
New category filter

galund:

The major layout change here should only make an appearance once G9 is the latest live framework, which we are going to test on Preview shortly.

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -32,11 +32,22 @@ def get_filters_from_request(request):
 def allowed_request_lot_filters(lot_filters):
     """Create a set of (name, value) pairs for all form filters."""
 
-    return set(
-        (f['name'], f['value'])
-        for section in lot_filters
-        for f in section['filters']
-    )
+    filters = set()
+
+    def recursive_search(filters):
+        more_filters = set()
+        for f in filters:
+            more_filters.add((f['name'], f['value']))
+            children = f.get('children')
+            if children:
+                more_filters.update(recursive_search(children))
+        return more_filters
+    # recursive search to account for sub-filters (i.e. sub-categories)
+
+    for section in lot_filters:
+        filters.update(recursive_search(section['filters']))
+
+    return filters
 
 
 def clean_request_args(request_args, lot_filters, lots_by_slug):
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -96,6 +96,9 @@ def show_lots_and_categories_selection(lots, category_filter_group, request):
     Equivalent of set_filter_states but for where we are creating a tree of links i.e. the
     lots/categories widget. Adds links (where necessary) and shows the currently-selected
     lot and/or categories.
+    :param lots: a sequence of lot dicts, which this function will annotate
+    :param category_filter_group: a single filter group loaded from the framework search_filters manifest
+    :param request: current request so that we can figure out what lots and categories are selected
     """
     current_lot_slug = get_lot_from_request(request)
 
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -103,7 +103,6 @@ def show_lots_and_categories_selection(lots, category_filter_group, request):
         lot_selected = (lot['slug'] == current_lot_slug)
         if lot_selected:
             lot['selected'] = True
-            # TODO G9 test - don't want this for G7/8
             categories = category_filter_group['filters'] if category_filter_group else []
             category_was_selected = _show_category_selection(lot, categories, request)
             lot['categories'] = categories
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -1,3 +1,5 @@
+from collections import OrderedDict
+
 from ..helpers.search_helpers import get_filters_from_request
 
 
@@ -16,7 +18,7 @@ def sections_for_lot(lot, builder):
 
 def filters_for_lot(lot, builder):
     sections = sections_for_lot(lot, builder)
-    lot_filters = []
+    lot_filters = OrderedDict()
 
     for section in sections:
         section_filter = {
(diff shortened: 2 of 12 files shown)


app/main/presenters/search_presenters.py
```diff
@@ -43,20 +51,29 @@ def filters_for_question(question):
...
+            'id': '{}-{}'.format(
+                question['id'],
+                option['label'].lower().replace(' ', '-')),
+            'value': option['label'].lower(),
+        }
+        if option.get('options'):
+            presented_filter['children'] = []
+            _recursive_add_option_filters(question, option.get('options', []), presented_filter['children'])
+
+        filters_list.append(presented_filter)
```
risicle: Og course I would have done this as a recursive generator but w/e

app/main/presenters/search_presenters.py
```diff
@@ -72,3 +89,70 @@ def set_filter_states(filter_groups, request):
...
+            lot['link'] = url_for('.search_services', q=get_keywords_from_request(request), lot=lot['slug'])
+
+
+def _show_category_selection(lot, category_filters, request):
+    """
+    Recursive setting of 'selected' state and adding of links to categories
+    as part of building the lots/categories tree.
+    :param category_filters: set of category filters as previously produced by filters_for_question
+    :param request: request object from which to extract active filters
+    :return: True iff a category from the list was selected
```
risicle: iff

galund: I was taught to do this, you are not the first to be confused, but I would find it difficult to stop now :-(

"iff" === "if and only if", i.e. "True if a category from the list was selected, but otherwise False", in other words, extremely pedantic logical ass-covering.

TheDoubleK: I'm fine with `iff`, but only because I did courses on formal logic in the past.

app/main/presenters/search_presenters.py
galund: I think the most annoying thing is that I've ended up with basically the same recursion in about a zillion different places. Replacing them all with a single "flatten tree" generator thing would be a nice thing for the future, but not on the timescales I have right now :(

(Incidentally doing it with generators is a bit nicer once we can use `yield from` which isn't in Python 2.)

app/main/presenters/search_presenters.py
```diff
@@ -72,3 +89,70 @@ def set_filter_states(filter_groups, request):
                 filter['checked'] = (
                     filter['value'] in param_values
                 )
+
+
+def show_lots_and_categories_selection(lots, category_filter_group, request):
+    """
+    Equivalent of set_filter_states but for where we are creating a tree of links i.e. the
+    lots/categories widget. Adds links (where necessary) and shows the currently-selected
+    lot and/or categories.
```
risicle: So `lots` is a seq of lot dicts is it? (rather than a list of slugs)

risicle: Also hold on - its output is that it writes to these `lots`??

risicle: ohhhhkay

risicle: i'm not sure this is the best name for this function though

app/main/presenters/search_presenters.py
risicle: Cool, don't know how many people here are going to grok that

app/main/presenters/search_presenters.py
```diff
@@ -72,3 +89,73 @@ def set_filter_states(filter_groups, request):
...
+        if not lot_selected or category_was_selected:
+            # we need a link back to the lot _without_ a category selected
+            lot['link'] = url_for('.search_services', q=get_keywords_from_request(request), lot=lot['slug'])
+
+
+def _show_category_selection(lot, category_filters, request):
+    """
+    Recursive setting of 'selected' state and adding of links to categories
+    as part of building the lots/categories tree.
+    :param category_filters: set of category filters as previously produced by filters_for_question
```
risicle: Do you mean an actual `set`?

app/main/presenters/search_presenters.py
```diff
@@ -72,3 +89,73 @@ def set_filter_states(filter_groups, request):
...
+
+            category['link'] = url_for('.search_services',
+                                       q=get_keywords_from_request(request),
+                                       lot=lot['slug'],
+                                       **url_args)
+
+    # Remove parent categories (except the selected one), when any subcategory is selected - we
+    # we think this may make navigation easier (and it's similar to what Amazon does)
+    if any_descendant_selected:
+        category_filters[:] = (c for c in category_filters if c['selected'])  # in-place filter(!)
```
risicle: `in-place filter(!)` - The fact that you're having to do this suggests that this function has a weird interface, but I understand you're torn by trying to fit in with existing interfaces.

app/main/presenters/search_summary.py
```diff
@@ -103,17 +103,31 @@ def _group_request_filters(self, request_args, filter_groups):
...
+            """
+            for filter in filters:
+                if filter[key] == value:
+                    return filter
+                children = filter.get('children')
+                if children:
+                    search_children = _get_filter_recursive(key, value, children)
+                    if search_children is not None:
+                        return search_children
+
```
risicle: No actual reason for this to be a nested function (but then again the same is true for `_is_option` above)

app/main/helpers/search_helpers.py
```diff
@@ -32,11 +32,22 @@ def get_filters_from_request(request):
...
+            children = f.get('children')
+            if children:
+                more_filters.update(recursive_search(children))
+        return more_filters
+    # recursive search to account for sub-filters (i.e. sub-categories)
+
+    for section in lot_filters:
+        filters.update(recursive_search(section['filters']))
+
+    return filters
```
risicle:
```
def _flattened(filter_):
    for t in chain(((f['name'], f['value'],),), chain.fromiterable(_flattened(child) for child in f.get("children", ()))):
        yield t
```

Merry xmas.

risicle: oh, but otherwise :+1:

PR #473, opened 2017-04-11
Adjust category filter

galund:
A UI change (re which categories are visible once a selection has been made) and a functionality fix (the filters should combine with the category selection as well as the lot selection). Both affect G9 only.

UI change - before:

screen shot 2017-04-18 at 13 12 59

After:

screen shot 2017-04-18 at 13 13 31

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -99,20 +99,24 @@ def annotate_lots_with_categories_selection(lots, category_filter_group, request
     :param lots: a sequence of lot dicts, which this function will annotate
     :param category_filter_group: a single filter group loaded from the framework search_filters manifest
     :param request: current request so that we can figure out what lots and categories are selected
+    :return: the category filter that was directly selected, if any
     """
     current_lot_slug = get_lot_from_request(request)
+    selected_category = None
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
Looks good to me 👍 

On a small note, it was a bit hard to come in with not much context. Things that might have helped:
- The trello story could be more explicit. It feels like a general conversation around filtering or a design story rather than a explicitly stated piece of technical work.
- A screenshot of the expected before and after behaviour for hiding the categories

I'm sure Sam would have had no problems but it just made it take a bit longer for someone from another team to really understand what was going on and why.

galund:
Thanks David, yes:

 - I thought code review was more about technical things rather than whether the functionality being changed/introduced was right or not, but without context I agree it's quite hard to review either;
 - the story in this case didn't cover either of the things I was changing here. If it had done, it would have been because I had written it in, which I don't feel provides much validation of the change anyway :) But still it might have helped.
 - always happy to add screenshots if it helps.

PS I don't feel like I have much context for the changes I'm making in the system either, given I'm pretty new. Don't suppose that fills you with confidence either, tho ;-) .. cheers again

PR #475, opened 2017-04-12
Fixes validation masthead

MuriloDalRi:
Changes minimum amount of errors necessary to trigger validation masthead to 1 on three different templates.

### DIFF
--- a/app/templates/auth/create-user.html
+++ b/app/templates/auth/create-user.html
@@ -4,7 +4,7 @@
 
 {% block main_content %}
 
-{% if form.errors|length > 1 %}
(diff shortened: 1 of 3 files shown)


risicle:
Again, I think this is Not A Bug, but was a deliberate decision. See my comment in https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/693 and story https://www.pivotaltracker.com/story/show/143115633

TheDoubleK:
Yep this is a deliberate design decision from way back.  I think we should close these two PRs.

TheDoubleK:
Closing as this was designed behaviour.

PR #476, opened 2017-04-18
Fix categories with commas

galund:
Bug-fix - search/filter where value contains a comma.

https://trello.com/c/Fi4N1ypL/382-categories-with-a-comma-in-name-broken

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -121,6 +121,20 @@ def replace_g5_search_dots(keywords_query):
     )
 
 
+def get_filter_value_from_question_option(option):
+    """
+    Processes `option` into a search term value suitable for the Search API.
+
+    The Search API does not allow commas in filter values, because they are used
+    to separate the terms in an 'OR' search - see `group_request_filters`. They
+    are removed from the field values by the indexer by the Search API itself,
+    see `dm-search-api:app.main.services.conversions.strip_and_lowercase`.
+
+    :return: normalized string value
+    """
+    return (option.get('value') or option.get('label', '')).lower().replace(',', '')
+
+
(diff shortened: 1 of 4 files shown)


app/main/helpers/search_helpers.py
```diff
@@ -121,6 +121,20 @@ def replace_g5_search_dots(keywords_query):
...
+    Processes `option` into a search term value suitable for the Search API.
+
+    The Search API does not allow commas in filter values, because they are used
+    to separate the terms in an 'OR' search - see `group_request_filters`. They
+    are removed from the field values by the indexer by the Search API itself,
+    see `dm-search-api:app.main.services.conversions.strip_and_lowercase`.
+
+    :return: normalized string value
+    """
+    return (option.get('value') or option.get('label', '')).lower().replace(',', '')
```
TheDoubleK: Is it worth using the exact same processing here as in search API? i.e. `re.sub(r'\W+', '', value).lower()`?  Might it be worth extracting into a utils method so we guarantee identical processing in both places?

TheDoubleK:
Left a comment, see what you think... fine as-is if you don't think it's worth extracting a utils method for this :+1:

app/main/helpers/search_helpers.py
galund: Yeah, I do kinda agree, this is a slight mess, because they don't quite _need_ to perform the same operation - that regex removes all the spaces, which we don't actually require here. The regex _does already_ get run over the values once they are posted for a search into the Search API, but before we get to that stage we have the delimiter clash.

That said, stripping the spaces before we sent the query wouldn't actually do any harm. Ho hum.

galund:
Have decided to come back to the utils question another day... feeling virtuous enough that I've referenced the similar function :)

PR #481, opened 2017-04-27
Don't break if no content for service framework

TheDoubleK:
Fixes this bug: https://www.pivotaltracker.com/story/show/144457253

We've seen some 500s in production for someone trying to look at DOS services "service pages".  We should 404 requests for service IDs on frameworks that don't have a display_service manifest.

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from dmcontent.errors import ContentNotFoundError
 from flask import abort, render_template, request, redirect, current_app, url_for
 
 from dmutils.formats import dateformat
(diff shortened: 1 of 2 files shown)


galund:
Have we definitely checked referer for these log entries? I wonder if we could redirect to some other page instead, perhaps there's _something_ we can show?

tests/main/views/test_services.py
```diff
@@ -305,6 +305,20 @@ def test_pre_live_framework_causes_404(self):
...
+        self.service = self._get_g6_service_fixture_data()
+        self.service['services']['frameworkSlug'] = 'digital-outcomes-and-specialists'
+        data_api_client.get_service.return_value = self.service
+        data_api_client.get_framework.return_value = self._get_framework_fixture_data(
+            'digital-outcomes-and-specialists'
+        )
+
+        service_id = self.service['services']['id']
+
+        res = self.client.get('/g-cloud/services/{}'.format(service_id))
```
risicle: I'm a little confused about what's going on here - you're getting g-cloud (6) service data, setting the `frameworkSlug` to `digital-outcomes-and-specialists` and then accessing the url `'/g-cloud/services/{}'`?

TheDoubleK: I'll add a comment to the test - it is right, but does look a bit funky.

app/main/views/g_cloud.py
```diff
@@ -116,13 +117,17 @@ def get_service_by_id(service_id):
...
-        )
+        try:
+            service_view_data = Service(
+                service_data,
+                content_loader.get_builder(framework_slug, 'display_service').filter(
+                    service_data
+                ),
+                framework_helpers.get_lots_by_slug(framework)
+            )
+        except ContentNotFoundError as e:
```
TheDoubleK: Yep that feels like a better check doesn't it, Rob said a similar thing yesterday too.  If there's no manifest for a g-cloud framework then something is really wrong.

TheDoubleK:
The 'referer' is empty for these requests - which I think means it's someone entering the URL directly?  Quite possibly one of us to be honest.

TheDoubleK:
Updated to check for `g-cloud` meta-framework rather than swallow all ContentNotFoundErrors.

PR #485, opened 2017-05-02
Update content on DM homepage for G-Cloud 9

samuelhwilliams:
## Summary
Updates the content on the homepage when G-Cloud 9 is live to use new terminology (eg cloud hosting, software and support). Removes some redundancy from the template listing content twice.

It would be nice to pull this content from the frameworks repo, but crown-hosting is a separate framework altogether (Crown Hosting Data Centres Framework). Putting it in g-cloud would be wrong, and creating an entire new framework directory just for these messages feels just as bad. I'm not adverse to putting the specific g-cloud text in the frameworks repo, but think it's easier left as-is until we tackle this more properly across the board.

## Ticket
https://trello.com/c/qF5KbUwX/377-update-content-on-homepage-for-g-cloud-9

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -22,9 +22,9 @@
 {% endif %}
   <div class="column-two-thirds">
     <h2>Find technology or people for digital projects in the public sector</h2>
+      {% set items = [] %}
       {% if dos_slug %}
-        {% with
-          items = [
+        {% set items = [
             {
               "link": url_for("dos.info_page_for_starting_a_brief", framework_slug=dos_slug, lot_slug='digital-specialists'),
               "title": "Find an individual specialist",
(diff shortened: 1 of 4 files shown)


app/templates/index.html
```diff
@@ -47,13 +47,13 @@
...
-              "title": "Find cloud technology and support",
-              "body": "eg web hosting or IT health checks",
+              "title": "Find cloud hosting, software and support" if show_new_g9_content else "Find cloud technology and support",
+              "body": "eg content delivery networks or accounting software" if show_new_g9_content else "eg web hosting or IT health checks",
             },
             {
               "link": "/crown-hosting",
-              "title": "Buy physical datacentre space",
-              "body": "eg for services that can’t be migrated to the cloud",
+              "title": "Buy physical datacentre space for legacy systems" if show_new_g9_content else "Buy physical datacentre space",
```
TheDoubleK: We (well, @superroz) only just removed the "legacy systems" text, I think at the behest of CCS. Are we 100% sure we're putting it straight back in again?

app/templates/index.html
```diff
@@ -64,13 +64,13 @@
           items = [
```
TheDoubleK: This might be a good opportunity to extract the `g-cloud` and `crown-hosting` items into a single template variable to avoid duplication?  e.g. then have something like `if dos_slug ... items = dos_items + base_items else items = base_items` and only have each piece of text once.

TheDoubleK:
This was the commit "19 days ago" removing the legacy systems stuff: https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/9b6a113b4ea6376676624c17d80052f7bbdb400e

PR #494, opened 2017-05-05
Translate internal values to human-readable labels

samuelhwilliams:
## Summary
G-Cloud 9 uses internal representations of answers to store data about services. This functionality had been implemented but not used before, and the buyer-fe doesn't know how to change these internal values back to answer labels suitable for presenting to users.

## Old
screen shot 2017-05-05 at 16 23 33

## New
screen shot 2017-05-05 at 16 23 03

## Ticket
https://trello.com/c/SmYXzVyu/374-buyer-service-summary-page-displays-question-values-not-labels


app/main/presenters/service_presenters.py
```diff
@@ -33,15 +34,35 @@ def __init__(self, service_data, service_questions, lots_by_slug):
...
         ]:
-            self._add_as_attribute_if_key_exists(key, service_data)
-        self.attributes = self._get_service_attributes(service_data)
-        self.meta = self._get_service_meta(service_data)
+            self._add_as_attribute_if_key_exists(key, translated_service_data)
+
+        self.attributes = self._get_service_attributes(translated_service_data)
+        self.meta = self._get_service_meta(translated_service_data)
+
+    def _convert_values_to_labels(self, service_data):
```
TheDoubleK: Doesn't the `.summary()` in the content loader do this for us?

samuelhwilliams: Yes. That summary is used by this function to update the raw JSON in service_data, which the Service presenter uses, with their equivalent labels.

samuelhwilliams: I take it back. You were 100% correct.

app/main/views/g_cloud.py
```diff
@@ -121,7 +121,7 @@ def get_service_by_id(service_id):
             service_data,
             content_loader.get_builder(framework_slug, 'display_service').filter(
                 service_data
-            ),
+            ).summary(service_data),
```
galund: I wonder if it would be preferable to have this call inside the Service [presenter] constructor?

galund:
Hold off landing this until I've shown you my branch :)

galund:
My PR would essentially replace this, I _think_. https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/503

 - the filtering of blank questions adjustment moves into the template
 - I had the summary() call inside the presenter, but eh.

PR #497, opened 2017-05-08
Only show the parent of the selected sub-category.

samuelhwilliams:
## Summary
Currently, when a sub-category in a hierarchy is selected, all parents with that sub-category will be displayed. This change will make it so that only the direct parent of the sub-category chosen by the user will be visible.

## Old
screen shot 2017-05-08 at 14 28 11

## New
screen shot 2017-05-08 at 14 27 15

## Ticket
https://trello.com/c/NywNOnaD/410-category-filter-make-sub-category-selection-only-show-one-parent-category

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -122,12 +122,13 @@ def annotate_lots_with_categories_selection(lots, category_filter_group, request
     return selected_category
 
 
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
Yes, it won't, but that was an assumption I made when complying with the principle of coding 'what works', not "what will work forever'. :)

PR #499, opened 2017-05-09
Summary manifest for initial G9 support filters.

galund:
Summary manifest for new support filters (see frameworks PR <https://github.com/alphagov/digitalmarketplace-frameworks/pull/430>)

https://trello.com/c/3XUKd0tX/345-add-filters-to-search-results-page

### DIFF
--- a/app/main/helpers/search_summary_manifest.yml
+++ b/app/main/helpers/search_summary_manifest.yml
@@ -85,6 +85,28 @@
   filtersPreposition:
   filterRules:
   conjunction: and
+
+
+-
+  id: Supplier type
+  labelPreposition: of type
+  conjunction: or
+-
+  id: Staff security clearance
+  labelPreposition:
+  filtersPreposition: with
+  conjunction: or
+  filterRules:
+  -
+    preposition: staff security clearance that
+    id: Conforms to BS7858:2012
+  -
+    preposition: staff having
+    id: Other security clearance
+-
+  id: Pricing options
+  labelPreposition: with a
+
 -
   id: User support
   labelPreposition: with
--- a/app/main/helpers/search_summary_manifest.yml
+++ b/app/main/helpers/search_summary_manifest.yml
@@ -8,16 +8,13 @@
   filterRules:
   conjunction: and
 -
-  id: Cloud deployment model
-  labelPreposition: deployable to a
-  conjunction: and
--
   id: Pricing
   labelPreposition:
   label:
   filtersPreposition: with a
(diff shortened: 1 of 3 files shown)


app/main/helpers/search_summary_manifest.yml
```diff
@@ -85,6 +85,28 @@
   filtersPreposition:
   filterRules:
   conjunction: and
+
+
+-
+  id: Supplier type
+  labelPreposition: of type
```
samuelhwilliams: This preposition is really weird for the associated filter labels, especially when contrasted against the staff security clearance ones: '272 results found in Cloud hosting of type Reseller providing extra features and support and with staff security clearance that Conforms to BS7858:2012'.

Also the capitalisation on the filter labels is weird; if we're trying to construct proper sentences, we should probably have the filter labels be lower-case and then capitalise appropriately in the templates where they stand alone (e.g. in the filter boxes).

galund: I think if you want to suggest a better preposition thing then that's cool. Basically the more filters there are, the closer to impossible it becomes to have the thing make a sensible sentence. And I don't think it will be possible to add punctuation that consistently works with this mechanism, although maybe it would be and maybe that could help.

After this sprint we need to look at whether this summary thing is right (placeholder ticket @ <https://trello.com/c/SpHlwgBj/440-search-summary-refactoring>), short run we need _something_ so that every filter shows up in the summary.

karlchillmaid: I'm going back through the filter labels I've done to correct capitalisation.

We'll try to get some of Roz's time tomorrow to discuss best solution for prepositions etc for MVP.

samuelhwilliams: Yeah Karl popped over a bit ago and said the same thing. I don't have a better suggestion off the top of my head, and I understand the pain of trying to come up with one.

samuelhwilliams:
Might be worth landing the capitalisation for filters here as well, given it's going to pull in the new lowercase filter labels. There are just two places the filter labels get pulled in and you can just drop .capitalise() on the end of them IIRC.

galund:
@samuelhwilliams have been looking at the capitalize thing, it isn't completely obvious to me where to stick that in. The filter groups are used for both the filter widget and for the summary generation, so somehow we'd have to split off yet another string (i.e. filter_label_for_filter vs filter_label_for_summary). Unless I'm missing something obvious, hopefully :)

PR #500, opened 2017-05-09
Only show parent categories when no sub-category is selected.

galund:
Per 'version 4' mockup in ticket - the A/C for that (which I just wrote are:

"We now want to see sibling categories of the currently-selected category, but only at the bottom level (i.e. sub-categories in the case of the SaaS lot, but categories in the other two lots)."

Before:
screen shot 2017-05-09 at 17 34 07

After:
screen shot 2017-05-09 at 17 34 00

https://trello.com/c/Fyi0uec3/415-clicking-back-on-categories-navigation

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -177,9 +177,10 @@ def _annotate_categories_with_selection(lot, category_filters, request, parent_c
                                        lot=lot['slug'],
                                        **url_args)
 
-    # Remove categories (except the selected one), when any is selected - we
-    # think this may make navigation easier (and it's similar to what Amazon does)
+    # When there's a selection, remove parent categories (i.e. preserve the selection, plus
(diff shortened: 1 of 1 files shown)


idavidmcdonald:
The code looks legit and matches the screenshots you've provided so from that point of view I'm happy. I'm a bit confused about the in line comment though and the difference between categories and subcategories etc. I'm not sure if that is just my lack of context or maybe the comment could be improved (maybe thinking about using the word `siblings`).

Happy to discuss tomorrow in person if it's important to make sure we are both on the same page but from a technical level I'm happy.

I've leave the second review to give the official thumbs.

galund:
I'm completely sure that comment could be improved... I guess the comment just says what the code says, which is less helpful, and yes getting the word 'siblings' in there is maybe more clear. hmmmm :) ...

PR #503, opened 2017-05-10
Questions with dependencies service page

galund:
A significant refactoring to remove obsolete code from the buyer app.

Effective change is really just to apply bold formatting to questions that have dependencies.

 - depends on https://github.com/alphagov/digitalmarketplace-content-loader/pull/41 landing first

https://trello.com/c/9mmLtRCC/352-questions-with-dependencies

### DIFF
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -24,8 +24,8 @@ class Service(object):
         if key[0] in service_data:
             setattr(self, key[1], service_data[key[0]])
 
-    def __init__(self, service_data, service_questions, lots_by_slug):
-        self.service_questions = service_questions
+    def __init__(self, service_data, builder, lots_by_slug):
+        self.summary_manifest = builder.summary(service_data)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
         self.serviceSummary = service_data['serviceSummary']\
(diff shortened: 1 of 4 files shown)


app/main/presenters/service_presenters.py
```diff
@@ -24,8 +24,8 @@ def _add_as_attribute_if_key_exists(self, key, service_data):
         if key[0] in service_data:
             setattr(self, key[1], service_data[key[0]])
 
-    def __init__(self, service_data, service_questions, lots_by_slug):
-        self.service_questions = service_questions
+    def __init__(self, service_data, builder, lots_by_slug):
+        self.summary_manifest = builder.summary(service_data)
```
samuelhwilliams: Can we call this manifest rather than builder? Builder is the word/name we're trying to move away from in the content-loader IIRC.

samuelhwilliams:
Good work removing those unnecessary layers.

Looks good, but haven't run locally and won't be able to until I'm back in. Might be worth having someone else look over and test.

app/main/presenters/service_presenters.py
galund: @TheDoubleK and I don't really like the word 'manifest' either, as 'manifest' is really the _file_, that the (whatever) class uses to assemble the questions together.

Given the name of the class is actually manifest, then probably you're right and we should stick to that, but I'm sorely tempted to talk to @allait about renaming it, because I think calling it 'manifest' is actually really confusing. Really, it's a `Form` - I think :-)

samuelhwilliams: I'm fine with renaming it, but yeah, given the object is a ContentManifest, that's what my recommendation is based on. :)

app/templates/_service_attributes.html
```diff
@@ -1,17 +1,21 @@
...
+
+    {% call(question) summary.list_table(
+      section.questions,
+      caption=section.name,
+      field_headings_visible=False
+    ) %}
+      {% if not question.is_empty %}
+        {% call summary.row() %}
+          {{ summary.field_name(question.label, bold=question.followup) }}
+          {{ summary[question.type](question.value, question.assurance, bold=question.followup) }}
```
samuelhwilliams: So there's a bit of extra logic required here: only the first question in a dependency chain should be bold; any subsequent followups within the same chain should *not* be bold.

galund:
I'll move the refactoring into a separate PR...

PR #505, opened 2017-05-15
Remove hardcoded G-Cloud descriptions

samuelhwilliams:
## Summary
Lots of templates/etc had hardcoded references to 'Cloud technology and support', which was the standard description for G-Cloud 8 and earlier. We are now changing that to 'Cloud hosting, software and support' and are aware this may change again in the future. Therefore, we want to remove hardcoded references to this and have the sentence fragment as a framework-mutable string.

## Implementation Details
* New descriptions.yml file in messages for each framework, containing a 'description' key which stores the framework's description fragment.
* New get_framework_description helper to retrieve this fragment, given a framework slug.
* Update templates to use this per-framework value.

## Ticket
https://trello.com/c/pYYRBhNu/336-content-on-g-cloud-page

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -52,7 +52,7 @@
       {% set items = items + [
         {
           "link": "/g-cloud",
-          "title": "Find cloud hosting, software and support" if show_new_g9_content else "Find cloud technology and support",
+          "title": "Find " + gcloud_framework_description,
           "body": "eg content delivery networks or accounting software" if show_new_g9_content else "eg web hosting or IT health checks",
         },
         {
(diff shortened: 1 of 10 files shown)


app/templates/index.html
```diff
@@ -52,7 +52,7 @@
       {% set items = items + [
         {
           "link": "/g-cloud",
-          "title": "Find cloud hosting, software and support" if show_new_g9_content else "Find cloud technology and support",
+          "title": "Find " + gcloud_framework_description,
```
galund: Everywhere else I've worked for a long time, this would be a no-no because translations. I suppose in this case it's fine, but having it stored as a separate string would have been fine too :)

PR #509, opened 2017-05-17
Remove hanging bullets on service page

samuelhwilliams:
## Summary
Jeremy doesn't like the hanging bullets on the service page (nor do I), and we've been moving away from them generally, so let's bring them _in line_. (Get it? Hahahaha...)

## Old
screen shot 2017-05-17 at 08 35 58

## New
screen shot 2017-05-17 at 08 35 38


TheDoubleK:
:+1: after a rebase.

PR #510, opened 2017-05-17
Free trial link in meta section of service page

samuelhwilliams:
## Summary
In the Meta section of a service page we currently show whether or not the service offers a free trial. In G-Cloud 9 we want to turn this into a hyperlink if the supplier has provided a link to the free trial. We will also show the hyperlink provided as part of the summary table. Currently we will not turn this text into a hyperlink in the summary table, though we may do so in the future.

## Ticket
https://trello.com/c/Sxd04wcI/416-free-trial-text-becomes-a-link

## Old (from G8)
screen shot 2017-05-17 at 16 10 34

## New (from G9)
screen shot 2017-05-17 at 16 11 26

### DIFF
--- a/app/templates/_service_meta.html
+++ b/app/templates/_service_meta.html
@@ -3,7 +3,18 @@
   <p class="price">{{ service.meta.price }}</p>
   <ul class="price-caveats">
   {% for caveat in service.meta.priceCaveats %}
-    <li>{{ caveat }}</li>
+    <li>
+      {% if caveat['link'] %}
+        {% with text = caveat['text'],
+                link = caveat['link'],
+                target = "_blank"
+        %}
+          {% include "toolkit/external-link.html" %}
+        {% endwith %}
+      {% else %}
+        {{ caveat['text'] }}
+      {% endif %}
+    </li>
   {% endfor %}
   </ul>
   <h2 class="visuallyhidden">Service documents</h2>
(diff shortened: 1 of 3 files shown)


TheDoubleK:
I thought we had a style for "rel=external" that adds a little arrow to indicate "leaving this site", but it seems to have gone from everywhere (e.g. the links to CCS framework pages, the link to supplier's own site shown on the supplier dashboard).  It feels like these might be a good thing if we're going to show arbitrary links on service pages.

The style seems to have gone from our toolkit though.  I'm going to have a quick ask round if that was a deliberate decision or something that got lost by mistake...  

Otherwise code looks good.

TheDoubleK:
Fix the tests.  Then :+1:

TheDoubleK:
For the record here, those little arrow things went as part of wider GOV.UK style changes - there is some discussion about it here: https://github.com/alphagov/govuk_frontend_toolkit/pull/293

app/templates/_service_meta.html
```diff
@@ -3,7 +3,11 @@ <h2 class="visuallyhidden">Pricing</h2>
   <p class="price">{{ service.meta.price }}</p>
   <ul class="price-caveats">
   {% for caveat in service.meta.priceCaveats %}
-    <li>{{ caveat }}</li>
+    <li>
```
TheDoubleK: There could be an argument for using the toolkit external link style rather than baking the HTML ourselves here...
```
{%
  with
  text = caveat['text'],
  link = caveat['link'],
  target = "_blank"
%}
  {% include "toolkit/external-link.html" %}
{% endwith %}
```

TheDoubleK: But I think that style needs an update to include the `nofollow`.

PR #513, opened 2017-05-18
Create category filter for the G-Cloud search

pcraig3:
This pull request creates the category filter by following [this design](https://projects.invisionapp.com/share/W3BKI4M8X#/screens/232623076) from @JeremyHHY.  

The basic idea was to try and generalise the bits we need to build a the filter list so that they're not tied to a particular data structure. Thought @galund's initial idea of a recursive include was a pretty neat trick, so I've continued with that. Makes sense since we're rendering a list of lists of (potentially) unknown depth.

I've built it into this app in this pull request, but it will be easy to relocate this in the frontend toolkit if we think it's a pattern worth moving (I think it probably is).

There's lots more specifics in the individual commits, but @carolinegreen, I'm curious to know what you think of:

- how to highlight which filter link is the currently selected one? [I've gone with `aria-current`](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/881acb54ee719e482b0338021d4dcfe101c2b7d8), but there may be a better way. 
- whether we're happy to use a psuedo-element for the "back" chevron, and, if so, [whether we need the hidden "Back" text](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/4a03e0054838c6754b572b054fd593c9a7c0422b) (possibly it's too annoying).

Edit: removed "Back" text after reckoning that the hierarchy is made clear by the nested list structure.

[Trello card is here](https://trello.com/c/Fyi0uec3/415-clicking-back-on-categories-navigation).

***

Screenshot -> nothing selected
screen shot 2017-05-18 at 14 07 14

Screenshot -> category with children selected
screen shot 2017-05-18 at 14 40 34

Screenshot ->child category selected
screen shot 2017-05-18 at 14 08 40

Gif 
search


pcraig3:
[Fix up the categories picker for the catalogues search page](https://trello.com/c/XBjX4PEB/127-fix-up-the-categories-picker-for-the-catalogues-search-page)

app/templates/search/_categories.html
```diff
@@ -1,14 +1,47 @@
...
-      {% if category.selected %}</strong>{% endif %}{% if category.link %}</a>{% endif %}
-      {% if category.children and category.selected %}
-        {% with categories=category.children %}
-          {% include 'search/_categories.html' %}
-        {% endwith %}
+{%- macro top_level_link(text, url) -%}
+  <a href="{{ url }}" class="lot-filters__top-level-link"><span class="visually-hidden">Back to </span>{{ text }}</a>
+{%- endmacro -%}
+
+{%- macro list_item__unselected(text, url) -%}
```
galund: So I don't particularly like jinja macros - where they're the only solution to a problem that avoids a lot of repetition, then they're fine. But in this case it's just for two nested elements, and the macro's only used the once, so I'd prefer the markup just to appear inline?

app/templates/search/_categories.html
```diff
@@ -1,14 +1,47 @@
...
-    <li>
-      {% if category.link %}<a href="{{ category.link }}">{% endif %}{% if category.selected %}<strong>{% endif %}
-        {{ category.label }}
-      {% if category.selected %}</strong>{% endif %}{% if category.link %}</a>{% endif %}
-      {% if category.children and category.selected %}
-        {% with categories=category.children %}
-          {% include 'search/_categories.html' %}
-        {% endwith %}
+{%- macro top_level_link(text, url) -%}
+  <a href="{{ url }}" class="lot-filters__top-level-link"><span class="visually-hidden">Back to </span>{{ text }}</a>
```
galund: The user didn't necessarily get to the current page by going clicking through all of the levels, so I don't quite like 'back to'. Right now it's obviously the most likely way they got there, but I think in future we will want to have clever ways of encouraging people to search across specific categories (e.g. a suggested search based on keywords they typed). So if we do have to have some extra content for screen readers etc (I'm not sure I agree that this is necessary for accessibility, though), then some other text? "Up a level to" might actually explain the structure better.

app/templates/search/_categories.html
pcraig3: So you're saying just leave everything in the `category_list`?

I've pulled them out mostly so that the `selected` and `unselected` list items can handle their own presentation because the `category_list` is already a pretty chunky macro. 
My worry is that putting everything into the `category_list` will make it even harder to understand.

app/templates/search/_categories.html
galund: Just to document our discussion, I suggested it might actually be clear enough from the context that a given category (or lot) name is more broad than a currently-selected sub-category. So the impact on the user's search results is clear enough in a screen-reading context, even without any introductory text like this at all .

pcraig3: Removing the "Back to" text for now as Voiceover is pretty good about describing how the list is structured. When @carolinegreen is back, we can see what she thinks.

PR #514, opened 2017-05-18
Show withdrawn briefs on buyer dash

benvand:
### Context
When a brief is withdrawn. Buyers may need to republish the brief. 

### User need
As a buyer, I need to create a copy of my withdrawn brief so that I can republish with limited time and effort.

### Acceptance criteria
Given that a brief has been withdrawn, when I go to my accounts, Then I should see the opportunity and I should be able to create a copy

screen shot 2017-05-18 at 16 59 44

### DIFF
--- a/app/buyers/helpers/buyers_helpers.py
+++ b/app/buyers/helpers/buyers_helpers.py
@@ -17,12 +17,12 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, allowed_sta
     return framework, lot
 
 
-def is_brief_correct(brief, framework_slug, lot_slug, current_user_id):
+def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False):
     return (
         brief['frameworkSlug'] == framework_slug
         and brief['lotSlug'] == lot_slug
         and is_brief_associated_with_user(brief, current_user_id)
-        and not brief_is_withdrawn(brief)
+        and (True if allow_withdrawn else not brief_is_withdrawn(brief))
     )
 
 
--- a/app/templates/buyers/dashboard.html
+++ b/app/templates/buyers/dashboard.html
@@ -38,7 +38,7 @@
                 {% endfor %}
             {% endif %}
         {% endwith %}
-        
+
         <div class="column-two-thirds">
             {% with
                 context = current_user.email_address,
@@ -66,9 +66,9 @@
 
     {% call summary.row() %}
         {{ summary.service_link(item.title, url_for(".view_brief_overview", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
-        
+
         {{ summary.text(item.createdAt|dateformat) }}
-        
+
         {% if item.unanswered_required > 0 and item.unanswered_optional > 0 %}
             {{ summary.text("{} required<br>{} optional".format(item.unanswered_required, item.unanswered_optional)|safe) }}
         {% elif item.unanswered_required > 0 %}
@@ -103,7 +103,7 @@
         {{ summary.button(text="Make a copy", action=url_for(".copy_brief", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
     {% endcall %}
 {% endcall %}
-        
+
 {{ summary.heading("Closed requirements", id="closed_requirements") }}
 {% call(item) summary.list_table(
     closed_briefs,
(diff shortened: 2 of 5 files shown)


tests/buyers/helpers/test_buyers_helpers.py
```diff
@@ -101,6 +101,28 @@ def test_is_brief_correct(self):
...
+            'digital-outcomes-and-specialists',
+            'digital-specialists',
+            123,
+            allow_withdrawn=True
+        ) is True
+
+    @pytest.mark.parametrize(
+        ['status', 'allow_withdrawn', 'result'],
+        [
+            ('withdrawn', True, True),
```
idavidmcdonald: This first test scenario is covered by the one you have just written above.
The second test scenario is the same logically but is testing the default argument.
Would be nice to have all the tests regarding status of the briefs be done in the same way (i.e. moving the above two tests into the parametrized version).

tests/buyers/views/test_buyers.py
```diff
@@ -22,38 +22,115 @@
...
 
+@pytest.fixture()
+def find_briefs_mock():
+    base_brief_values = {
+        "createdAt": "2016-02-01T00:00:00.000000Z",
+        "frameworkSlug": "digital-outcomes-and-specialists",
+        "lot": "digital-specialists"
+    }
+
+    find_briefs_response = {
```
idavidmcdonald: Good, sensible idea to have one brief of every status which was lacking in existing tests.

tests/buyers/views/test_buyers.py
```diff
@@ -22,38 +22,115 @@
...
 
-            res = self.client.get("/buyers")
-            document = html.fromstring(res.get_data(as_text=True))
+    def setup_method(self, method):
+        super(TestBuyerDashboard, self).setup_method(method)
+        self.login_as_buyer()
 
-            assert res.status_code == 200
+    def test_draft_briefs_section(self, data_api_client, find_briefs_mock):
+        data_api_client.find_briefs.return_value = find_briefs_mock
```
idavidmcdonald: This test class is nice.
The first 3 lines on every test are being repeated but up to you if you want to try and move that into the setup or not

tests/buyers/views/test_buyers.py
```diff
@@ -315,7 +392,7 @@ def test_copy_brief_and_redirect_to_copied_brief_edit_title_page(self):
     def test_copy_brief_for_expired_framework_redirects_to_edit_page_for_new_framework(self):
         self.data_api_client.get_brief.return_value = api_stubs.brief()  # dos1 brief
 
-        new_brief = self.brief  # dos2 brief
+        new_brief = self.brief.copy()  # dos2 brief
```
idavidmcdonald: Do we need to do this `copy`?

app/templates/buyers/dashboard.html
```diff
@@ -119,9 +119,15 @@
...
     {% call summary.row() %}
-        {{ summary.service_link(item.title, url_for(".view_brief_overview", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
-        {{ summary.text(item.applicationsClosedAt|dateformat) }}
-        {{ summary.link("View responses", url_for(".view_brief_responses", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
+        {% if item.status == "closed" %}
+          {{ summary.service_link(item.title, url_for(".view_brief_overview", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
+          {{ summary.text(item.applicationsClosedAt|dateformat) }}
+          {{ summary.link("View responses", url_for(".view_brief_responses", framework_slug=item.frameworkSlug, lot_slug=item.lot, brief_id=item.id)) }}
+        {% elif item.status == "withdrawn" %}
+          {{ summary.service_link(item.title, url_for("main.get_brief_by_id", framework_framework=item.frameworkSlug, brief_id=item.id)) }}
```
idavidmcdonald: And maybe think about if we test this behaviour explicitly too

tests/buyers/views/test_buyers.py
benvand: So the functionality will mean we're using the old brief as the new brief here. Which isn't really the point of the copy function.
Putting the copy here in the test indicates what we're trying to achieve slightly better IMO.

Perhaps if we called self.brief something like self.brief_data that would be more indicative of what's going on?

idavidmcdonald: Ah I see what you mean.

Don't mind. Whatever you feel is best

benvand: Yeah i've stuck copy on the other instances.
Don't think I explained very well what I meant above. 
The `self.brief` object is mutable so without copy we end up mutating one object that is both our `brief` and our `new_brief`
```
self.brief = api_stubs.brief()
new_brief = self.brief
new_brief["briefs"]["id"] = 1235
new_brief['briefs']['id'] == self.brief['briefs']['id']
>>> True
```

app/templates/buyers/dashboard.html
benvand: Done

app/buyers/helpers/buyers_helpers.py
```diff
@@ -17,12 +17,12 @@ def get_framework_and_lot(framework_slug, lot_slug, data_api_client, allowed_sta
     return framework, lot
 
 
-def is_brief_correct(brief, framework_slug, lot_slug, current_user_id):
+def is_brief_correct(brief, framework_slug, lot_slug, current_user_id, allow_withdrawn=False):
```
benvand: Ah yeah, was thinking of doing that but thought it would get a bit gnarly with having to string check for `'withdrawn'` status to avoid doing the `allow_withdrawn` check.

tests/buyers/helpers/test_buyers_helpers.py
benvand: I've parameterised both now and got rid of the duplication

idavidmcdonald:
Good work both of you. Looks good to me 👍 

Can't be merged until the other PR has gone in obviously...

idavidmcdonald:
Oh and what is the deal with the Make a copy button being over two lines in your screenshot? We probably don't want this going out if it looks like that screenshot where just the first table does it...

benvand:
Ah yeah that was just cause I zoomed in like 50% to take the screenshot

PR #526, opened 2017-05-22
Update third party support question summary.

galund:
- this removes a duplicative 'and and' construction in the search
   summary when one of the support filters is used.

https://trello.com/c/fVpk5C97/468-bug-search-summary-for-user-support-has-and-and


TheDoubleK:
:+1:

PR #528, opened 2017-05-22
Update brief fixture to be on latest framework and latest keys

idavidmcdonald:
### Singular brief fixture
- We update the brief fixture to be on the latest framework.
- We update the format of `startDate` to be saved in the parseable
date format rather than just a simple string.
- We add keys that usually exist on a brief that were not included,
such as `technicalWeighting` and `requirementsLength`.
- We add keys related to the clarification question time period.
These before had not been added and it was therefore assumed
that the brief was open for clarification questions. To make this
more explicit we give dates for `clarificationQuestionsClosedAt`
and `clarificationQuestionsPublishedBy` that are realistic compared
to the `publishedAt` date.

This means that our fixture now represents a brief it were to
be published 'today' and our tests should have more realistic
coverage.

### Multiple brief fixtures
- Order keys alphabetically for better readability
- Add keys that would normally exist on a brief but were missing from our stubs
- Update `startDate` to use our new dateformat for one of the briefs

On the downside with all of these fixtures, by updating it we may lose coverage for whatever was being tested before...

### DIFF
--- a/tests/fixtures/dos_brief_fixture.json
+++ b/tests/fixtures/dos_brief_fixture.json
@@ -1,8 +1,21 @@
 {
   "briefs": {
     "backgroundInformation": "Testing the new platform.",
+    "budgetRange": "Maximum day rate of up to £700.",
+    "clarificationQuestions": [
+      {"question": "Why?", "answer": "Because", "publishedAt": "2016-12-02T00:00:00.000000Z"}
+    ],
+    "clarificationQuestionsAreClosed": true,
+    "clarificationQuestionsClosedAt": "2016-12-07T11:09:28.054129Z",
+    "clarificationQuestionsPublishedBy": "2017-06-10T23:59:59.000000Z",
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Is it worth updating the multiple briefs fixture at the same time?

tests/fixtures/dos_brief_fixture.json
```diff
@@ -45,8 +66,7 @@
         "role": "buyer"
       }
     ],
-    "clarificationQuestions": [
-      {"question": "Why?", "answer": "Because", "publishedAt": "2016-12-02T00:00:00.000000Z"}
-    ]
+    "workingArrangements": "9-5",
```
TheDoubleK: 🎵  What a way to make a living 🎵

TheDoubleK:
Thanks!  Looks good to me :+1:

PR #529, opened 2017-05-23
Update search summary to match filter name in framework filter manifest.

galund:
Update search summary to match filter group name in framework filter manifest.

 - API access should share same preposition as other filters, so remove the filterRules override.

https://trello.com/c/WfqYkQhr/486-metrics-reporting-filter-change


galund:
So as the frameworks change had already landed, there is now no update to bower.json at all.

PR #530, opened 2017-05-24
Remove some logic gated on G-Cloud 9 being live

samuelhwilliams:
## Summary
Some parts of the app require a check for G-Cloud 9 being live to support two incompatible behaviours. Now that G-Cloud 9 is indeed live, let's remove the checks to support G-Cloud 8 and earlier.

## Ticket
https://trello.com/c/7hq3FqyB/425-clean-up-after-g-cloud-9-goes-live

### DIFF
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -28,9 +28,7 @@ class Service(object):
         self.summary_manifest = manifest.summary(service_data)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
-        self.serviceSummary = service_data['serviceSummary']\
-            if 'serviceSummary' in service_data else \
-            service_data['serviceDescription']
+        self.serviceSummary = service_data.get('serviceSummary', service_data.get('serviceDescription'))
         self.lot = lots_by_slug.get(service_data['lot'])
         self.frameworkName = service_data['frameworkName']
         # optional attributes directly mapped to service_data values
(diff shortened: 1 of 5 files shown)


app/main/presenters/service_presenters.py
```diff
@@ -28,9 +28,7 @@ def __init__(self, service_data, manifest, lots_by_slug):
         self.summary_manifest = manifest.summary(service_data)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
-        self.serviceSummary = service_data['serviceSummary']\
-            if 'serviceSummary' in service_data else \
-            service_data['serviceDescription']
+        self.serviceSummary = service_data.get('serviceSummary', service_data.get('serviceDescription'))
```
TheDoubleK: Why didn't we do that in the first place? :)

app/templates/_service_summary_features_and_benefits.html
```diff
@@ -1,4 +1,4 @@
-<p class="service-summary-lede">{{ service.serviceSummary or service.serviceDescription }}</p> {# G-Cloud 9 uses serviceDescription, not serviceSummary. #}
```
TheDoubleK: I think we need to keep this - this template is also used for previous framework iteration service pages, which still exist now.

app/templates/search/_services_results.html
```diff
@@ -9,7 +9,7 @@ <h2 class="search-result-title">
     </p>
 
     <p class="search-result-excerpt">
-        {{ service.serviceSummary or service.serviceDescription }} {# G-Cloud 9 uses serviceDescription, not serviceSummary. #}
```
TheDoubleK: I know that in theory only G9 services will come back from search, but feels like one that we could leave in as a "just in case"?

samuelhwilliams:
👍

TheDoubleK:
:+1:

PR #532, opened 2017-05-25
Bugfix: Square search bar on iOS

pcraig3:
A bug was reported the other day where iOS devices show a rounded search bar.  [Here's the trello card](https://trello.com/c/svwLalaz/153-form-inputs-on-g-cloud-release-when-looking-on-iphone-don-t-appear-to-be-as-intended). [There are a few other tiny layout fixes in this update](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/350), but the search bar is the main one.

### before

img_6333_720

### after

screen shot 2017-05-25 at 13 17 22

***

Note: [we changed all the visuallyhiddens to visually-hidden in the frontend toolkit](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/347) and I didn't think it was a breaking change, but it looks like it might have been.  This pull request makes the relevant fixes (ie, adds two hyphens) in the buyer app.

### DIFF
--- a/app/assets/scss/_service-page.scss
+++ b/app/assets/scss/_service-page.scss
@@ -1,5 +1,5 @@
 // styles specific to service pages
-@import "shared_placeholders/_visuallyhidden.scss";
+@import "shared_placeholders/_visually-hidden.scss";
 
 %sub-heading {
   @include core-24;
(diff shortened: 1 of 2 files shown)


TheDoubleK:
:+1:

PR #535, opened 2017-05-31
Add autospec to many of our mocks

idavidmcdonald:
For this tech debt story - https://trello.com/c/qL9vEC1r/392-ensure-autospec-on-buyer-frontend-mocks.

I have already gone through all of our applications and bumped the mock library to `2.0.0`. This brings in the lovely new feature where if you misspell a test assertion then it will realise for you. (https://bugs.python.org/issue21238). This means that if we have `assret_called_once_with` or `assert_calld_once_with` in our tests then these will raise errors. This essentially is the low hanging fruit picked off.

Until I discovered this, I had trawled through the buyer frontend tests manually and added autospec to nearly all the mocks. This is the results of my effort. Note, I tried changing the `tests/main/test_login.py` file to add autospec but ran into issues and so have not touched this one or spent the time trying to get to the bottom of it.

I think this is a useful PR but I don't know whether I would now suggest doing the same in all of the other apps given the `mock` bump should prevent the majority of errors. We should still be using autospec with new tests but I don't know if we will be adding much value going through our entire codebase. Happy to hear thoughts.

### DIFF
--- a/tests/buyers/views/test_buyers.py
+++ b/tests/buyers/views/test_buyers.py
@@ -64,7 +64,7 @@ def find_briefs_mock():
     return find_briefs_response
 
 
-@mock.patch('app.buyers.views.buyers.data_api_client')
+@mock.patch('app.buyers.views.buyers.data_api_client', autospec=True)
 class TestBuyerDashboard(BaseApplicationTest):
 
     def setup_method(self, method):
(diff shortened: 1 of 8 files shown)


benvand:
Hey so out of interest how much effort would it be to go through and add `autospec=True` to all the mocks assuming we could accurately track down each instance? Are there problems you envisage other than just locating all the unspec'd mocks and adding the autospec? I agree that this PR on its own does add value so 👍

PR #537, opened 2017-06-01
Show the role for specialist opportunity on the list page

CrystalPea:
### Context/Problem
Currently, Suppliers cannot identify specialists role for different opportunities until they click on the opportunity. The opportunity title does not always indicate the role.

### User need
As a supplier, I need to identify relevant specialist roles on the Digital Marketplace so that I only view opportunities I am interested in

### Acceptance criteria
Users can see the opportunities for the roles they are interested in
  
  
Trello card: https://trello.com/c/bZvmAPTQ/369-show-the-role-for-specialist-opportunity-on-the-list-page

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -89,7 +89,7 @@ def get_brief_by_id(framework_framework, brief_id):
         for index, question in enumerate(brief['clarificationQuestions'])
     ]
 
-    brief_content = content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(brief)
+    brief_content = content_loader.get_manifest(brief['frameworkSlug'], 'display_brief').filter(brief)
 
     return render_template(
         'brief.html',
@@ -118,7 +118,7 @@ def list_opportunities(framework_framework):
 
     briefs = [{
         "data": brief,
-        "content": content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(brief)
+        "content": content_loader.get_manifest(brief['frameworkSlug'], 'display_brief').filter(brief)
     } for brief in api_result["briefs"]]
 
     links = api_result["links"]
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -89,9 +89,7 @@ def get_brief_by_id(framework_framework, brief_id):
         for index, question in enumerate(brief['clarificationQuestions'])
     ]
 
-    brief_content = content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(
-        brief
-    )
+    brief_content = content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(brief)
 
     return render_template(
         'brief.html',
@@ -118,7 +116,11 @@ def list_opportunities(framework_framework):
 
     api_result = form.get_briefs()
 
-    briefs = api_result["briefs"]
+    briefs = [{
+        "data": brief,
+        "content": content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(brief)
+    } for brief in api_result["briefs"]]
+
     links = api_result["links"]
 
     api_prev_link_args = parse_link(links, "prev")
--- a/app/templates/search/_briefs_results.html
+++ b/app/templates/search/_briefs_results.html
@@ -1,44 +1,46 @@
 {% for brief in briefs %}
 <div class="search-result">
     <h2 class="search-result-title">
-        <a href="{{ url_for('.get_brief_by_id', framework_framework=framework.framework, brief_id=brief.id) }}">{{ brief.title }}</a>
+        <a href="{{ url_for('.get_brief_by_id', framework_framework=framework.framework, brief_id=brief.data.id) }}">{{ brief.data.title }}</a>
     </h2>
 
     <ul class="search-result-important-metadata">
         <li class="search-result-metadata-item">
-            {{ brief.organisation }}
+            {{ brief.data.organisation }}
         </li>
         <li class="search-result-metadata-item">
-            {{ brief.location }}
+            {{ brief.data.location }}
         </li>
     </ul>
 
     <ul class="search-result-metadata">
         <li class="search-result-metadata-item">
-            {{ brief.lotName }}
+            {{ brief.data.lotName }}
         </li>
+        {% if 'specialistRole' in brief.data %}
         <li class="search-result-metadata-item">
-            {{ brief.frameworkName }}
+            {{ brief.content.summary(brief.data).get_question("specialistRole").value }}
         </li>
+        {% endif %}
     </ul>
 
     <ul class="search-result-metadata">
-        {% if brief.status == 'closed' %}
+        {% if brief.data.status == 'closed' %}
             <li class="search-result-metadata-item">
                 Closed
             </li>
         {% else %}
             <li class="search-result-metadata-item">
-                Published: {{ brief.publishedAt|dateformat }}
+                Published: {{ brief.data.publishedAt|dateformat }}
             </li>
             <li class="search-result-metadata-item">
-                Closing: {{ brief.applicationsClosedAt|dateformat }}
+                Closing: {{ brief.data.applicationsClosedAt|dateformat }}
             </li>
         {% endif %}
     </ul>
 
     <p class="search-result-excerpt">
-        {{ brief.summary }}
+        {{ brief.data.summary }}
     </p>
 </div>
 
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -864,8 +864,9 @@ class TestCatalogueOfBriefsPage(BaseApplicationTest):
         ss_elem = document.xpath("//p[@class='search-summary']")[0]
         assert self._normalize_whitespace(self._squashed_element_text(ss_elem)) == "2 opportunities"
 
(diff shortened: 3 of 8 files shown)


app/templates/search/_briefs_results.html
```diff
@@ -18,7 +18,7 @@ <h2 class="search-result-title">
             {{ brief.lotName }}
         </li>
         <li class="search-result-metadata-item">
-            {{ brief.frameworkName }}
+            {{ brief.specialistRole }}
```
TheDoubleK: Are we happy not showing anything here at all if it's not on the specialists lot?  Because for all other lots this won't exist and we end up with an empty list item.

tests/fixtures/dos_multiple_briefs_fixture.json
```diff
@@ -3,6 +3,7 @@
     {
       "applicationsClosedAt":"2016-03-24T00:00:00.000000Z",
       "backgroundInformation":"Here is the first requirements summary.",
+      "specialistRole":"Content Specialist",
```
TheDoubleK: I don't think this is how `specialistRole` looks in serialized briefs.  The stored values are things like `"specialistRole": "contentDesigner"` and `"specialistRole": "businessAnalyst"`, so our test fixtures should reflect this.

tests/main/views/test_marketplace.py
```diff
@@ -864,10 +864,10 @@ def test_catalogue_of_briefs_page(self):
         ss_elem = document.xpath("//p[@class='search-summary']")[0]
         assert self._normalize_whitespace(self._squashed_element_text(ss_elem)) == "2 opportunities"
 
-        dos1_framework_label = document.xpath("//div[@class='search-result']/ul[2]/li[2]/text()")[0].strip()
-        dos2_framework_label = document.xpath("//div[@class='search-result']/ul[2]/li[2]/text()")[1].strip()
-        assert dos1_framework_label == "Digital Outcomes and Specialists"
-        assert dos2_framework_label == "Digital Outcomes and Specialists 2"
+        first_brief_specialist_role = document.xpath("//div[@class='search-result']/ul[2]/li[2]/text()")[0].strip()
+        second_brief_specialist_role = document.xpath("//div[@class='search-result']/ul[2]/li[2]/text()")[1].strip()
+        assert first_brief_specialist_role == "Content Specialist"
```
TheDoubleK: "Content Specialist" isn't a role that exists in DOS2 briefs.  We should test for some of the actual roles listed here: https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-2/questions/briefs/specialistRole.yml

TheDoubleK:
I tried running this, and it is displaying the `value` direct from the brief data:
screen shot 2017-06-01 at 17 43 36

Instead we'll need to load the `label` for the corresponding value from the content here: https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-2/questions/briefs/specialistRole.yml

The catalogues have just done something similar for displaying labels instead of values on the G9 service pages, so it might be worth having a quick chat with Sam or George about how this works - I think there might be a method in the ContentLoader that you can use to get this easily, although I'm not 100% sure it will work right out of the box.

app/templates/search/_briefs_results.html
TheDoubleK: I had a look and I think it's fine to show nothing for other lots:
screen shot 2017-06-01 at 17 54 59

tests/fixtures/dos_multiple_briefs_fixture.json
idavidmcdonald: This is a brief for the user research participants lot. We only have specialist roles for the digital specialists lot so we don't want this key in here for this brief

tests/fixtures/dos_multiple_briefs_fixture.json
```diff
@@ -44,6 +45,7 @@
       "additionalTerms":"You'll never work in this town again.",
       "applicationsClosedAt":"2016-03-22T00:00:00.000000Z",
       "backgroundInformation":"We'd like to build a thing to do athing.",
+      "specialistRole":"Content Specialist 2",
```
idavidmcdonald: Would be nice to have this as one of our real specialist roles e.g.
`"specialistRole": "deliveryManager"`

Note here that data coming back from the api sends the role type as camel case. However in our frontend we are going to need to show the user the pretty string of `Delivery manager` like we do on the page to view a brief it self:

image

This converting to a pretty string on the brief page itself might be done via the content loader but I might need to have a little look tomorrow to figure out how...

tests/fixtures/dos_multiple_briefs_fixture.json
TheDoubleK: Yes, I meant to say that too :P

tests/main/views/test_marketplace.py
```diff
@@ -864,10 +864,10 @@ def test_catalogue_of_briefs_page(self):
...
+        assert second_brief_specialist_role == "Content Specialist 2"
```
idavidmcdonald: These will need updating based on my comments regarding fixtures not always having a specialist role and also un-camelcasing it.

tests/fixtures/dos_multiple_briefs_fixture.json
idavidmcdonald: Oh and it would be good to keep our keys in alphabetical order please :)

app/templates/search/_briefs_results.html
```diff
@@ -18,7 +18,7 @@ <h2 class="search-result-title">
             {{ brief.lotName }}
         </li>
         <li class="search-result-metadata-item">
```
idavidmcdonald: We should only include the `li` if it has content to show. Some of the briefs won't have a specialist role as they are not on the correct lot so probably need to wrap this `li` in some kind of if statement

idavidmcdonald:
Ah, you beat me to it Kev. Looks like we spotted all the same stuff though. I didn't know about the g9 stuff that had been done though, that's a good place to look regarding labels.

app/main/views/marketplace.py
```diff
@@ -118,7 +116,11 @@ def list_opportunities(framework_framework):
 
     api_result = form.get_briefs()
 
-    briefs = api_result["briefs"]
+    briefs = [{
```
TheDoubleK: It feels weird to have to pass through both the raw data and content to the template.  Does the content alone not have everything we need to display the brief?  And if not, can we make the content have everything we need to display the brief?

app/main/views/marketplace.py
```diff
@@ -118,7 +116,11 @@ def list_opportunities(framework_framework):
...
+        "data": brief,
+        "content": content_loader.get_builder(brief['frameworkSlug'], 'display_brief').filter(brief)
```
TheDoubleK: I'm relatively sure (about 95%) that `get_builder` is deprecated and we should use `get_manifest` everywhere these days.

app/main/views/marketplace.py
idavidmcdonald: The content does not contain everything we need because we need attributes such as `publishedAt` and `title`. `publishedAt` is not a question in any manifest as it is an attribute on the brief only. `title` is a question for the edit_brief manifest but not for the display_brief manifest.

We have copied the pattern for a single brief of doing this - https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/main/views/marketplace.py#L92

However we did spend a bunch of time deliberating what the best option was as this does feel a bit complicated. We did consider the simplest of overriding `brief['specialistRole'] = thing_formatted_by_content_loader` before passing the brief into the template but thought this seemed to be more presentation related and should maybe live in the template which therefore requires passing both the brief and the content in.

Any suggestions or content loader tricks we aren't aware of?

TheDoubleK: I'd say that ideally everything needed to display a brief should be in the "display brief" manifest, and that would be the only thing passed to the template. For services I know there are some "fake" questions in the frameworks content, never rendered as form questions but just there for display manifests (e.g. lotName I think). Maybe something like that could work here?  Or is everything in the manifest going to appear in the summary table part of the brief page and so extra fake questions will make a mess of that?

I'm not 100% against this as it is, but it does feel like a proliferation of slightly different "special" ways of doing things, which it would be nice to avoid where possible.

idavidmcdonald: I searched for 'lotName' in the frameworks repo but didn't find it included in any display manifests so not quite sure where to look. I may have to ask one of the catalogues team but it seems like a potential solution.

Based on our required fields for this view, using 'fake questions', I can see how we could drive it by a new manifest just for this page for
`id`, `title`, `summary`, `organisation`, `location`, `lotName`, `specialistRole` but I don't know how it would handle some of the logic on the page using the brief's status.

```
{% if brief.status == 'closed' %}
                Closed
            </li>
        {% else %}
                Published: {{ brief.publishedAt|dateformat }}
            </li>
                Closing: {{ brief.applicationsClosedAt|dateformat }}
            </li>
        {% endif %}
```
I don't think the content loader can do the logic around brief status at the moment, which means we are still passing in the brief and the content summary. The content loader also doesn't have a way of formatting these datetimes (we have only a `Date` question rather than a `DateTime` question). We could go to the effort of adding new question types for these fields which can do this logic and parse datetimes but I'm not sure if that feels like a good idea or not...

TheDoubleK:
Fine, if we need the brief data anyway let's do it like this :+1:

idavidmcdonald:
Cool, cheers @TheDoubleK. Although I agree something doesn't feel quite right so I am going to bring up in the tech debt meeting.

PR #538, opened 2017-06-02
Changing category preserves filters

galund:
Preserve filters (other than category) when changing categories.
 - also preserve filters when changing lot (other than category and lot!)
 - also works for 'All categories' filter (involved some changes / simplification in the templates)
 - we now create 'lot filters' from the original lots, rather than annotate those lot objects, which was always a bit weird;

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -96,7 +96,7 @@ def _get_category_filter_key_set(category_filter_group):
     """
     Returns the set of keys used in the category filter group. In practice
     there should only be one, but for completeness we allow for the possibility
-    of their being several. In G6/7/8, this was {'serviceTypes'}, in G9 it's
+    of there being several. In G6/7/8, this was {'serviceTypes'}, in G9 it's
     {'serviceCategories'} .
     :return: set[string]
     """
@@ -143,7 +143,7 @@ def get_lots_and_categories_selection(lots, category_filter_group, request):
     for lot in lots:
         selected_categories = []
         lot_selected = (lot['slug'] == current_lot_slug)
-        lot_filter = dict(lot)
+        lot_filter = lot.copy()
         if lot_selected:
             lot_filter['selected'] = True
             categories = category_filter_group['filters'] if category_filter_group else []
@@ -160,7 +160,6 @@ def get_lots_and_categories_selection(lots, category_filter_group, request):
             lot_filter['link'] = search_link_builder(url_args)
 
         if lot_selected or current_lot_slug is None:
-            # only place this lot into the tree if it is selected (or if no lot is selected at all)
             root_node['children'].append(lot_filter)
 
     return selection
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -112,10 +112,13 @@ def build_lots_and_categories_link_tree(lots, category_filter_group, request):
     lots/categories widget. Adds links (where necessary) and shows the currently-selected
     lot and/or categories.
 
+    Returns a list of selected filters: there will always be at least one element, which is
+    the root of the tree; there may then be a category; there may then follow a sub-category.
+
     :param lots: a sequence of lot dicts applicable to the live framework(s)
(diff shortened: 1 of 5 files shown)


tests/main/presenters/test_search_presenters.py
```diff
@@ -198,22 +199,26 @@ def test_show_lots_and_categories_selection(self):
...
-            assert lots[1].get('slug') == 'cloud-software'
-            assert lots[1].get('selected')
-            category_links = lots[1]['categories']
-            assert 'checkboxTreeExample=option+1' in category_links[0]['link']
-            assert not category_links[0].get('children')
-            assert 'checkboxTreeExample=option+2' in category_links[1]['link']
-            sub_categories = category_links[1]['children']
-            assert 'checkboxTreeExample=option+2.1' in sub_categories[0]['link']
-            assert 'checkboxTreeExample=option+2.2' in sub_categories[1]['link']
+        with self.app.test_request_context("/g-cloud/search?q=&lot={}&otherfilter=somevalue".format(lot_slug)):
```
samuelhwilliams: This test is kinda dense to parse, might be useful to add a couple of comments.

tests/main/presenters/test_search_presenters.py
```diff
@@ -194,26 +195,52 @@ def test_request_filters_are_set(self):
...
+            assert selected_lot.get('name') == 'Cloud software'
+            # there should be a link to the lot without a category, as a category has been selected within it
+            assert 'lot={}'.format('cloud-software') in selected_lot['link']
+            assert 'checkboxTreeExample' not in selected_lot['link']
+
+            # check that we have links in place to a search with the relevant category filter applied,
+            # except to the currently-selected category
+            category_filters = selected_lot['children']
+            assert 'link' not in category_filters[0]
+            assert not category_filters[0].get('children')
```
samuelhwilliams: Can we do the same as the line above to be consistent/pin this more precisely? Eg `'children' not in ...`. Or, if it should be there but empty, `== []/is None`. Quite a few things could be Falsey.

galund: As these things are getting punted into the templating layer, it seems safest to allow None/empty/not-even-there to all be equivalent.

One difference is that 'children' is derived from 'options' in the original framework data, and it's optional, so as a point of principle we liberally accept it not being there, or being there but empty (see `search_presenters._recursive_add_option_filters`, which I am reluctant to touch right now!!). Whereas it's our tree-builder which adds the 'link' value (or, sometimes, which shouldn't be adding it.)

app/main/presenters/search_presenters.py
```diff
@@ -91,45 +92,99 @@ def set_filter_states(filter_groups, request):
...
     """
     current_lot_slug = get_lot_from_request(request)
-    selected_category = None
+
+    # Links in the tree should preserve all the filters, except those relating to this tree (i.e. lot
+    # and category).
+    search_link_builder = Href(url_for('.search_services'))
+    keys_to_remove = _get_category_filter_key_set(category_filter_group)
+    keys_to_remove.add('lot')
+    keys_to_remove.add('page')
```
samuelhwilliams: I wonder if we can do this non-magically, e.g. by removing/preserving all args that aren't used as filters - just in case we have or add something else in the future that shouldn't be affected.

Lot would probably still have to be magical for the time being.

samuelhwilliams: `_get_category_filter_key_set` + `set(chain.from_iterable([[x['name'] for x in v['filters']] for k,v in filters.items()]))` seems to be, logically, what gives us all the keys that are valid for filtering/categories... though it's pretty dense to interpret. :trollface: 

Up to you...

galund: Deciding not to do this. If there are more things that need to be removed/special-cased, then at that point we'd go for this kind of treatment I think.

samuelhwilliams: As per chat, let's leave this out for now and try to remember to re-visit if more params become relevant. :)

PR #543, opened 2017-06-11
Add aggregations to service categories

samuelhwilliams:
## Summary
We are surfacing service counts next to the categories on the G-Cloud search page so that users can see how many results they can expect to get.
Updated tests to run against G-Cloud 9 rather than G-Cloud 8.

## Key criteria
* Aggregation counts should always reflect what the user will see when they click the link (i.e. filters must be applied correctly).
* The counts will only show on categorisation levels below that currently selected, as it is intended to help buyers decide where to go next to refine their search.
* Any categories that do not have results will no longer generate a link, preventing users from making choices that will not help them.
* Filters should be retained in generated URLs only if they are applicable to the context of the link (e.g. filters specific to cloud-hosting should not be retained in the URL generated for the 'All categories' link).

## Examples
screen shot 2017-06-12 at 08 45 45

screen shot 2017-06-12 at 08 45 31

screen shot 2017-06-12 at 08 45 38

screen shot 2017-06-12 at 10 50 12

screen shot 2017-06-13 at 11 43 08

## Ticket
https://trello.com/c/5LRAkydD/481-show-number-of-services-within-category


samuelhwilliams:
@karlchillmaid Do you have any issues with the aggregations showing in parenthesis next to acronyms? Ala `Enterprise resource planning (ERP) (297)`?

karlchillmaid:
In isolation it could be confusing, but in context with all the other categories it's probably ok.

Categories with acronyms are the exception, rather than the rule, aren't they?

samuelhwilliams:
Yes, they should be (and appear to be).

app/main/presenters/search_presenters.py
```diff
@@ -178,6 +253,7 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, pre
     :param category_filters: iterable of category filters as previously produced by filters_for_question
     :param request: request object from which to extract active filters
     :param preserved_request_args: MultiDict of arguments to be preserved when generating links
```
galund: We're no longer passing in preserved_request_args I think?

app/main/presenters/search_results.py
```diff
@@ -27,3 +27,13 @@ def _add_highlighting(self, service):
...
+
+
+class AggregationResults(object):
+    """Provides access to the aggregation results information"""
+
+    def __init__(self, response):
+        self.results = response['aggregations']
+        self.total = response['meta']['total']
+        if 'page' in response['meta']['query']:
+            self.page = response['meta']['query']['page']
```
galund: If we're making objects, I'd prefer them to have a consistent interface - in this case, having `page=None` rather than not having that field at all.

So in this case maybe something like `self.page=response['meta']['query'].get('page')`?

app/templates/search/_categories.html
```diff
@@ -1,17 +1,17 @@
-{%- macro top_level_link(text, url) -%}
+{%- macro top_level_link(text, count, url) -%}
   <a href="{{ url }}" class="lot-filters__top-level-link">{{ text }}</a>
 {%- endmacro -%}
 
-{%- macro list_item__unselected(text, url) -%}
-  <li><a href="{{ url }}">{{ text }}</a></li>
+{%- macro list_item__unselected(text, count, url) -%}
+  <li>{% if count > 0 %}<a href="{{ url }}">{% endif %}{{ text }} ({{ count }}){% if count > 0 %}</a>{% endif %}</li>
```
galund: It'd be less verbose just to write `{% if count %}` but I guess this is personal preference :)

app/main/presenters/search_presenters.py
```diff
@@ -188,15 +264,17 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, pre
 
     for category in category_filters:
         category['selected'] = False
+        category['count'] = aggregations.results.get('serviceCategories', {}).get(category['label'], 0)
```
galund: I'd maybe prefer `'service_count'` rather than `'count'`

app/main/presenters/search_results.py
samuelhwilliams: This interface is consistent with the SearchResults object above it.

app/templates/search/_categories.html
samuelhwilliams: Less verbose and less explicit. :)

app/main/presenters/search_results.py
samuelhwilliams: On second inspection... There are tests asserting that page isn't set, so I will leave it as-is for now because it's 'expected' behaviour.

galund: Seems a pity, but yep your call

PR #544, opened 2017-06-14
Add basic Google Tag Manager configuration.

galund:
- this GTM configuration is controlled per-environment from config.py;
 - this is because in principle GTM can inject completely arbitrary JS into your pages, so it allows for a more strict separation of environments. (Compare with GA itself where the 'live' view is based on ‘what was the URI for that event/pageview’: in GTM you get to control which ‘tags’ (arbitrary JS) are deployed to which of your environments.)
 - test to verify that the GTM data layer is set up as expected;
 - removed test on Google Analytics that literally just checked the implementation details of the GOVUK.GDM.analytics.init function.

https://trello.com/c/fRM4B4wm/462-tag-manager-initial-set-up-on-staging-and-live


config.py
```diff
@@ -137,14 +147,17 @@ class Live(Config):
 
 class Preview(Live):
     FEATURE_FLAGS_NEW_SUPPLIER_FLOW = enabled_since('2017-02-06')
+    GOOGLE_TAG_MANAGER = dict(Config.GOOGLE_TAG_MANAGER, environmentID='env-7', authToken='Kpjle7g00efQOB2j1o3fVQ')
```
Wynndow: Where do the environment ID's get set? They seem to be in a weird order. Although, I'm not sure what a 'correct' order would be...

galund: They are generated auto-inc keys in GTM, as far as I can see. Some environments are created automatically: "Live" and "Latest" (which I am taking to mean "Dev") are there from the start. Then I added Staging and Preview as two more custom ones. I might have created and deleted one or two more in between :) So yeah it's a bit random.

TheDoubleK: Should these actual auth tokens live in the credentials repo rather than a public Github? (And probably be rotated at the same time?)

config.py
```diff
@@ -82,6 +82,15 @@ class Config(object):
         'g-cloud-5': 'g-cloud-6',
     }
 
+    # Note that because our GTM auth token is rendered into the client environment (see
+    # _javascripts.html), there isn't any point in encrypting it, as we would with other
+    # similar tokens.
```
galund: @TheDoubleK does this comment answer your question?

TheDoubleK: The `gtmEnvironmentIDShouldBeOverriddenByLocalConfig` confused me, as this seemed to be a deliberately obfuscating thing.

galund: Happy to change how I've done that, it was an attempt to be super-clear about what's happening. (Fail! - sorry!)

Incidentally I don't know of any way to rotate the keys, everything about GTM seems slightly meh :-)

TheDoubleK: If they're public anyway probably no need to ever rotate them I guess!

galund: Which is good because I don't think there's a mechanism to rotate them, at least I didn't see it.

@samuelhwilliams said he'd been reading some things about GTM spoofing/security so we'll discuss those before this lands.

PR #546, opened 2017-06-19
Use section_id rather than section_slug in URL params

TheDoubleK:
I noticed that the GET and POST views for one of our routes used different parameter names, which could be confusing:
```
GET 
 /buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/edit/<section_slug>/<question_id>
POST 
 /buyers/frameworks/<framework_slug>/requirements/<lot_slug>/<brief_id>/edit/<section_id>/<question_id>
```

I checked across apps and everywhere else (supplier and admin apps) we use `section_id` as a parameter rather than `section_slug`, so I went with keeping `section_id` for the views here.

### DIFF
--- a/app/buyers/views/buyers.py
+++ b/app/buyers/views/buyers.py
@@ -136,7 +136,7 @@ def copy_brief(framework_slug, lot_slug, brief_id):
         framework_slug=new_brief["frameworkSlug"],
         lot_slug=new_brief["lotSlug"],
         brief_id=new_brief["id"],
-        section_slug=section.slug,
+        section_id=section.slug,
         question_id=section.questions[0].id
     ))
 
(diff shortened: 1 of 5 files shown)


TheDoubleK:
@benvand - it suggested you as a reviewer, and this is a "briefs" kind of route, so you're the lucky one :)

TheDoubleK:
In terms of the buyer app there's only these two views to resolve section_slug/section_id, but there are another ten between the supplier and admin apps that would need to change if we're going for consistency across all our views (none of the views there use `section_slug`).

I'm not going to do them now.  If you want to just leave this until the briefs app happens (after firebreak sometime?) then fine.  But it is weird to have the GET and POST defined differently for the exact same URL.

idavidmcdonald:
Yep, these routes just got deleted by core. 

What I'll do is add a tech debt ticket to the briefs sprint board to resolve this, either by changing just this route as you've done or going through all the apps and changing them all.

Thanks for spotting!

PR #547, opened 2017-06-20
Remove briefs-frontend endpoints and templates

allait:
### Add external views for brief-frontend endpoints

Adds buyer dashboard and DOS lot start pages to the external views.
The URLs need an explicit `/buyers/` prefix since the blueprint

### Remove CF pip dependencies

We don't need gunicorn and awscli installed through pip in the
docker containers.

Also removes `--pull` from the docker-build step since we've
switched to using versioned base images.


app/__init__.py
```diff
@@ -42,7 +42,7 @@ def create_app(config_name):
                 content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
 
     from .main import main as main_blueprint
-    from .main import external as external_blueprint
+    from .external.views.external import external as external_blueprint
```
Wynndow: Yo Dawg

PR #548, opened 2017-06-20
Bring in frontend toolkit updates

carolinegreen:
Bring in frontend toolkit updates with no javascript buttons.  Also add GOV.UK Sass Elements as a dependency.

Relies on: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/353


Wynndow:
Not really related to this PR but`"test" : "./node_modules/gulp/bin/gulp.js test"` appears twice in the build script. We can remove one.

galund:
Reopening so that the buyer app can be brought up to date with future fe-toolkit changes.

PR #549, opened 2017-06-21
Comprehensive test for every filter in G9.

galund:
Unit test to catch some of the crashes we had when we had missing data in the search summary manifest. It's G9 specific, and not much interest once we refactor the search summary, probably.

Ideally we would also have a higher-level test than this to check that the entire page renders okay for each and every filter, without particular reference to the implementation of the search summary.

### DIFF
--- a/tests/main/presenters/test_search_summary.py
+++ b/tests/main/presenters/test_search_summary.py
@@ -266,6 +266,38 @@ class TestSearchSummary(BaseApplicationTest):
         summary_markup = search_summary.markup()
         assert "category <em>Analytics</em>" in summary_markup
 
+    def test_each_filter(self):
+        """
+        Test each filter individually adds a single text string to the summary text
+        """
+        for lot in self._g9_lots_by_slug:
+            self.request_args['lot'] = lot
+            for filter_group in g9_filter_groups:
+                for f in filter_group['filters']:
+                    request_args = self.request_args.copy()
+                    request_args[f['name']] = f['value']
+                    search_summary = SearchSummary('9', request_args, g9_filter_groups, self._g9_lots_by_slug)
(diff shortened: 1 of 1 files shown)


tests/main/presenters/test_search_summary.py
```diff
@@ -266,6 +266,38 @@ def test_subcategory_filters_are_available(self):
...
+        """
+        for lot in self._g9_lots_by_slug:
+            self.request_args['lot'] = lot
+            for filter_group in g9_filter_groups:
+                for f in filter_group['filters']:
+                    request_args = self.request_args.copy()
+                    request_args[f['name']] = f['value']
+                    search_summary = SearchSummary('9', request_args, g9_filter_groups, self._g9_lots_by_slug)
+                    summary_markup = search_summary.markup()
+                    assert summary_markup.count("<em>") == 3  # the keyword, the lot, and one filter
```
samuelhwilliams: Heh heh heh.

I see what this is doing... but it feels a bit weird to rely test it this way. Doesn't feel like it would be completely insane for us to decide one/more of these shouldn't be bold, and we'd break the test without it functionally breaking. If we're validating that the correct labels are displayed somewhere else, maybe this would be OK after adding some extra comments.

galund: I think I'm reluctant to try and get something better until after we've refactored the widget completely. The current design has been in place for a while, and this would catch the cases that we broke in/after G9 go-live.

We can't very well validate that the correct labels are going in without reproducing the entire logic from the code that processes the summary manifest. Again I'm reluctant to go there until we've refactored that (and potentially moved the content into the frameworks repo, depending if any such separate content continues to exist).

PR #550, opened 2017-06-22
Bugfixes for G-Cloud search

samuelhwilliams:
## Summary
Three bugfixes and one optimisation for G-Cloud search. It's probably best to review these per-commit as they should be self-contained that way.

1) Page 2 on DM shows zero results for service aggregations
2) Searching with an invalid lot shows broken links
3) Subcategories no longer pass the parentCategory query param.
4) Reduce the number of search-api aggregation calls when a lot has already been selected.

## Ticket
https://trello.com/c/WPzIXFIk/557-bug-category-links-g8-services-pagination-and-parent-categories

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -4,8 +4,9 @@ from math import ceil
 from werkzeug.datastructures import MultiDict
 
 
-def get_lot_from_request(request):
-    return request.args.get('lot', None)
+def get_lot_from_request(request, all_lots):
+    lot = request.args.get('lot', None)
+    return lot if (not lot or lot in all_lots) else None
 
 
 def get_keywords_from_request(request):
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -50,7 +50,7 @@ def allowed_request_lot_filters(lot_filters):
     return filters
 
 
-def clean_request_args(request_args, lot_filters, lots_by_slug):
+def clean_request_args(request_args, lot_filters, lots_by_slug, for_aggregation=False):
     """Removes any unknown args keys or values from request.
 
     Compares every key/value pair from request query parameters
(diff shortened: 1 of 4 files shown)


app/main/presenters/search_presenters.py
```diff
@@ -204,9 +204,13 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
...
     selected_filters.append(root_node)
 
+    # If we're searching against a specific lot, we should only build the tree for that lot.
+    if current_lot_slug:
+        lots = list(filter(lambda lot: lot['slug'] == current_lot_slug, lots))
+
     aggregations_by_lot = {lot['slug']: _get_aggregations_for_lot_with_filters(lot['slug'], content_manifest,
                                                                                framework, request)
-                           for lot in framework['lots']}
+                           for lot in lots}
```
MuriloDalRi: Neat!

samuelhwilliams: Possibly the best change I've ever made.

app/main/helpers/search_helpers.py
```diff
@@ -4,8 +4,9 @@
 from werkzeug.datastructures import MultiDict
 
 
-def get_lot_from_request(request):
-    return request.args.get('lot', None)
+def get_lot_from_request(request, all_lots):
+    lot = request.args.get('lot', None)
+    return lot if lot in all_lots or not all_lots else None
```
TheDoubleK: I think this will this break if `all lots` is `None`, something like a "NoneType is not iterable" error.  It should be OK if you do the "not all_lots" check first, i.e.: `lot if not all_lots or lot in all_lots else None`

TheDoubleK: And I know Python hates brackets but some parentheses might help make it more readable?
`lot if (not all_lots or lot in all_lots) else None`

samuelhwilliams: Damn, yeah. I had the not check first, and then moved it because I didn't like starting with a not. My bad. Good spot.

PR #552, opened 2017-07-03
Preserve line breaks for textarea formatted text

pcraig3:
We're going to allow newline formatting for `textbox_large` content submitted to us and displayed back to users.

In this pull request, we're allowing newlines in:

- textbox_large areas of individual brief pages
  (brief summary, workplace description, etc.)
- clarification questions and answers for briefs
- textbox_large areas of individual G8/G9 service pages
  (includes most of the follow-up questions)

Where we can, we're checking if the question type is `textbox_large` before applying the filter. But on the brief questions and answers, for example, we don't have a question object to inspect -- only a list of strings. We know they're all textbox large questions anyway, so there we just use the filter without checking.

We are *not* adding this to search results because search result rankings are a competitive space and we don't want there to be an incentive to create very large service summaries and/or brief descriptions.

## example

This example is a lot like [the one in the briefs app](https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/5), but here we also have a search result page.

### with this json 

```json
  {"summary": "As Engagement Lead, you will:\r\n\r\n* Lead on DAS engagement  and 
    DAS stakeholder engagement requirements with stakeholders and as part of the 
    multidisciplinary delivery team.\r\n\r\n* Create briefing and communications 
    materials for inter-team and cross-government engagement\r\n\r\n* Represent needs 
    of stakeholders and communications activity in the DAS senior management team"
  }
```
<sub>ie, [Engagement Lead](https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities/127)</sub>

| before & after |
|--------|
| screen shot 2017-07-03 at 14 32 57 |

| search results are _not_ formatted |
|--------|
| screen shot 2017-07-03 at 14 33 24 |

### DIFF
--- a/app/templates/_brief_attributes.html
+++ b/app/templates/_brief_attributes.html
@@ -33,7 +33,9 @@
     ) %}
       {% call summary.row() %}
         {{ summary.field_name(item.label) }}
-        {{ summary[item.type](item.value) | format_links}}
+        {% with item_value = summary[item.type](item.value) | format_links %}
+            {{ (item_value | preserve_line_breaks) if item.type == 'textbox_large' else item_value }}
+        {% endwith %}
       {% endcall %}
     {% endcall %}
 {% endfor %}
--- a/app/templates/_brief_q_and_a.html
+++ b/app/templates/_brief_q_and_a.html
@@ -25,9 +25,9 @@
   {% call summary.row() %}
     {% call summary.field(first=True, wide=False) -%}
       <span aria-label="question">{{ question.number }}.</span>
-      {{ question.question }}
+      {{ question.question | preserve_line_breaks }}
     {%- endcall %}
-    {{ summary.text(question.answer | format_links) }}
+    {{ summary.text(question.answer | format_links | preserve_line_breaks) }}
   {% endcall %}
 {% endcall %}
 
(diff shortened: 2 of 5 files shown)


app/templates/_brief_attributes.html
```diff
@@ -33,7 +33,9 @@
     ) %}
       {% call summary.row() %}
         {{ summary.field_name(item.label) }}
-        {{ summary[item.type](item.value) | format_links}}
+        {% with item_value = summary[item.type](item.value) | format_links %}
+            {{ item_value | preserve_line_breaks if item.type == 'textbox_large' else item_value }}
```
samuelhwilliams: Some parenthesis might be nice, if supported, to show that `item_value | preserve_line_breaks` is definitely interpreted as one item.

pcraig3:
Lots of people will just type in the bullet point character to make things that look like lists.  From what I've seen, many more people do that than use asteriskxes. 

The problem then would be that we could automatically put bullet points for people but then we would possibly end up with 
`<li>• We are the leading hyper cloud provider</li>` 
which would render onscreen as 
`• • We are the leading hyper cloud provider`.

So the easy thing to do is not try to format anyone's text as HTML at all -- not try to guess their intention and fit it into semantic elements -- but just make it look more like whatever they wrote into the textarea.

PR #554, opened 2017-07-05
Add default.nix

risicle:
See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/719

This should be near-identical, probably best to have any discussion there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 *.pyc
 .vimrc
 .DS_store
-venv
+venv*
 node_modules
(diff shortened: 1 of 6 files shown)


risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

TheDoubleK:
This worked for me as well as I can tell. Let's get these in 👍

risicle:
Nix 4 Life

PR #555, opened 2017-07-05
Flash messages move

galund:
Move flash messages into single location for sharing between apps.

 - we want to be able to set flash messages from anywhere;
 - therefore this needs to be centralised, despite therefore various login-related messages now appearing in a general template;
 - potential for further refactoring to remove the 'named messages' concept, e.g. the message text should constants in the app.main.views.login module.
 - added some tests for login messages that were not already tested.

https://trello.com/c/Uak7y047/8-feedback-forms

See also https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/355

### DIFF
--- a/app/main/views/login.py
+++ b/app/main/views/login.py
@@ -1,9 +1,12 @@
+# coding: utf-8
+from __future__ import unicode_literals
 from __future__ import absolute_import
 
 import six
 
 from flask_login import current_user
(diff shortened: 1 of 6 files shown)


risicle:
I do think this is broadly sensible. Fact is, there are a million ways to implement this, all with minor disadvantages, but this seems as ok as any of them. Solves the cross-app issue.

PR #558, opened 2017-07-11
Implement filter-on-click

samuelhwilliams:
# Summary
Adds filter-on-click functionality to the G-Cloud catalogue. Borrows heavily from the GOV.UK Finder-Frontend, but has been adapted to allow the replacement of an arbitrary set of elements with pre-rendered HTML so that we take advantage of our server-side templating, which does quite a bit of work, especially when rendering the lot/categories links and numbers.

## Details
The frontend-toolkit pattern for this expects the view to detect a `live-results` query parameter, and based on that, switches to returning a JSON blob for all of the dynamic parts of the part. Some refactoring was required to split the dynamic parts into individual templates that can be rendered separately. Other than that refactoring, this just pulls in the new frontend-toolkit and implements the JSON return segment of the G-Cloud search view.

## How to test
Get everything up and running using this PR (no special linking necessary). Navigate to `/g-cloud/search` and play around with the filters and the text search. The page results and related navigation links should all update accordingly.

## Example
filter on click - high - 640

## Ticket
https://trello.com/c/SGxQeBdA/502-filter-on-click

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -86,15 +86,16 @@ def set_filter_states(filter_groups, request):
     request_filters = get_filters_from_request(request)
 
     for filter_group in filter_groups:
-        for filter in filter_group['filters']:
-            filter['checked'] = False
+        for filter_item in filter_group['filters']:
+            filter_item['checked'] = False
+
             param_values = request_filters.getlist(
-                filter['name'],
+                filter_item['name'],
                 type=str
             )
             if len(param_values) > 0:
-                filter['checked'] = (
-                    filter['value'] in param_values
+                filter_item['checked'] = (
+                    filter_item['value'] in param_values
                 )
 
 
--- a/tests/main/views/test_search.py
+++ b/tests/main/views/test_search.py
@@ -1,4 +1,5 @@
 from lxml import html
+import json
 import mock
 import re
 import pytest
@@ -52,7 +53,7 @@ class TestSearchResults(BaseApplicationTest):
 
         res = self.client.get('/g-cloud/search?q=email')
         assert res.status_code == 200
-        assert '<form action="/g-cloud/search" method="get">' in res.get_data(as_text=True)
+        assert '<form action="/g-cloud/search" method="get" id="js-dm-live-search-form">' in res.get_data(as_text=True)
 
     def test_search_page_allows_non_keyword_search(self):
         self._search_api_client.search_services.return_value = \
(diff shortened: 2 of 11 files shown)


app/main/presenters/search_presenters.py
```diff
@@ -81,13 +81,16 @@ def _recursive_add_option_filters(question, options_list, filters_list):
             filters_list.append(presented_filter)
 
 
-def set_filter_states(filter_groups, request):
+def set_filter_states(filter_groups, base_url, clean_request_query_params):
     """Sets a flag on each filter to mark it as set or not"""
```
idavidmcdonald: This description could be updated to say it's not just marking a filter as set or not but also setting the links for those filters

app/main/presenters/search_presenters.py
```diff
@@ -97,6 +100,10 @@ def set_filter_states(filter_groups, request):
                     filter['value'] in param_values
                 )
 
+            filter_params = request_filters.copy()
+            filter_params.update({} if filter['checked'] else {filter['name']: filter['value']})
```
idavidmcdonald: I didn't quite have as much time as I'd like to review this but I wasn't quite clear about what this line is doing...

tests/main/presenters/test_search_presenters.py
```diff
@@ -197,7 +200,11 @@ def test_request_filters_are_set(self):
             'booleanExample1': 'true'
         })
 
-        set_filter_states(search_filters, request)
+        lots_by_slug = get_lots_by_slug(self._get_framework_fixture_data('g-cloud-6')['frameworks'])
```
idavidmcdonald: Is there a reason we use g6? I'm guessing it makes no difference...

tests/main/presenters/test_search_presenters.py
```diff
@@ -302,21 +315,26 @@ def test_filter_groups_have_correct_state_when_changed(self):
         })
 
         search_filters = list(filters_for_lot('paas', g6_builder).values())
-        set_filter_states(search_filters, request)
+        lots_by_slug = get_lots_by_slug(self._get_framework_fixture_data('g-cloud-6')['frameworks'])
+        clean_request_query_params = clean_request_args(request.args, search_filters, lots_by_slug)
```
idavidmcdonald: This test and the one above it share a fair amount of duplication so you could consider whether it can be parametrized (saying that it might impact a bit too much on readability so maybe have a think but don't do it if it's too bad for readability/ease)

tests/main/views/test_search.py
```diff
@@ -522,3 +523,64 @@ def test_category_with_no_results_is_not_a_link(self):
...
+        render_template_patch.return_value = '<p>some html</p>'
+
+        res = self.client.get('/g-cloud/search')
+
+        urls = {'search/services.html'}
+        assert urls == set(x[0][0] for x in render_template_patch.call_args_list)
+
+    @mock.patch('app.main.views.g_cloud.render_template', autospec=True)
+    def test_live_results_switch_renders_correct_templates(self, render_template_patch):
+        render_template_patch.return_value = {'categories': {'selector': '#id', 'html': "<p>hi</p>"}}
```
idavidmcdonald: I know this is just some dummy data but is this realistic from render_template? Would we not expect some html rather then JSON? If so, would be less confusing as just HTML

tests/main/presenters/test_search_presenters.py
samuelhwilliams: They are very similar. I started out with them together as a parameterized test and I felt it was going to get too confusing, so I split them. In tests, I don't mind a little duplication if it makes them clearer and simpler.

tests/main/views/test_search.py
samuelhwilliams: Oh good spot, I mixed these up. It should just be HTML. 👍

PR #560, opened 2017-07-13
Time displayed for closing dates

katstevens:
The clarification question deadline and the application closing date both use a new date filter `utcdatetimeformat` from dmutils (see https://github.com/alphagov/digitalmarketplace-utils/pull/317). 

Closing dates are now shown as "Thursday 12 May 2016 at 11:59pm". If daylight savings time means the localised string would end up being "Friday 13 May 2016 at 12.59am", the new filter will force the timestamp to still display 11.59pm on the correct day, rather than 12.59am on the following day. This might reduce the window by an hour, but will hopefully reduce confusion about dates for the supplier.

Also includes some cleanup on the `test_marketplace` module (caught by Pycharm code inspect).

Ticket: https://trello.com/c/8CJAv25W/714-show-time-opportunity-closes

### DIFF
--- a/app/templates/_brief_attributes.html
+++ b/app/templates/_brief_attributes.html
@@ -2,8 +2,8 @@
 {% call(item) summary.list_table(
   [
     {"label": "Published", "value": brief.publishedAt|dateformat},
-    {"label": "Deadline for asking questions", "value": brief.clarificationQuestionsClosedAt|dateformat},
-    {"label": "Closing date for applications", "value": brief.applicationsClosedAt|dateformat}
+    {"label": "Deadline for asking questions", "value": brief.clarificationQuestionsClosedAt|utcdatetimeformat},
+    {"label": "Closing date for applications", "value": brief.applicationsClosedAt|utcdatetimeformat}
   ],
   caption="Important dates",
   field_headings=[
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ unicodecsv==0.14.1
 werkzeug==0.10.4
 odfpy==1.3.3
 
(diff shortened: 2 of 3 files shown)


katstevens:
@samuelhwilliams I've just had a look and the requirement overview template is in the `digitalmarketplace-briefs-frontend` repo, so wouldn't be part of this PR. It might make sense to create a small separate story for it?

Also, I checked with the team regarding the timezone display and the consensus seems to be not to include it - according to Tobi, framework applications are always done in GMT (which is what's displayed here).

TheDoubleK:
That's not true.  It depends what time of year it is.  
DOS2 was GMT:
https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/digital-outcomes-and-specialists-2/messages/homepage-sidebar.yml#L11

G-Cloud 9 was BST: https://github.com/alphagov/digitalmarketplace-frameworks/blob/master/frameworks/g-cloud-9/messages/homepage-sidebar.yml#L11

samuelhwilliams:
Yeah as Kev says, when we did G-Cloud 9 any datetimes that fell in BST were marked up as such. I really think we need to be displaying the timezone to negate any chance for confusion or miscommunication. At the end of the day it's a decision for your team, but it's something I feel pretty strongly about.

If the requirement overview needs to be done as a separate story, fine, but I think it could equally be snuck in under this one. Just depends if you think it would actually get done if a new story was made for it. It would be good to make sure we're being consistent and clear about date/times. As we both saw earlier - each of us had a separate interpretation of when the brief would close, neither of which were correct.

katstevens:
Aha, interesting! I'll discuss with the team.

katstevens:
I've amended the dmutils filter to append `GMT`: https://github.com/alphagov/digitalmarketplace-utils/pull/319 @samuelhwilliams do you want to take a look?

PR #561, opened 2017-07-14
Pin requirements

allait:
requirements.txt is generated using a Python 3 virtualenv to keep the py2 backports of Python 3 stdlib packages (like `enum34` and `ipaddress`) out. This leaves those packages unpinned, but shouldn't cause any issues otherwise.

### Reorganize Makefile and requirements files

Renames requirements.txt to requirements-app.txt and requirements_for_test.txt to requirements-dev.txt.

Removes any unused dependencies or dependencies provided by other packages since those will be listed in the full requirements.txt.

Updates Makefile to add freeze-requirements and rename targets to be more consistent with other apps.

(see alphagov/digitalmarketplace-api#615 for details)


galund:
I'm a bit confused about the "to keep the py2 backports of Python 3 stdlib packages (like enum34 and ipaddress) out" part.

enum34 *is* in the requirements.txt, but ipaddress is _not_. So how does ipaddress get installed for Python 2 now, assuming it's needed?

Makefile
```diff
@@ -8,53 +8,83 @@ else
...
-test_pep8: virtualenv
+.PHONY: test-pep8
+test-pep8: virtualenv
 	${VIRTUALENV_ROOT}/bin/pep8 .
 
-test_python: virtualenv
-	${VIRTUALENV_ROOT}/bin/py.test ${PYTEST_ARGS}
+.PHONY: test-python
+test-python: virtualenv
+	${VIRTUALENV_ROOT}/bin/py.test ${PYtest-ARGS}
```
Wynndow: Lower case.

allait:
@galund oh yeah, `enum34` is being installed even under python3 because we haven't upgraded the apiclient version (which would make it a conditional requirements) on all frontend apps. Since it seems to run without issues on Travis I'll leave it as is for now until we move the apps to Python 3.

requirements.txt doesn't limit the list of packages that can be installed, it just specifies the ones that MUST be installed. So if there's a dependency that's not listed it will still get installed, but the version will be whatever the package pulling it in asked for (which is usually unpinned). `ipaddress` is a conditional dependency of one of the app dependencies, so it only gets installed on pyhon2, which is why it's missing from requirements.txt generated from a python3 virtualenv.

galund:
What's weird about this is I *definitely* had problems with enum34 on travis at some point. But there was a specific point version of Python 3 that had the problem, and maybe the travis system has been upgraded now.

PR #562, opened 2017-07-17
Only add filter-on-click classes for g-cloud search

samuelhwilliams:


idavidmcdonald:
Do we want to add unit tests for this or are we happy with functional tests being the things that cover it?

samuelhwilliams:
We know functional tests catch this, which is good, so I was happy in that regard. But you're right to imply that it might save time to add a unit test so that developers catch it sooner locally. I'll add one.

tests/main/views/test_marketplace.py
```diff
@@ -1013,6 +1013,15 @@ def test_catalogue_of_briefs_page_404_for_framework_that_does_not_exist(self):
         assert res.status_code == 404
         self._data_api_client.find_frameworks.assert_called_once_with()
 
+    def test_briefs_search_does_not_have_js_hidden_filter_button(self):
+        res = self.client.get('/digital-outcomes-and-specialists/opportunities')
+        assert res.status_code == 200
+
+        document = html.fromstring(res.get_data(as_text=True))
+
+        filter_button = document.xpath('//button[@class="button-save"]')
```
idavidmcdonald: Just to check, does this definitely fail before we added the change in the template?

At a glance, this button will still exist in the content because the JS is the thing that hides it and therefore is the class match an exact match or a containing match (i.e. passes if you have classes `button-save` or `button-save another-class`

samuelhwilliams: Yep, the equality check here tests the entire class string. To match subsets you have to use `contains(@class, 'button-save')`.

PR #565, opened 2017-07-19
Dh 539 add zero results state

DilwoarH:
**Changes**
- Adds in suggestion message for zero results state.
- Adds in test for suggestion showing for 0 results.
- Adds in test for not showing when results are shown.

Ticket: https://trello.com/c/RDI7f8mf/539-add-zero-results-state

### DIFF
--- a/app/templates/search/_services_summary.html
+++ b/app/templates/search/_services_summary.html
@@ -2,4 +2,20 @@
   <p class="search-summary">
     {{ summary }}
   </p>
+
+  {% if total == 0 %}
+
+  <div class="search-suggestion">
+    <p>Suggestions:</p>
+    <ul class="list list-bullet">
+        <li>try removing filters</li>
+        <li>choose a different category</li>
+        <li>double-check your spelling</li>
+        <li>use fewer keywords</li>
+        <li>search for something less specific, you can refine your results later</li>
+    </ul>
+  </div>
+
+  {% endif %}
+
 </div>
--- a/tests/main/views/test_search.py
+++ b/tests/main/views/test_search.py
@@ -18,6 +18,23 @@ def find_search_summary(res_data):
         r'<span class="search-summary-count">.+</span>[^\n]+', res_data)
 
 
+def find_0_results_suggestion(res_data):
+    return re.findall(
+        r'<p>Suggestions:<\/p>', res_data)
+
+
+def get_0_results_search_response():
+    return {
+        "services": [],
+        "meta": {
+            "query": {},
+            "total": 0,
+            "took": 3
+        },
+        "links": {}
+    }
+
+
 class TestSearchResults(BaseApplicationTest):
     def setup_method(self, method):
         super(TestSearchResults, self).setup_method(method)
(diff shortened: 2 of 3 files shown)


tests/main/views/test_search.py
```diff
@@ -159,6 +164,38 @@ def test_should_render_summary_for_0_results_in_cloud_software_no_keywords(self)
...
+                "query": {},
+                "total": 0,
+                "took": 3
+            },
+            "links": {}
+        }
+
+        res = self.client.get('/g-cloud/search?lot=cloud-software')
+        assert res.status_code == 200
+        suggestion = find_0_results_suggestion(res.get_data(as_text=True))[0]
```
samuelhwilliams: Probably better to just store the list here and assert that it has one entry on the next line, rather than duplicate the string you're validating against in two disparate places. Also more consistent with your approach below.

app/templates/search/_services_summary.html
```diff
@@ -2,4 +2,20 @@
   <p class="search-summary">
     {{ summary }}
   </p>
-</div>
+
+  {% if total == 0 %}
+
+  <div class="search-suggestion">
```
samuelhwilliams: Looks like this doesn't capture any CSS and is appearing noticeably larger than other text on the page. I think on the prototype they're the same size, or at least _much_ more similar. We should probably use core-14/core-16 in new CCS under `app/assets/scss/_search-page.scss`. Add the core-14/core-16 includes from `scss/toolkit/search/_search-summary.scss`.

screen shot 2017-07-19 at 14 16 40

tests/main/views/test_search.py
```diff
@@ -159,6 +164,38 @@ def test_should_render_summary_for_0_results_in_cloud_software_no_keywords(self)
...
+    def test_should_not_render_suggestions_for_when_results_are_shown(self):
+        self._search_api_client.search_services.return_value = {
+            "services": [],
+            "meta": {
+                "query": {},
+                "total": 2,
+                "took": 3
+            },
+            "links": {}
+        }
```
samuelhwilliams: We have this same fixture in two tests; would be nice to de-duplicate it.

app/templates/search/_services_summary.html
DilwoarH: CSS updated in the toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/360/files

tests/main/views/test_search.py
DilwoarH: Code updated to reflect the suggestion.

tests/main/views/test_search.py
DilwoarH: hey wasn't too sure what you meant by this bit? could you expand by any chance?

If you meant that I should store the value of the response in a variable then use it on the next test then I intentionally did not do this. This is to make all tests independent and executable on their own.

DilwoarH: code has been updated to assert length of array rather than specific string

samuelhwilliams:
Code looks good now. Can you please squash the commits down to just one and then I'll give a thumbs up. Thanks.

samuelhwilliams:
(need to bump `bower.json`'s frontend-toolkit dependency to `23.0.3`)

DilwoarH:
Updated.

PR #567, opened 2017-07-24
Adds in Clear All button to clear checkbox selection.

DilwoarH:
Pre-requisite: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/361

As a buyer
I need to be able to remove all of the filters I have applied quickly
So that I can improve my search results

Should trigger filter-on-click

Acceptance criteria
- Add a clear all filters link next to 'Apply filters'
- When clicked, all filters are removed
- Content should match the prototype: "Clear all filters"
- It works when JS is turned off
- When JS is enabled clicking should not reload entire page but url should update.

Ticket: https://trello.com/c/IrcO0G3U/491-add-clear-all-filters-link

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -2,6 +2,8 @@ import re
 from math import ceil
 
 from werkzeug.datastructures import MultiDict
+from werkzeug.urls import Href
+from flask import url_for
 
 
 def get_lot_from_request(request, all_lots):
@@ -18,6 +20,33 @@ def get_page_from_request(request):
         return int(request.args['page'])
     except (KeyError, ValueError, TypeError):
         return None
+
+
+def get_request_url_without_any_filters(request, filters):
+    """
+    This function will returns the url path without any filters.
+    It will still retain the categories, keyword and lots parameters as well as any others included.
+
+    Args:
+        request: Request Object (import from flask) of the current request.
+        filters: list of all filters from digital marketplace framework
+
+    Returns:
+        URL path in string format.
+    """
+
+    all_request_filters = MultiDict(request.args.copy())
+
+    for section in filters:
+        for _filter in filters[section]['filters']:
+            all_request_filters.poplist(_filter['name'])
+
+    all_request_filters.poplist('page')
+
+    search_link_builder = Href(url_for('.search_services'))
+    url = search_link_builder(all_request_filters)
+
+    return url
 
 
 def get_filters_from_request(request):
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -20,7 +20,7 @@ from ..helpers.search_helpers import (
     get_keywords_from_request, pagination,
     get_page_from_request, query_args_for_pagination,
     get_lot_from_request, build_search_query,
-    clean_request_args
+    clean_request_args, get_request_url_without_any_filters
 )
 from ..helpers import framework_helpers
 
@@ -222,6 +222,8 @@ def search_services():
             if 'label' in filter_instance:
                 filter_instance['label'] = capitalize_first(filter_instance['label'])
 
+    clear_filters_url = get_request_url_without_any_filters(request, filters)
+
     template_args = dict(
         current_lot=current_lot,
         framework_family=framework['framework'],
@@ -236,7 +238,8 @@ def search_services():
         summary=search_summary.markup(),
         title='Search results',
         total=search_results_obj.total,
-        gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'))
+        gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'),
+        clear_filters_url=clear_filters_url)
 
     if request.args.get('live-results'):
         from flask import jsonify
--- a/app/templates/search/_filters.html
+++ b/app/templates/search/_filters.html
@@ -1,13 +1,18 @@
 <div class="dm-filters">
-  <h2 class="apply-filters-title">Apply filters</h2>
-  {% for filter in filters %}
-    {%
-      with
-      label = filter.label,
-      options = filter.filters
-    %}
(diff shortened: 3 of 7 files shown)


DilwoarH:
required: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/361

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,33 @@ def get_page_from_request(request):
...
+    '''
+        Logic:
+            -> get all url args
+            -> loop through the filters list and pop any matching url args
+            -> build and return url path
+
+        Arguements:
+            -> request - Request Object (import from flask)
+            -> filters - list of all filters from digital market place framework
+    '''
```
samuelhwilliams: A few Python style issues here:
1) Docstrings for methods should use triple double-quotes rather than single-quotes.
2) Don't really need to document the logic; we can work that out from your implementation. It's good to document why rather than how.
3) Typo in spelling of 'arguments'
4) It would be good to use a recognised docstring format, whether it's reST or Google-esque. See https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format. (I prefer Google-esque... 🙂 ).

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,33 @@ def get_page_from_request(request):
...
+
+    # stores all args from url path
+    all_request_filters = MultiDict(request.args.copy())
+
+    # loop to remove filters
+    for section in filters:
+        for _filter in filters[section]['filters']:
+            all_request_filters.poplist(_filter['name'])
+
+    # build url path with query string
```
samuelhwilliams: I don't think these three inline comments add a lot of value; they're describing what you're doing rather than why you're doing it. And we can see what you're doing. These comments could probably be removed.

app/main/views/g_cloud.py
```diff
@@ -222,6 +222,9 @@ def search_services():
             if 'label' in filter_instance:
                 filter_instance['label'] = capitalize_first(filter_instance['label'])
 
+    # Clears filters from URL to use for "Clear All functionality in the front end"
```
samuelhwilliams: Again this comment doesn't add much that we can't deduce from the code itself.

app/main/views/g_cloud.py
```diff
@@ -236,7 +239,8 @@ def search_services():
         summary=search_summary.markup(),
         title='Search results',
         total=search_results_obj.total,
-        gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'))
+        gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'),
+        clearFiltersUrl=clear_filters_url)
```
samuelhwilliams: Let's stick to the precedent here and call this variable `clear_filters_url` on both sides.

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,33 @@ def get_page_from_request(request):
...
+    '''
+
+    # stores all args from url path
+    all_request_filters = MultiDict(request.args.copy())
+
+    # loop to remove filters
+    for section in filters:
+        for _filter in filters[section]['filters']:
+            all_request_filters.poplist(_filter['name'])
+
```
samuelhwilliams: We will also need to remove `page` from the query params, if present.

DilwoarH: what is the `page` param used for? Pagination?

samuelhwilliams: Yep.

DilwoarH:
@samuelhwilliams - comments addressed.

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,32 @@ def get_page_from_request(request):
...
 
 
+def get_request_url_without_any_filters(request, filters):
+    """
+    This function will returns the url path without any filters.
+    It will still retain the categories, keyword and lots parameters as well as any others included.
+
+    Args:
+        request: Request Object (import from flask) of the current request.
+        filters: list of all filters from digital market place framework
```
samuelhwilliams: `marketplace` is one word in our context (don't need to fix this, just FYI)

samuelhwilliams: `marketplace` is just one word. FYI only; don't bother fixing for this PR.

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,32 @@ def get_page_from_request(request):
...
+
+    Returns:
+        URL path in string format.
+
```
samuelhwilliams: Nicer to not have a blank newline at the end of the docstring. Again FYI and future reference only; don't bother fixing for this.

app/main/helpers/search_helpers.py
```diff
@@ -20,6 +22,32 @@ def get_page_from_request(request):
...
+    """
+
+    all_request_filters = MultiDict(request.args.copy())
+
+    for section in filters:
+        for _filter in filters[section]['filters']:
+            all_request_filters.poplist(_filter['name'])
+
+    search_link_builder = Href(url_for('.search_services'))
+    url = search_link_builder(all_request_filters)
```
samuelhwilliams: I think we still need to strip out the `page` query parameter.

app/templates/search/_filters.html
```diff
@@ -1,13 +1,18 @@
...
-      with
-      label = filter.label,
-      options = filter.filters
-    %}
-      {% include "toolkit/forms/option-select.html" %}
-    {% endwith %}
-  {% endfor %}
+  <div class="dm-filter-title">
+    <h2 class="apply-filters-title">Apply filters</h2>
+    <a id="clearAllFilters" class="clear-filters-link" href="{{ clear_filters_url }}">Clear filters</a>
```
samuelhwilliams: This ID still needs updating to `clear-all-filters` or `dm-clear-all-filters` (and respectively in the frontend-toolkit PR).

PR #569, opened 2017-07-27
Show application stats on opportunity page

CrystalPea:
The aim is to show application statistics on individual brief/opportunity page. 

To do that we have added started and completed brief resposes to the view, changed the template and pull in a macro from frontend-toolkit.

Trello card: https://trello.com/c/9qlTFjK2/764-2-show-application-stats-on-opportunity-page


app/main/views/marketplace.py
```diff
@@ -75,14 +75,22 @@ def terms_and_conditions():
 def get_brief_by_id(framework_framework, brief_id):
     briefs = data_api_client.get_brief(brief_id)
     brief = briefs.get('briefs')
+    brief_responses = (data_api_client.find_brief_responses(brief_id=brief_id)).get("brief_responses")
+    if brief_responses is None:
+        brief_responses = []
+    started_brief_responses_len = len([response for response in brief_responses if "submitted_at" not in response])
+    completed_brief_responses_len = len([response for response in brief_responses if "submitted_at" in response])
+
```
MuriloDalRi: Could we have just one for loop here and increment the counters?

CrystalPea: The thing is the numbers should not be cumulative as per story ticket. So if there are 10 applications in total, of which 4 are completed, it will show: 6 started, 4 completed.

samuelhwilliams: It's not overly going to impact performance, but yeah, we could just do the loop once and then deduce `completed_brief_responses_len = len(brief_responses) - started_brief_responses_len`

tests/main/views/test_marketplace.py
```diff
@@ -551,10 +586,6 @@ def test_supplier_start_application(self):
 
     def test_supplier_applied_view_application_for_live_opportunity(self):
         self.login_as_supplier()
-        # mocking that we have applied
```
idavidmcdonald: These tests should include some set up that sets the supplier who is logged in as a supplier who has applied successfully (i.e. a submitted brief response)

I think these might need adding back in and changing so that they include a response for the supplier ID e.g.
`self._data_api_client.find_brief_responses.return_value = {
            "briefResponses": [{"supplierId": supplier id goes here}],
}
`

I think this is due to incorrect use of `supplier_id` rather than `supplierId` in your view which is searching for one response for a supplier.

tests/fixtures/dos_brief_responses_fixture.json
```diff
@@ -0,0 +1,65 @@
+{
+  "brief_responses": [
+    {
+      "id": 14275,
+      "data": {},
+      "brief_id": 1,
+      "supplier_id": 1234,
+      "created_at": "2016-12-02T11:09:28.054129Z",
```
CrystalPea: Good call, it would't work in real conditions as it was 😆 .

tests/fixtures/dos_brief_responses_fixture.json
```diff
@@ -0,0 +1,65 @@
...
+      "brief_id": 1,
+      "supplier_id": 706033,
+      "created_at": "2016-12-02T11:09:28.054129Z",
+      "submitted_at": "2016-12-05T11:09:28.054129Z"
+    },
+    {
+      "id": 14279,
+      "data": {},
+      "brief_id": 1,
+      "supplier_id": 706033,
```
CrystalPea: Good call again :)

tests/main/views/test_marketplace.py
```diff
@@ -373,6 +375,39 @@ def test_dos_brief_has_correct_title(self):
...
+
+        document = html.fromstring(res.get_data(as_text=True))
+        responses = self.brief_responses["brief_responses"]
+        started_responses_len = str(len([response for response in responses if "submitted_at" not in response]))
+        completed_responses_len = str(len([response for response in responses if "submitted_at" in response]))
+
+        started_responses_section = document.xpath('//div[@id="started-applications"]')[0]
+        completed_responses_section = document.xpath('//div[@id="completed-applications"]')[0]
+
+        assert started_responses_section.xpath('h2[@class="big-statistic"]/text()')[0] == started_responses_len
```
CrystalPea: Ahh, yes, good find, thank you! :)

tests/main/views/test_marketplace.py
```diff
@@ -373,6 +375,39 @@ def test_dos_brief_has_correct_title(self):
...
+        assert completed_responses_section.xpath('h2[@class="big-statistic"]/text()')[0] == completed_responses_len
+        assert started_responses_section.xpath('h3[@class="statistic-name"]/text()')[0] == "started applications"
+        assert completed_responses_section.xpath('h3[@class="statistic-name"]/text()')[0] == "completed applications"
```
CrystalPea: 👍

tests/main/views/test_marketplace.py
```diff
@@ -540,7 +575,7 @@ def test_supplier_start_application(self):
         self.login_as_supplier()
         # mocking that we haven't applied
         self._data_api_client.find_brief_responses.return_value = {
-            "briefResponses": [],
+            "brief_responses": [],
```
CrystalPea: you're right, changed it back now

tests/main/views/test_marketplace.py
CrystalPea: I set the supplierId in one of the briefResponses to the supplierId from our fixture, this is why the test passes. I don't think we need to overrun return_value set up at the top of the file for this one then, right?

app/main/views/marketplace.py
```diff
@@ -75,14 +75,22 @@ def terms_and_conditions():
 def get_brief_by_id(framework_framework, brief_id):
     briefs = data_api_client.get_brief(brief_id)
     brief = briefs.get('briefs')
+    brief_responses = data_api_client.find_brief_responses(brief_id=brief_id).get("briefResponses")
+    if brief_responses is None:
```
benvand: Don't need to assign to empty list here api does it for us

app/main/views/marketplace.py
```diff
@@ -75,14 +75,22 @@ def terms_and_conditions():
 def get_brief_by_id(framework_framework, brief_id):
     briefs = data_api_client.get_brief(brief_id)
     brief = briefs.get('briefs')
+    brief_responses = data_api_client.find_brief_responses(brief_id=brief_id).get("briefResponses")
```
benvand: add status kwarg to apiclient call here :arrow:
`client.find_brief_responses(brief_id = 5594, status='submitted,draft')`

app/main/views/marketplace.py
```diff
@@ -75,14 +75,22 @@ def terms_and_conditions():
...
+    if brief_responses is None:
+        brief_responses = []
+    started_brief_responses_count = len([response for response in brief_responses if "submittedAt" not in response])
+    completed_brief_responses_count = len([response for response in brief_responses if "submittedAt" in response])
```
benvand: Let's use status here so we don't get confused when changing statuses

app/main/views/marketplace.py
```diff
@@ -75,14 +75,23 @@ def terms_and_conditions():
...
         abort(404, "Opportunity '{}' can not be found".format(brief_id))
 
     if getattr(current_user, "supplier_id", None) is None:
         # user unauthenticated or not a supplier
-        brief_responses = None
+        supplier_brief_response = None
     else:
-        brief_responses = data_api_client.find_brief_responses(brief_id, current_user.supplier_id)["briefResponses"]
+        supplier_brief_response = [
+            response for response in brief_responses if response["supplierId"] == current_user.supplier_id
```
samuelhwilliams: I can't see where this brief response itself is ever used; `supplier_brief_response` only seems to be used as a boolean. We should probably turn it into one and call it something like `supplier_has_responded_to_brief` so that it's clearer what this is being used for. I know you've not written this bit from scratch, but since you're touching it we might as well reduce it to the simplest case.

tests/main/views/test_marketplace.py
```diff
@@ -373,6 +375,40 @@ def test_dos_brief_has_correct_title(self):
...
 
+    def test_dos_brief_shows_application_stats(self):
+        brief_id = self.brief['briefs']['id']
+        res = self.client.get('/digital-outcomes-and-specialists/opportunities/{}'.format(brief_id))
+        assert res.status_code == 200
+
+        document = html.fromstring(res.get_data(as_text=True))
+        responses = self.brief_responses["briefResponses"]
+        started_responses_len = str(len([response for response in responses if "submittedAt" not in response]))
+        completed_responses_len = str(len([response for response in responses if "submittedAt" in response]))
```
idavidmcdonald: We no longer need these two lines

tests/main/views/test_marketplace.py
idavidmcdonald: Ah OK. Makes sense. Might be worth a comment in here as to realise the data is set up like this you need to go into a different file and spot one ID in there matches the ID of the supplier you are logged in as (which you might not remember off the top of your head)

CrystalPea: You mean like a comment in test file?

app/main/views/marketplace.py
CrystalPea: Good idea, I will do that

tests/main/views/test_marketplace.py
idavidmcdonald: yep something like `# this supplier has applied for this brief as part of our brief responses fixture in the setup`

tests/main/views/test_marketplace.py
CrystalPea: You are right I will delete them

app/main/views/marketplace.py
CrystalPea: Done now

tests/main/views/test_marketplace.py
CrystalPea: ok, I added the comment now ☺️

app/main/views/marketplace.py
```diff
@@ -75,14 +75,25 @@ def terms_and_conditions():
...
     if brief['status'] not in ['live', 'closed', 'withdrawn'] or brief['frameworkFramework'] != framework_framework:
         abort(404, "Opportunity '{}' can not be found".format(brief_id))
 
-    if getattr(current_user, "supplier_id", None) is None:
-        # user unauthenticated or not a supplier
-        brief_responses = None
-    else:
-        brief_responses = data_api_client.find_brief_responses(brief_id, current_user.supplier_id)["briefResponses"]
+    has_supplier_responded_to_brief = False
+    if brief_responses and getattr(
```
benvand: Because you're only using the boolean in the if statement this can be a `hasattr` call rather than a `getattr` call.

You could also use a try, except block to follow the beg forgiveness not permission paradigm like so:
```

try:
    has_supplier_responded_to_brief = current_user.supplier_id in [i['supplierId'] for i in brief_responses]
except AttributeError:
    has_supplier_responded_to_brief = False

File: app/main/views/marketplace.py [hunk 14]
```diff
@@ -75,14 +75,25 @@ def terms_and_conditions():
...
+        current_user,
+        "supplier_id",
+        None
+    ) and current_user.supplier_id in brief_responses[0].values():
```
benvand: `brief_responses` is at this point a `list` of `dict` objects like so:
```
[
    {'supplierId': 1, 'briefId': 123, 'id': 1234 ...}, 
    {'supplierId': 2, 'briefId': 123, 'id': 1235 ...}, 
    {'supplierId': 3, 'briefId': 123, 'id': 1236 ...}
]``` 
That means that the check:
`current_user.supplier_id in brief_responses[0].values()`
will check if the supplier id is in the values of the first element of the list, ie:
`[1, 123, 1234]`
But we want the _all_ the supplier ids:
`[1, 2, 3]`

app/main/views/marketplace.py
benvand: https://stackoverflow.com/questions/12265451/ask-forgiveness-not-permission-explain

CrystalPea: thanks, this has been done now :)

app/main/views/marketplace.py
CrystalPea: ahh, you're right!

PR #570, opened 2017-07-28
Fix bad flash message

galund:
Test to see if the flash message apparently intended only for google analytics is appearing in the user interface.

https://trello.com/c/4JthSqdx/614-flash-message-problem-on-buyer-account-creation

Probably only makes any sense in the context of https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/364

### DIFF
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -4,6 +4,7 @@ import mock
 from six import iteritems
 from six.moves.urllib.parse import urlparse, parse_qs
 from lxml import html
+import re
 from ...helpers import BaseApplicationTest
 import pytest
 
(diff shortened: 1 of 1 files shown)


tests/main/views/test_marketplace.py
```diff
@@ -34,6 +35,10 @@ def test_data_analytics_track_page_view_is_shown_if_account_created_flag_flash_m
         data = res.get_data(as_text=True)
 
         assert 'data-analytics="trackPageView" data-url="buyers?account-created=true"' in data
+        # however this should not be shown as a regular flash message
+        flash_banner_match = re.search('<p class="banner-message">\s*(.*)', data, re.MULTILINE)
```
katstevens: Is there a particular reason we're using `re` rather than `xpath` to look for the element? In the briefs frontend we've been using the latter but I have no idea which is more consistent!

galund: No particular reason, I suppose there's an increasing order of complexity from 

 - just checking if some text is in the response (i.e. what the assert just before the line I added does)
 - using a regex to check if some text is in there, and extract some nearby text too (i.e. this thing)
 - using xpath to build a full DOM and do stuff with that.

They are all fine (and maybe my assumption that regex's are cheaper than xpath is wrong), but choosing the least complex option in each case seems okay. (Although consistency is good too, so in a test that already used all xpath, I'd probably have gone with that.)

katstevens: Cool - if it's cheaper and the regex pattern itself is readable (which this one is!) then I'm happy 😸

risicle: Yeah I'd generally prefer the xpath... path, simply because it's more robust, but I don't see it as particularly important in this case.

PR #573, opened 2017-08-01
Restore legacy brief start/ continue functionality

benvand:
Fixes this functional test failure
https://ci.marketplace.team/job/functional-tests-preview/18877/functional_test_report
introduced in 
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/569

We went from considering that a supplier had applied to a brief if they had a _'submitted'_ brief response to considering them applied if they had a _'draft' or 'submitted'_ brief response.

This restored the previous functionality. We only consider a supplier as having applied if they have a _'submitted'_ brief response.

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -80,15 +80,16 @@ def get_brief_by_id(framework_framework, brief_id):
         status='draft,submitted'
     ).get('briefResponses')
     started_brief_responses_count = len([response for response in brief_responses if response['status'] == 'draft'])
-    completed_brief_responses_count = len(
-        [response for response in brief_responses if response['status'] == 'submitted']
-    )
+    completed_brief_responses = [response for response in brief_responses if response['status'] == 'submitted']
+    completed_brief_responses_count = len(completed_brief_responses)
 
     if brief['status'] not in ['live', 'closed', 'withdrawn'] or brief['frameworkFramework'] != framework_framework:
         abort(404, "Opportunity '{}' can not be found".format(brief_id))
 
     try:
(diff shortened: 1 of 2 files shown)


app/main/views/marketplace.py
```diff
@@ -80,15 +80,16 @@ def get_brief_by_id(framework_framework, brief_id):
...
+    completed_brief_responses = [response for response in brief_responses if response['status'] == 'submitted']
+    completed_brief_responses_count = len(completed_brief_responses)
 
     if brief['status'] not in ['live', 'closed', 'withdrawn'] or brief['frameworkFramework'] != framework_framework:
         abort(404, "Opportunity '{}' can not be found".format(brief_id))
 
     try:
-        has_supplier_responded_to_brief = current_user.supplier_id in [i['supplierId'] for i in brief_responses]
+        has_supplier_responded_to_brief = (
+            current_user.supplier_id in [br['supplierId'] for br in completed_brief_responses]
```
Wynndow: Would using `brief` be marginally more readable than `br`? Tiny point, no objections if you don't agree.

benvand: I do, especially considering they're responses and we're using response elsewhere. Good call, fixed

katstevens: Isn't it a `brief_response`, not a `brief`? 😄 How about `cbr` for `completed_brief_response`?

Wynndow: Oh yeah, it totally is a `brief_response`!

benvand:
Thanks all! Merging

PR #576, opened 2017-08-02
Add respondent supplier size to statistic on briefs page

benvand:
* Pass numbers into template from view
* Format string in template
* Update toolkit to bring in new version supporting descripotion text
* Update tests and add new assertions

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -79,9 +79,27 @@ def get_brief_by_id(framework_framework, brief_id):
         brief_id=brief_id,
         status='draft,submitted'
     ).get('briefResponses')
-    started_brief_responses_count = len([response for response in brief_responses if response['status'] == 'draft'])
+
+    started_brief_responses = [response for response in brief_responses if response['status'] == 'draft']
     completed_brief_responses = [response for response in brief_responses if response['status'] == 'submitted']
-    completed_brief_responses_count = len(completed_brief_responses)
+
+    # Counts for application statistics
+    started_sme_responses_count = len([
+        response for response in started_brief_responses
+        if response['supplierOrganisationSize'] in ['micro', 'small', 'medium']
+    ])
+    started_large_responses_count = len([
+        response for response in started_brief_responses
+        if response['supplierOrganisationSize'] == 'large'
+    ])
+    completed_sme_responses_count = len([
+        response for response in completed_brief_responses
+        if response['supplierOrganisationSize'] in ['micro', 'small', 'medium']
+    ])
+    completed_large_responses_count = len([
+        response for response in completed_brief_responses
+        if response['supplierOrganisationSize'] == 'large'
+    ])
 
     if brief['status'] not in ['live', 'closed', 'withdrawn'] or brief['frameworkFramework'] != framework_framework:
         abort(404, "Opportunity '{}' can not be found".format(brief_id))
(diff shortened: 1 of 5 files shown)


app/templates/_brief_applications_stats.html
```diff
@@ -1,4 +1,22 @@
 {% import "toolkit/statistics.html" as statistics %}
 
-{{ statistics.big_stat(started_brief_responses_count, "Started applications" if started_brief_responses_count != 1 else "Started application", id="started-applications") }}
-{{ statistics.big_stat(completed_brief_responses_count, "Completed applications" if completed_brief_responses_count != 1 else "Completed application", id="completed-applications") }}
+{{ statistics.big_stat(
+     started_brief_responses_stats["started_brief_responses_total"],
+     "Started applications" if started_brief_responses_stats["started_brief_responses_total"] != 1 else "Started application",
```
Wynndow: Might it be a bit cleaner to move this logic out of the template and into the view? You could pass it in as kwarg to render template?

samuelhwilliams: We also have a Jinja filter that wraps this logic called `pluralize`. Used like `pluralize(started_brief_responses_stats["started_brief_responses_total"], "Started application", "Started applications")` I think.

app/main/views/marketplace.py
```diff
@@ -103,8 +121,18 @@ def get_brief_by_id(framework_framework, brief_id):
     return render_template(
         'brief.html',
         brief=brief,
-        started_brief_responses_count=started_brief_responses_count,
-        completed_brief_responses_count=completed_brief_responses_count,
+        started_brief_responses_stats={
+            'started_sme_brief_responses_count': started_sme_brief_responses_count,
+            'started_large_brief_responses_count': started_large_brief_responses_count,
+            'started_brief_responses_total': started_sme_brief_responses_count + started_large_brief_responses_count
```
samuelhwilliams: I think it would be reasonable to just name the keys `sme`, `large`, and `total`. It's apparent that they're started brief response counts because of the name of the dict they're stored in, so repeating all of that detail feels a bit redundant and makes it harder to parse.

app/templates/_brief_applications_stats.html
```diff
@@ -1,4 +1,28 @@
 {% import "toolkit/statistics.html" as statistics %}
 
-{{ statistics.big_stat(started_brief_responses_count, "Started applications" if started_brief_responses_count != 1 else "Started application", id="started-applications") }}
-{{ statistics.big_stat(completed_brief_responses_count, "Completed applications" if completed_brief_responses_count != 1 else "Completed application", id="completed-applications") }}
+{{ statistics.big_stat(
+     started_brief_responses_stats["started_brief_responses_total"],
```
samuelhwilliams: As above, if this read `started_brief_responses_stats['total']` this would read better IMO.

idavidmcdonald: I second this

app/main/views/marketplace.py
benvand: @idavidmcdonald @Wynndow @samuelhwilliams 

Totally agree with your comments about the variable naming. I've changed them to be
`'sme_count', 'large_count', 'total'`
I made this decision based on the fact that they're already in a container that denotes that they are statistics about brief responses with a certain status `'completed_responses_stats', 'started_responses_stats'` so all that needs to be revealed by their key names is what differentiates them. That is that one is the count of smes, one is the count of large and one is the total.

I also changed the logic in the template to use the total for the if statement to decide whether to show the stats at all.
It was:
```
show_statistics if started_responses_stats['sme_count'] + started_responses_stats['large_count'] else do_not_show_statistics
```
It's now
```
show_statistics if started_responses_stats['total'] else do_not_show_statistics
```

benvand:
@samuelhwilliams searched for 'pluralise' and didn't find it 'cause I spell pluralise the proper way ;) Have dropped that in

benvand:
@Wynndow Decided to keep the string formatting logic in the template but cleaned it up substantially so it's much more readable. I like keeping content as close to the template as possible but you're right it was a bit messy to look at.

benvand:
@idavidmcdonald Updated 0 responses case and 1 response case tests to accurately demonstrate what we expect in these situations.

PR #581, opened 2017-08-15
Skeleton direct award pages

samuelhwilliams:
## Summary
Skeleton direct award pages/alterations to allow testing the flow. This skeleton has not been designed, but implements the basic steps and pages required for functionality. Following stories in Compare and Decide will re-design the pages heavily.

At a high level, this flow will let you run a search against G-Cloud, save those search parameters into a new Direct Award Procurement Project (name still in the air), and view an overview page for that project that links you back to the saved search. The new flow has been implemented under a 'direct_award' Blueprint which will allow us to move this into another app if we need to do so.

This flow will be hidden behind a Feature Flag in staging and production, but be available for use in preview.

**Please don't review the UI itself, which is out of scope for this PR, but do review how the code supporting them has been implemented (i.e. evertything else)**

## Complementary PRs
~~https://github.com/alphagov/digitalmarketplace-api/pull/636~~
https://github.com/alphagov/digitalmarketplace-functional-tests/pull/321
~~https://github.com/alphagov/digitalmarketplace-apiclient/pull/88~~

## Example
procurement project smaller

## Ticket
https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -42,11 +42,13 @@ def create_app(config_name):
                 content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
 
     from .main import main as main_blueprint
+    from .main import direct_award as direct_award_blueprint
     from .external.views.external import external as external_blueprint
     from .status import status as status_blueprint
 
     application.register_blueprint(status_blueprint)
     application.register_blueprint(main_blueprint)
+    application.register_blueprint(direct_award_blueprint)
     application.register_blueprint(external_blueprint)
 
     login_manager.login_view = 'main.render_login'
--- a/app/main/__init__.py
+++ b/app/main/__init__.py
@@ -1,6 +1,20 @@
-from flask import Blueprint
+from flask import Blueprint, flash, Markup, current_app
+from flask_login import login_required, current_user
+import flask_featureflags
 
 main = Blueprint('main', __name__)
+direct_award = Blueprint('direct_award', __name__, url_prefix='/buyers/direct-award')
+
+LOGIN_REQUIRED_MESSAGE = Markup("""You must log in with a buyer account to see this page.""")
+
+
+@direct_award.before_request
+@flask_featureflags.is_active_feature('DIRECT_AWARD_PROJECTS')
+@login_required
+def require_login():
+    if current_user.is_authenticated() and current_user.role != 'buyer':
+        flash(LOGIN_REQUIRED_MESSAGE, 'error')
+        return current_app.login_manager.unauthorized()
 
 from . import errors
 from .views import (
--- a/app/main/helpers/direct_award_helpers.py
+++ b/app/main/helpers/direct_award_helpers.py
@@ -1,0 +1,2 @@
+def is_direct_award_project_accessible(project, user_id):
+    return any([user['id'] == user_id for user in project['users']])
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -170,8 +170,7 @@ def get_filter_value_from_question_option(option):
     return (option.get('value') or option.get('label', '')).lower().replace(',', '')
 
 
-def build_search_query(framework, request_args, lot_filters, content_builder, lots_by_slug,
-                       for_aggregation=False):
+def build_search_query(request_args, lot_filters, content_builder, lots_by_slug, for_aggregation=False):
     """Match request args with known filters.
 
     Removes any unknown query parameters, and will only keep `page`, `q`
@@ -188,13 +187,7 @@ def build_search_query(framework, request_args, lot_filters, content_builder, lo
         for_aggregation=for_aggregation
     )
 
-    if 'q' in query:
-        query['q'] = replace_g5_search_dots(query['q'])
-
-    params = group_request_filters(query, content_builder)
-    params.update({'index': framework['slug']})
-
-    return params
+    return group_request_filters(query, content_builder)
 
 
 def query_args_for_pagination(args):
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -138,6 +138,25 @@ def group_request_filters(request_filters, content_builder):
     return filter_query
 
 
+def ungroup_request_filters(request_filters, content_builder):
+    """Carries out the inverse of the above method which is required when going from search api url parameters
+    to frontend url parameters. Where group_request_filters takes request_filters as a MultiDict, this method
+    takes request_filters as a tuple of tuple pairs ((param_name, param_value), ...)"""
+    filter_query = []
+
+    for key, value in request_filters:
+        if is_radio_type(content_builder, key):
+            if ',' in value:
+                for value in value.split(','):
+                    filter_query.append((key, value))
+            else:
+                filter_query.append((key, value))
+        else:
+            filter_query.append((key, value))
+
+    return tuple(filter_query)
+
+
 def is_radio_type(content_builer, key):
     if key == 'lot':
         return True
--- a/app/main/helpers/shared_helpers.py
+++ b/app/main/helpers/shared_helpers.py
@@ -1,7 +1,4 @@
-try:
-    from urlparse import urlparse, parse_qs
-except ImportError:
-    from urllib.parse import urlparse, parse_qs
+from urllib.parse import urlparse, parse_qs, urlencode, urlunparse
 
 
 def parse_link(links, label):
@@ -13,3 +10,20 @@ def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_o
         for framework in frameworks:
             if framework.get('status') == status:
                 return framework
+
+
+def construct_url_from_base_and_params(base_url, query_params):
+    """
+    Combine a base url (MUST NOT HAVE ANY QUERY PARAMS TO BE RETAINED) with supplied query params
+    :param base_url: A url e.g. http://localhost
+    :param query_params: A tuple of key/value tuples ((param_name, param_val), ...)
+    :return: A url with the query params inline e.g. http://localhost?param_name=param_val
+    """
+    # urlparse breaks a URL into six items (scheme, netloc, path, params, query, fragment)
+    parsed_url = list(urlparse(base_url))
+
+    # We want to use all existing segments except the current query params (5th item), which we'll overwrite.
+    parsed_url[4] = urlencode(query_params)
+
+    # Reconstruct the full URL with inline params from our six-item iterable
+    return urlunparse(parsed_url)
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
 from flask_login import current_user
 import flask_featureflags
-from urllib.parse import urlencode, urlunparse, urlparse
 from werkzeug.datastructures import MultiDict
 
 from dmutils.formats import dateformat
@@ -24,16 +23,18 @@ from ..helpers.search_helpers import (
     get_keywords_from_request, pagination,
     get_page_from_request, query_args_for_pagination,
     get_lot_from_request, build_search_query,
-    clean_request_args, get_request_url_without_any_filters
+    clean_request_args, get_request_url_without_any_filters,
+    ungroup_request_filters
 )
 from ..helpers import framework_helpers
 from ..helpers.direct_award_helpers import is_direct_award_project_accessible
+from ..helpers.shared_helpers import construct_url_from_base_and_params
 
 from ..exceptions import AuthException
 from app import search_api_client, data_api_client, content_loader
 
 
-PROJECT_CREATED_MESSAGE = Markup("""Your new procurement project has been created.""")
+PROJECT_CREATED_MESSAGE = Markup("""Your new project has been created.""")
 
 
 @main.route('/g-cloud')
@@ -232,7 +233,6 @@ def search_services():
                 filter_instance['label'] = capitalize_first(filter_instance['label'])
 
     clear_filters_url = get_request_url_without_any_filters(request, filters)
-    show_save_button = flask_featureflags.is_active('DIRECT_AWARD_PROJECTS')
 
     template_args = dict(
         current_lot=current_lot,
@@ -249,8 +249,7 @@ def search_services():
         title='Search results',
         total=search_results_obj.total,
         gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'),
-        clear_filters_url=clear_filters_url,
-        show_save_button=show_save_button)
+        clear_filters_url=clear_filters_url)
 
     if request.args.get('live-results'):
         from flask import jsonify
@@ -278,11 +277,11 @@ def search_services():
     )
 
 
-@direct_award.route('/save-search', methods=['GET'])
-def save_search():
+@direct_award.route('/<string:framework_framework>/save-search', methods=['GET'])
+def save_search(framework_framework):
     # Get core data
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
-    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
     lots_by_slug = framework_helpers.get_lots_by_slug(framework)
     current_lot_slug = get_lot_from_request(request, lots_by_slug)
 
@@ -308,13 +307,14 @@ def save_search():
 
     return render_template(
         'direct-award/save-search.html',
+        framework_framework=framework_framework,
         search_summary=search_summary,
         search_api_url=search_api_url,
     )
 
 
-@direct_award.route('/projects/create', methods=['POST'])
-def project_create():
+@direct_award.route('/<string:framework_framework>/projects/create', methods=['POST'])
+def project_create(framework_framework):
     try:
         api_project = data_api_client.create_direct_award_project(user_id=current_user.id,
                                                                   user_email=current_user.email_address,
@@ -337,15 +337,16 @@ def project_create():
     flash(PROJECT_CREATED_MESSAGE, 'success')
 
     return redirect(url_for('.view_project',
+                            framework_framework=framework_framework,
                             project_id=project['id']
                             ))
 
 
-@direct_award.route('/projects/<int:project_id>', methods=['GET'])
-def view_project(project_id):
+@direct_award.route('/<string:framework_framework>/projects/<int:project_id>', methods=['GET'])
+def view_project(framework_framework, project_id):
     # Get core data
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
-    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
     content_manifest = content_loader.get_manifest(framework['slug'], 'search_filters')
     lots_by_slug = framework_helpers.get_lots_by_slug(framework)
 
@@ -362,6 +363,7 @@ def view_project(project_id):
 
     # We need to get buyer-frontend query params from our saved search API URL.
     search_query_params = search_api_client.get_frontend_params_from_search_api_url(search['searchUrl'])
+    search_query_params = ungroup_request_filters(search_query_params, content_manifest)
     search_query_params_multidict = MultiDict(search_query_params)
 
     current_lot_slug = search_query_params_multidict.get('lot', None)
@@ -369,10 +371,7 @@ def view_project(project_id):
     clean_request_query_params = clean_request_args(search_query_params_multidict, filters.values(), lots_by_slug)
 
     # Now build the buyer-frontend URL representing the saved Search API URL
-    search_page_base_url = url_for('main.search_services')
-    parsed_url = list(urlparse(search_page_base_url))
-    parsed_url[4] = urlencode(search_query_params)
-    search_page_full_url = urlunparse(parsed_url)
+    buyer_search_page_url = construct_url_from_base_and_params(url_for('main.search_services'), search_query_params)
 
     # Get the saved Search API URL result set and build the search summary.
     search_api_response = search_api_client._get(search['searchUrl'])
@@ -385,6 +384,6 @@ def view_project(project_id):
 
     return render_template('direct-award/view-project.html',
                            project_name=project['name'],
-                           search_page_url=search_page_full_url,
+                           search_page_url=buyer_search_page_url,
                            search_created_at=search['createdAt'],
                            search_summary=search_summary.markup())
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 
 from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
 from flask_login import current_user
-import flask_featureflags
 from werkzeug.datastructures import MultiDict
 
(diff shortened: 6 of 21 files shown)


app/__init__.py
```diff
@@ -42,11 +42,13 @@ def create_app(config_name):
                 content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
 
     from .main import main as main_blueprint
+    from .main import projects as projects_blueprint
```
galund: can we call it direct_award_projects, so this would be `from .main import direct_award_projects as projects_blueprint`

app/main/__init__.py
```diff
@@ -1,6 +1,20 @@
-from flask import Blueprint
+from flask import Blueprint, flash, Markup, current_app
+from flask_login import login_required, current_user
+import flask_featureflags
 
 main = Blueprint('main', __name__)
+direct_award = Blueprint('direct_award', __name__, url_prefix='/buyers/direct-award')
+
+LOGIN_REQUIRED_MESSAGE = Markup("""You must be logged in to view this page.""")
```
galund: The equivalent message in the briefs app is "You must log in with a buyer account to see this page." and it would probably be good if these matched. Ideally they'd be shared - so I think we should potentially move `require_login` login into dm-utils. Somehow. No idea how we can share these `before_request` things between apps, but it seems like a valid need (cc @allait).

app/main/presenters/search_summary.py
```diff
@@ -126,7 +126,7 @@ def _sort_groups(groups):
 
         groups = defaultdict(list)
         for filter_name, filter_values in request_args.lists():
-            if filter_name in ('lot', 'q', 'page'):
+            if filter_name in ('lot', 'q', 'page', 'serviceCategories'):
```
galund: Why did this suddenly need to change?

app/main/views/g_cloud.py
```diff
@@ -179,6 +186,7 @@ def search_services():
         abort(404)
 
     search_api_response = search_api_client.search_services(
+        index=framework['framework'],  # TODO: change this to framework['slug'], also search_presenters.py:~132
```
galund: We're ready to do this now I think? If not let's make a plan (he said going on hol)

app/main/views/g_cloud.py
```diff
@@ -265,3 +275,143 @@ def search_services():
...
+
+@direct_award.route('/save-search', methods=['GET'])
+def save_search():
+    # if there are multiple live g-cloud frameworks, we must assume the same filters work on them all
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+    lots_by_slug = framework_helpers.get_lots_by_slug(framework)
+    current_lot_slug = get_lot_from_request(request, lots_by_slug)
+
+    # the bulk of the possible filter parameters are defined through the content loader. they're all boolean and the
```
galund: We shouldn't repeat the comment from the other file. If the code can be shared that'd be best (and the comment moved to the shared place); otherwise just say "see `module.function`"

app/main/views/g_cloud.py
```diff
@@ -265,3 +275,143 @@ def search_services():
...
+@direct_award.route('/projects/create', methods=['POST'])
+def project_create():
+    try:
+        api_project = data_api_client.create_direct_award_project(user_id=current_user.id,
+                                                                  user_email=current_user.email_address,
+                                                                  project_name=request.form.get('project_name',
+                                                                                                'My new project'))
+
+    except HTTPError as e:
+        abort(e.status_code)
```
galund: We should probably pass on the message as well as the code in these cases?

config.py
```diff
@@ -64,6 +64,7 @@ class Config(object):
     # Feature Flags
     RAISE_ERROR_ON_MISSING_FEATURES = True
     FEATURE_FLAGS_NEW_SUPPLIER_FLOW = False
+    FEATURE_FLAGS_DIRECT_AWARD = False
```
galund: Would DIRECT_AWARD_PROJECTS be more clear? We can't enabling/disabling G-Cloud after all :)

app/main/views/g_cloud.py
samuelhwilliams: Yep I think we are.

app/main/views/g_cloud.py
samuelhwilliams: We generally haven't been doing this in other places in the app.

app/main/__init__.py
samuelhwilliams: Let's add this as tech debt because it will need to be centralised and applied across repositories.

tests/main/helpers/test_search_helpers.py
```diff
@@ -247,42 +246,44 @@ def test_build_search_query_unknown_lot_is_dropped(self):
...
 
-        assert search_helpers.build_search_query(self.g6_framework, request.args, self.lot_filters, self._loader(),
+        assert search_helpers.build_search_query(request.args, self.lot_filters, self._loader(),
                                                  self._lots_by_slug) == {
-            'index': 'g-cloud-6',
             'q': 'some text 5-G4-1005-001',
         }
+
+
+def find_search_summary(res_data):
```
katstevens: Should this helper be in this module? Or in `tests.helpers`?

app/main/presenters/search_summary.py
samuelhwilliams: Lots superfluous; removed.

app/main/views/g_cloud.py
samuelhwilliams: 👍

tests/main/helpers/test_search_helpers.py
samuelhwilliams: Ah yes, good spot. I think I thought it was search helpers to help with testing. Brain fail.

I'll move it into tests.helpers.BaseApplicationTest

tests/main/views/test_direct_award.py
```diff
@@ -0,0 +1,93 @@
...
+        assert res.status_code == 200
+
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert len(doc.xpath('//button[@id="save-search" and @formaction="/buyers/direct-award/save-search"]')) == 1
+
+    def test_save_search_redirects_to_login(self):
+        self._search_api_client.search_services.return_value = self.search_results_multiple_page
+
+        res = self.client.get('/buyers/direct-award/save-search?lot=cloud-software')
+        assert res.status_code == 302
```
katstevens: Can we assert that the location we're redirecting to is the login url? Also, why do we need the state the API return value in this test?

tests/main/views/test_direct_award.py
```diff
@@ -0,0 +1,93 @@
...
+        assert len(doc.xpath('//h1[contains(normalize-space(text()), "Project - {}")]'.format(project_name))) == 1
+
+    def test_view_project_shows_active_search_details(self):
+        self._search_api_client.deconstruct_url.return_value = (('q', 'accelerator'), )
+        self.login_as_buyer()
+
+        return_value = self.search_results_multiple_page
+        return_value["services"] = [return_value["services"][0]]
+        return_value["meta"]["total"] = 1
+        self._search_api_client.search_services.return_value = return_value
```
katstevens: Rather than duplicating the return_value setup here and in `test_save_search_renders_summary_on_page`, can we just do the setup in `setup()`? Alternatively, if the fixture doesn't quite reflect what's going to be returned, should we make a new one and load that as the response instead?

tests/main/views/test_direct_award.py
samuelhwilliams: Good idea to pin the redirect location. Seems we don't need the fixture here so I've removed it.

tests/main/views/test_direct_award.py
samuelhwilliams: Good idea. I've simplified these a little.

app/main/helpers/direct_award_helpers.py
```diff
@@ -0,0 +1,2 @@
+def is_direct_award_project_accessible(project, user_id):
+    return user_id in [user['id'] for user in project['users']]
```
risicle: `any(user["id"] == user_id for user in project['users'])`

samuelhwilliams: Thanks. 👍

TheDoubleK:
```
"We will switch to using the content-loadear[sic] and a new manifest for naming the project as part of a later ticket, to cover validation/etc."
```
I don't think this should be necessary at all - the content here isn't linked to a particular framework and you're not filling out a JSON data field, it's general app functionality.   You should be able to stick with a one-page WTForm for this no worries I'd have thought.

samuelhwilliams:
👍

app/main/helpers/search_helpers.py
```diff
@@ -191,10 +190,7 @@ def build_search_query(framework, request_args, lot_filters, content_builder, lo
     if 'q' in query:
         query['q'] = replace_g5_search_dots(query['q'])
```
TheDoubleK: We can completely remove this `replace_g5_search_dots` thing now, as it's been a long time since G5 services came back from our index, with or without dashes instead of dots.  Doesn't have to be for this PR, but it's definitely completely pointless these days.

samuelhwilliams: Cool, I'll rip it out.

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+    search_query_params_multidict = MultiDict(search_query_params)
+
+    current_lot_slug = search_query_params_multidict.get('lot', None)
+    filters = filters_for_lot(current_lot_slug, content_manifest, all_lots=framework['lots'])
+    clean_request_query_params = clean_request_args(search_query_params_multidict, filters.values(), lots_by_slug)
+
+    # Now build the buyer-frontend URL representing the saved Search API URL
+    search_page_base_url = url_for('main.search_services')
+    parsed_url = list(urlparse(search_page_base_url))
+    parsed_url[4] = urlencode(search_query_params)
```
TheDoubleK: Could do with a comment here to say what we're expecting to find in [0], [1], [2] and [3]. And might be worth extracting this URL construction into a helper method? It feels a bit dark-artsy and a helper would make this bit directly testable (and more obvious exactly what it's doing).

app/templates/direct-award/save-search.html
```diff
@@ -0,0 +1,52 @@
...
+{% block breadcrumb %}
+  {%
+    with
+    items = [
+      {
+          "link": url_for('main.index'),
+          "label": "Digital Marketplace"
+      },
+      {
+          "label": "Save your search"
```
TheDoubleK: I don't think we generally have an unlinked breadcrumb for the current page? (Although counterexamples to this may exist...)

app/templates/direct-award/save-search.html
```diff
@@ -0,0 +1,52 @@
...
+    </h1>
+  </header>
+  <p>{{ search_summary.markup() }}</p>
+  <div class="grid-row">
+    <form method="post" action="{{ url_for('direct_award.project_create') }}">
+      <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
+      <input type="hidden" name="search_api_url" value="{{ search_api_url }}"/>
+      {%
+        with
+        question = "Name a new procurement project",
```
TheDoubleK: ~procurement~ --> buying

app/templates/direct-award/view-project.html
```diff
@@ -0,0 +1,40 @@
+{% extends "_base_page.html" %}
+
+{% block page_title %}Save your search - Digital Marketplace{% endblock %}
```
TheDoubleK: "Saved search"?

tests/fixtures/direct_award_project_fixture.json
```diff
@@ -0,0 +1,18 @@
...
+  "project": {
+    "active": true,
+    "createdAt": "2017-08-29T07:49:26.677778Z",
+    "id": 1,
+    "lockedAt": null,
+    "name": "My procurement project",
+    "users": [
+      {
+        "active": true,
+        "email": "buyer@example.com@",
```
TheDoubleK: Is  the trailing `@` intentional here?

tests/helpers.py
```diff
@@ -237,6 +245,10 @@ def strip_all_whitespace(content):
         pattern = re.compile(r'\s+')
         return re.sub(pattern, '', content)
 
+    @staticmethod
+    def find_search_summary(res_data):
+        return re.findall(r'<span class="search-summary-count">.+</span>[^\n]+', res_data)
```
TheDoubleK: Might be nicer to read into a document and find with XPath rather than a regex.

tests/main/views/test_search.py
```diff
@@ -14,11 +14,6 @@ def find_pagination_links(res_data):
         re.MULTILINE)
 
 
-def find_search_summary(res_data):
```
TheDoubleK: Oh, you just moved an existing regex thing that's over two years old?  Probably just leave it as-is for now then.

tests/fixtures/direct_award_project_searches_fixture.json
```diff
@@ -0,0 +1,31 @@
+{
+  "searches": [
+    {
+      "active": true,
+      "createdAt": "2017-08-29T07:00:02.000000Z",
+      "createdBy": 10000,
+      "id": 1,
+      "projectId": 1,
+      "searchUrl": "http://search-api.digitalmarketplace.service.gov.uk/g-cloud/services/search?q=accelerator",
```
TheDoubleK: Feels like the `searchUrl`s here don't fully exercise all the constructing/deconstructing of URLs that happens within the `view_project` route.  It feels quite complex and brittle, so would be nice to see tests that cover a wider range of search URLs with various odd combinations of filters, etc. and see them reconstructed into the correct frontend search URL.

I know that's not what you want to hear right now, but if it doesn't happen now it might never happen...

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+        'direct-award/save-search.html',
+        search_summary=search_summary,
+        search_api_url=search_api_url,
+    )
+
+
+@direct_award.route('/projects/create', methods=['POST'])
+def project_create():
+    try:
+        api_project = data_api_client.create_direct_award_project(user_id=current_user.id,
```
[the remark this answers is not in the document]
samuelhwilliams: Yes - there's an alternative way into this flow that creates a project with no saved search.

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+    search_api_response = search_api_client.search_services(
+        index=framework['slug'],
+        **build_search_query(request.args, filters.values(), content_manifest, lots_by_slug)
+    )
+
+    search_summary = SearchSummary(search_api_response['meta']['total'], clean_request_query_params.copy(),
+                                   filters.values(), lots_by_slug)
+
+    # Embed search api URL in form so that the next view can be aware of it
+    # This is maybe not ideal
```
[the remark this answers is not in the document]
samuelhwilliams: Yeah...

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
         'search/services.html',
         **template_args
     )
+
+
+@direct_award.route('/save-search', methods=['GET'])
+def save_search():
+    # Get core data
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
```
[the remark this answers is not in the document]
samuelhwilliams: Ugh good point. I'll have a think.

app/main/views/g_cloud.py
```diff
@@ -23,11 +27,15 @@
     clean_request_args, get_request_url_without_any_filters
 )
 from ..helpers import framework_helpers
+from ..helpers.direct_award_helpers import is_direct_award_project_accessible
 
 from ..exceptions import AuthException
 from app import search_api_client, data_api_client, content_loader
 
 
+PROJECT_CREATED_MESSAGE = Markup("""Your new procurement project has been created.""")
```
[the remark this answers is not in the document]
samuelhwilliams: I will take this word out across the board.

app/main/views/g_cloud.py
```diff
@@ -223,6 +232,7 @@ def search_services():
                 filter_instance['label'] = capitalize_first(filter_instance['label'])
 
     clear_filters_url = get_request_url_without_any_filters(request, filters)
+    show_save_button = flask_featureflags.is_active('DIRECT_AWARD_PROJECTS')
```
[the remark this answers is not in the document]
samuelhwilliams: Nice, that works, thanks.

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+        search_api_url=search_api_url,
+    )
+
+
+@direct_award.route('/projects/create', methods=['POST'])
+def project_create():
+    try:
+        api_project = data_api_client.create_direct_award_project(user_id=current_user.id,
+                                                                  user_email=current_user.email_address,
+                                                                  project_name=request.form.get('project_name',
```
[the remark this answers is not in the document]
samuelhwilliams: They will have to submit a name (i.e. validation will happen) but that will be covered by a later PR; @MuriloDalRi is working on this now to implement the design for this page.

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+    return redirect(url_for('.view_project',
+                            project_id=project['id']
+                            ))
+
+
+@direct_award.route('/projects/<int:project_id>', methods=['GET'])
+def view_project(project_id):
+    # Get core data
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
```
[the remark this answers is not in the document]
samuelhwilliams: Done.

app/main/views/g_cloud.py
```diff
@@ -265,3 +276,115 @@ def search_services():
...
+    content_manifest = content_loader.get_manifest(framework['slug'], 'search_filters')
+    lots_by_slug = framework_helpers.get_lots_by_slug(framework)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(user_id=current_user.id, project_id=project_id)['project']
+    if not is_direct_award_project_accessible(project, current_user.id):
```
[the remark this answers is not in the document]
samuelhwilliams: The API currently doesn't enforce that check for this endpoint, and apparently access control is done frontend-side, so it's done here to be consistent with briefs.

app/main/views/g_cloud.py
samuelhwilliams: Done.

app/templates/direct-award/save-search.html
samuelhwilliams: Again, design to be done. But I've also put a tech debt ticket in about this a few days ago because we're really inconsistent across DM about how breadcrumbs work.

app/templates/direct-award/save-search.html
samuelhwilliams: Removed.

app/templates/direct-award/view-project.html
samuelhwilliams: :man_shrugging: I've fixed this in a PR that sits on top of this already (because you know, design to be done later), but I'll pull it in here anyway.

tests/fixtures/direct_award_project_fixture.json
samuelhwilliams: No.

tests/helpers.py
samuelhwilliams: As you noticed, this is just moving it out to a more central place.

tests/main/views/test_search.py
samuelhwilliams: Yep.

tests/fixtures/direct_award_project_searches_fixture.json
samuelhwilliams: Good idea.

app/main/views/g_cloud.py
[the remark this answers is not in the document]
TheDoubleK: It looked to me like the API will - as long as the client passes the user id in it will filter: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/direct_award.py#L23

But having checks in the frontend definitely doesn't hurt.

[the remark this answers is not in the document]
samuelhwilliams: This isn't calling list_projects, it's calling get_project (by id).

app/main/helpers/shared_helpers.py
```diff
@@ -13,3 +10,20 @@ def get_one_framework_by_status_in_order_of_preference(frameworks, statuses_in_o
         for framework in frameworks:
             if framework.get('status') == status:
                 return framework
+
+
+def construct_url_from_base_and_params(base_url, query_params):
```
TheDoubleK: Nice - this makes much more sense like this

app/main/__init__.py
```diff
@@ -1,6 +1,20 @@
-from flask import Blueprint
+from flask import Blueprint, flash, Markup, current_app
+from flask_login import login_required, current_user
+import flask_featureflags
 
 main = Blueprint('main', __name__)
+direct_award = Blueprint('direct_award', __name__, url_prefix='/buyers/direct-award')
```
TheDoubleK: Is this URL still up for grabs as part of the design/content makeover?  I'm not loving `direct-award` shown to users.

app/main/views/g_cloud.py
```diff
@@ -337,15 +337,16 @@ def project_create():
...
 
     return redirect(url_for('.view_project',
+                            framework_framework=framework_framework,
                             project_id=project['id']
                             ))
 
 
-@direct_award.route('/projects/<int:project_id>', methods=['GET'])
-def view_project(project_id):
+@direct_award.route('/<string:framework_framework>/projects/<int:project_id>', methods=['GET'])
```
TheDoubleK: Yep this `framework_framework` works here I think.

Same pattern as we already have for, e.g.
```
/<framework_framework>/opportunities
/<framework_framework>/opportunities/<brief_id>
```

app/main/views/g_cloud.py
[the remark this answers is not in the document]
TheDoubleK: Ah righto, sorry. My bad.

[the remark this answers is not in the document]
TheDoubleK: In that case it looks like the API client is wrong - this method shouldn't take or pass on a user ID: https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/data.py#L795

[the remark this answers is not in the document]
TheDoubleK: Because the API will just completely ignore it: https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/data.py#L795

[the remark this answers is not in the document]
samuelhwilliams: Yeah, I will fix that up shortly (but it's a lower priority at the moment).

TheDoubleK:
The API client is an issue and needs to be fixed up, but in terms of this PR I think it looks good enough for a feature that's a work-in-progress.

app/main/__init__.py
TheDoubleK: _also, you said you weren't going to do that_

samuelhwilliams: Hmm I thought you meant on the page, not in the URL. It's pretty much decided, but we can discuss it tomorrow if you feel strongly about it.

app/main/helpers/shared_helpers.py
samuelhwilliams: 👍

PR #585, opened 2017-08-21
Feedback forms

galund:
This feature was mostly developed during the Q1-end Firebreak sprint.

The feedback form (AKA "report-a-problem", per the gov.uk original) is in the DM Frontend Toolkit.

The POST handler sits here in the buyer app, because that's where publicly-available views go if they don't have another home.

Goes with this PR in the toolkit: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/367


app/main/views/feedback.py
```diff
@@ -0,0 +1,32 @@
...
+    if result.status_code != 200:
+        raise ServiceUnavailable('Google forms submission problem (status %d)'.format(result.status_code))
+    came_from = url_parse(request.form['uri'])
+    # strip netloc and scheme as we should ignore attempts to make us redirect elsewhere
+    replaced = came_from._replace(scheme='', netloc='')
+
+    flash(Markup(
+        """Thank you for your message. If you have more extensive feedback, please
+        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">email us</a> or
+        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
```
galund: @karlchillmaid can you give this message (that the users gets if they send us feedback) a last once-over please?

As far as a I know the air-table request is still valid, although might change "soon-ish" I think!

karlchillmaid: @galund @DilwoarH 

`If you have more detailed feedback, please email <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.`

Can't vouch for how long the air-table request will be valid, though.

DilwoarH: @karlchillmaid - would it be better to have

`Thank you for your message. If you have more detailed feedback, please email <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.`

karlchillmaid: Yep, sorry – hadn't intended to drop the thank you part of the message, it's a nice touch. Was focused on the rest of the message and only posted the bit to change.

samuelhwilliams: How will we know if the airtable request expires (other than getting support requests about it)? We shouldn't use that link if it's liable to die on us with no warning.

DilwoarH: screen shot 2017-08-29 at 15 50 28

@karlchillmaid - this is how it will look

app/main/views/feedback.py
```diff
@@ -0,0 +1,32 @@
...
+    came_from = url_parse(request.form['uri'])
+    # strip netloc and scheme as we should ignore attempts to make us redirect elsewhere
+    replaced = came_from._replace(scheme='', netloc='')
+
+    flash(Markup(
+        """Thank you for your message. If you have more detailed feedback, please email 
+        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
+        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
+        """))
+
```
DilwoarH: screen shot 2017-08-29 at 15 50 28
@karlchillmaid - this is how it will look

DilwoarH: Also @samuelhwilliams 's comment 
"How will we know if the airtable request expires (other than getting support requests about it)? We shouldn't use that link if it's liable to die on us with no warning."

samuelhwilliams: If it's just that we're changing from airtable to something else, that's probably fine - we just need to make sure we remember to update this link. :)

PR #586, opened 2017-08-22
Pull in apiclient 9.0.0

samuelhwilliams:
## Summary
Pull in apiclient 9.0.0 which removes the default value for `index` in the search API Client. Fix code and tests to provide this parameter as the latest live framework slug, which will be aliases to the current index.

## Ticket
Work required for https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -170,7 +170,8 @@ def get_filter_value_from_question_option(option):
     return (option.get('value') or option.get('label', '')).lower().replace(',', '')
 
 
-def build_search_query(request_args, lot_filters, content_builder, lots_by_slug, for_aggregation=False):
+def build_search_query(framework, request_args, lot_filters, content_builder, lots_by_slug,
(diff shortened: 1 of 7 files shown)


tests/main/presenters/test_search_presenters.py
```diff
@@ -220,8 +220,8 @@ def test_show_lots_and_categories_selection(self, search_api_client):
         url = "/g-cloud/search?q=&lot=cloud-software&otherfilter=somevalue&filterExample=option+1" \
               "&checkboxTreeExample=option+1&page=2"
         with self.app.test_request_context(url):
-            selection = build_lots_and_categories_link_tree(framework, lots, category_filter_group, flask.request,
-                                                            g9_builder)
+            selection = build_lots_and_categories_link_tree(framework, lots, category_filter_group,
+                                                            flask.request, g9_builder)
```
galund: not sure why these lines changed

PR #577, opened 2017-08-08
Show award status on opportunity page

CrystalPea:
For details please see the Trello ticket:
https://trello.com/c/fHudofRp/747-2-show-award-status-on-opportunity-page

awarded-opportunity-screenshot

PR #587, opened 2017-08-22
Change wording for application stats from "started" to "incomplete"

CrystalPea:
This was done to avoid confusion, following feedback from the
users. Users found former wording confusing, they thought that
completed applications should be included in started applications
count, as they have been started as well. To avoid such confusion
content decided to use binary oppositions completed/incomplete

Trello ticket: https://trello.com/c/n0sExgc5/811-update-content-for-application-stats

Code changed to match the content:
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/577/commits/fe71d56457bf717647572aa47492f6bbe110de74

### DIFF
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -392,7 +392,7 @@ class TestBriefPage(BaseBriefPageTest):
         completed_responses_section = document.xpath('//div[@id="completed-applications"]')[0]
 
         assert started_responses_section.xpath('div[@class="big-statistic"]/text()')[0] == '3'
-        assert started_responses_section.xpath('div[@class="statistic-name"]/text()')[0] == "Started applications"
+        assert started_responses_section.xpath('div[@class="statistic-name"]/text()')[0] == "Incomplete applications"
         assert started_responses_section.xpath('div[@class="statistic-description"]/text()')[0] == "3 SME, 0 large"
 
         assert completed_responses_section.xpath('div[@class="big-statistic"]/text()')[0] == '5'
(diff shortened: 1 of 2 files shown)


tests/main/views/test_marketplace.py
```diff
@@ -451,7 +451,7 @@ def test_dos_brief_has_application_stats_correctly_when_no_applications(self):
 
         assert started_responses_section.xpath('div[@class="big-statistic"]/text()')[0] == '0'
         assert completed_responses_section.xpath('div[@class="big-statistic"]/text()')[0] == '0'
-        assert started_responses_section.xpath('div[@class="statistic-name"]/text()')[0] == "Started applications"
+        assert started_responses_section.xpath('div[@class="statistic-name"]/text()')[0] == "Incomplete applications"
```
galund: Feels like the duplication in the test code could be factored out at some point.

CrystalPea: you mean serialising those assertions? I was thinking about it, in general I follow the rule that if similar code appears 1-2 times, I leave it as is, if it's >=3, then I serialise / refactor for DRYness, but I'm open to change my ways.

galund: Not sure I understand it as serializing, but you've had to update the same assertion three times, and those repeated things could be in a function. (Assuming we really need to assert the same thing in three different tests.)

CrystalPea: I think it is because the tests check if text is pluralised correctly for different number of applications.

PR #588, opened 2017-08-29
Backwards compatibility for new Brief statuses

katstevens:
Second of 5 backwards compatibility PRs to handle the two new Brief statuses, `cancelled` and `unsuccessful`.

1. Admin FE https://github.com/alphagov/digitalmarketplace-admin-frontend/pull/293
2. Buyer FE 
3. Scripts https://github.com/alphagov/digitalmarketplace-scripts/pull/144
4. Brief Responses FE https://github.com/alphagov/digitalmarketplace-brief-responses-frontend/pull/5
5. Briefs FE https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/17

See #582 for the equivalent changes for the awarded status.

- Adds new `cancelled` and `unsuccessful` statuses to existing Brief status lists
- Update templates for search results and public Brief page (treat both new statuses as 'closed' for now)
- Update tests and test fixtures

### DIFF
--- a/app/main/forms/brief_forms.py
+++ b/app/main/forms/brief_forms.py
@@ -3,7 +3,7 @@ from wtforms import IntegerField, SelectMultipleField
 from wtforms.validators import NumberRange
 
 
-CLOSED_BRIEF_STATUSES = ['closed', 'awarded', 'cancelled', 'unsuccessful']
+BRIEF_AWARD_STATUSES = ['awarded', 'cancelled', 'unsuccessful']
 
 
 class BriefSearchForm(Form):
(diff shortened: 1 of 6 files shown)


app/main/forms/brief_forms.py
```diff
@@ -47,12 +50,12 @@ def get_briefs(self):
         else:
             statuses = [id_ for id_, label in self.status.choices]
         if 'closed' in statuses:
-            statuses.append('awarded')
+            statuses.extend(CLOSED_BRIEF_STATUSES)
```
benvand: Feels a little odd adding 'closed' only to pull it out later.
Could go for something like:
```
if 'closed' in statuses:
    # Add hidden closed statuses
    statuses.extend(['awarded', 'cancelled', 'unsuccessful'])
```
Not _that_ bothered but it was a bit hard to read on the first pass

katstevens: I think this problem boils down to me being unable to think of a good name for the group of statuses `awarded, cancelled, unsuccessful`. How about I change the list to be `BRIEF_AWARD_STATUSES` and remove `closed`?

Hopefully this hack won't be needed if/when we start filtering by the statuses properly in the View Opportunities page (rather than just treating them as closed).

PR #591, opened 2017-09-01
Revert "Feedback forms"

samuelhwilliams:
The feedback form submission fails if a user is logged in due to missing CSRF token.

We will revert this for now as we have other dev activity prioritised above fixing this for the moment.


MuriloDalRi:
👍

bower.json
```diff
@@ -5,7 +5,7 @@
     "jquery": "1.11.2",
     "hogan": "3.0.2",
     "jquery-details": "https://github.com/mathiasbynens/jquery-details/archive/v0.1.0.tar.gz",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v24.0.0",
+    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v23.3.0",
```
TheDoubleK: Might be worth keeping the toolkit upgrade?

samuelhwilliams: I'd rather not so as this is an automatic revert of the entire PR. I'd rather not pick and choose which bits to keep if that's OK, as it makes it less clear what's happened.

TheDoubleK: As long as we're 100% sure none of the more recent commits are relying on new bits of toolkit...

galund:
Darn; wonder why it worked locally.

samuelhwilliams:
Because you probably weren't logged in.

PR #592, opened 2017-09-04
Buyer account page

DilwoarH:
As a G-Cloud buyer
I need to be able to return to a list of the projects i am running
So that I can keep track of my projects

Ticket: https://trello.com/c/FiR3NLyZ/679-buyer-account-page

buyer-frontend


app/assets/scss/_saved-searches.scss
```diff
@@ -0,0 +1,19 @@
...
+            text-align: left;
+            width: 25%;
+        }
+        
+        &.summary-item-body {
+            margin-bottom: 60px;
+        }
+    }
+
+}
```
samuelhwilliams: We apparently like to include an extra blank line at the end of files. Please add one and consider updating your IDE to automatically include it on save.

TheDoubleK: Not a blank line necessarily, but a line terminator on the final line, yes.  It's generally considered good practice, for several (mainly historical) reasons... some discussion here: https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

TheDoubleK: Yep, summary table looks like the thing here, and if it needs a variation from what's in the toolkit it's better to make the adjustment there rather than at a per-app level. 
 This is especially important for us now we're splitting out more and more apps - we need to rely on the toolkit to provide a consistent look and feel across all apps. 
 http://alphagov.github.io/digitalmarketplace-frontend-toolkit/summary-table.html

TheDoubleK: At this point adding any new per-app styles for stuff should set off alarm bells I think.

DilwoarH:
@TheDoubleK @samuelhwilliams @galund - updated templates to follow DM standards and removed any unused css

app/main/views/g_cloud.py
```diff
@@ -276,6 +276,25 @@ def search_services():
...
 
+@direct_award.route('', methods=['GET'])
+def saved_search_overview():
+    projects = data_api_client.find_direct_award_projects(current_user.id).get('projects', [])
+    open_searches = []
+    closed_searches = []
+
+    for project in projects:
+        if project['active']:
+            open_searches.append(project)
```
samuelhwilliams: Let's call these `open_projects` and `closed_projects` to keep terminology consistent.

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+      caption="Summary table",
+      empty_message="You have no saved searches",
+      field_headings=[
+        "Name",
+        "Saved"  ],
+      field_headings_visible=True
+    ) %}
+      {% call summary.row() %}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_framework="g-cloud", project_id=item.id)) }}
+        {{ summary.text(item.createdAt | datetimeformat ) }}
```
samuelhwilliams: We should use `utcdatetimeformat` so that the timezone is included in the date string (a requirement when displaying times).

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+{% endblock %}
+
+{% block main_content %}
+
+<div class="direct-award-saved-searches">
+  <header class="page-heading">
+    <h1>
+      Saved searches
+    </h1>
+    <p class="context">Find, compare and buy cloud services through the G-Cloud framework.</p>
```
TheDoubleK: Can this just be "Find, compare and buy cloud services"?

No need to mention the framework at all.

samuelhwilliams: This content needs to come via the frameworks repo using the content loader. Any content that ties a page to a particular framework should be dynamically loaded. Also as others have said, we shouldn't use the word 'framework' even for testing. Let's just drop `through the G-Cloud framework` altogether. The first half of the sentence should be built as `Find, compare and buy {{ framework_short }}`, where framework_short can be found in dm-frameworks/frameworks/g-cloud-9/messages/descriptions.yml

app/main/views/g_cloud.py
```diff
@@ -276,6 +276,25 @@ def search_services():
     )
 
 
+@direct_award.route('', methods=['GET'])
```
TheDoubleK: Could this route be parameterised e.g. `/saved-searches/<framework_framework>`, and then that could be passed in to the template to avoid hard coding `framework_framework="g-cloud"` in the `url_for` calls?

samuelhwilliams: This endpoint should be `/<string:framework_slug>`

TheDoubleK: Briefs do something similar for suppliers view of opportunities:
```
/<framework_framework>/opportunities
/<framework_framework>/opportunities/<brief_id>
```

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+    {% call(item) summary.list_table(open_searches,
+      caption="Summary table",
+      empty_message="You have no saved searches",
+      field_headings=[
+        "Name",
+        "Saved"  ],
+      field_headings_visible=True
+    ) %}
+      {% call summary.row() %}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_framework="g-cloud", project_id=item.id)) }}
```
samuelhwilliams: We shouldn't be hardcoding g-cloud into the template. Once we update the URL above to include the framework family, you can pass that into render_template to remove the hardcoding here.

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+    {% call(item) summary.list_table(closed_searches,
+      caption="Summary table",
+      empty_message="You have no saved searches",
+      field_headings=[
+        "Name",
+        "Shortlist created"  ],
+      field_headings_visible=True
+    ) %}
+      {% call summary.row() %}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_framework="g-cloud", project_id=item.id)) }}
```
samuelhwilliams: Same here with g-cloud.

DilwoarH:
@samuelhwilliams @TheDoubleK - comments addressed

app/main/views/g_cloud.py
```diff
@@ -276,6 +276,25 @@ def search_services():
...
+def saved_search_overview():
+    projects = data_api_client.find_direct_award_projects(current_user.id).get('projects', [])
```
galund: Can we call it `buyer_dashboard` instead? The name you have here could too easily apply to just one saved search.

app/main/views/g_cloud.py
```diff
@@ -276,6 +276,25 @@ def search_services():
...
+    open_searches = []
+    closed_searches = []
```
galund: 'Closed' is going to be confusing for DOS developers, because 'closed' is a status quite similar to our 'locked' status (i.e. it comes before 'awarded').

Can we just call this `active_searches` and `inactive_searches`?

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+        "Saved"  ],
+      field_headings_visible=True
+    ) %}
+      {% call summary.row() %}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_framework="g-cloud", project_id=item.id)) }}
+        {{ summary.text(item.createdAt | datetimeformat ) }}
+      {% endcall %}
+    {% endcall %}
+
+    {{ summary.heading("Search Ended", id="search_ended_table") }}
```
galund: Along similar lines to the previous comment, can we call this 'Archived searches'?

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+  <header class="page-heading">
+    <h1>
+      Saved searches
+    </h1>
+    <p class="context">Find, compare and buy cloud services through the G-Cloud framework.</p>
+  </header>
+
+  <div class="grid">
+    
+    {{ summary.heading("Searching", id="searching_table") }}
```
galund: 'Current searches' perhaps? We can leave the final decisions for @karlchillmaid but we can tweak in the meantime :)

app/templates/direct-award/index.html
```diff
@@ -0,0 +1,71 @@
...
+      {% call summary.row() %}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_framework="g-cloud", project_id=item.id)) }}
+        {{ summary.text(item.lockedAt | datetimeformat ) }}
+      {% endcall %}
+    {% endcall %}
+
+  </div>
+  
+</div>
+{% endblock %}
```
galund: It's hard to be consistent with indenting in templates but I think that the div should be indented because of the jinja block.

PR #595, opened 2017-09-05
709 save search

MuriloDalRi:
## Summary

Adds validation to name your search page and design

## Ticket

https://trello.com/c/ubKd3ztP/709-save-search-and-name-your-procurement-design-and-content

### DIFF
--- a/app/assets/scss/_save-search-page.scss
+++ b/app/assets/scss/_save-search-page.scss
@@ -1,0 +1,26 @@
+.save-search-page {
+    .search-results-summary{
+        -webkit-box-sizing: border-box;
+        -moz-box-sizing: border-box;
+        box-sizing: border-box;
+        clear: both;
+        border-left: 9px solid $border-colour;
+        padding: 0 0.78947em;
+        margin-bottom: 0.78947em;
+        font-size: 19px;
+        line-height: 1.31579;
+    }
+
+    h2 {
+        font-size: 24px;
+    }
+
+    .question {
+        margin-bottom: 0;
+    }
+
+    .button-save {
+        margin-top: 0px;
+        font-size: 20px;
+    }
+}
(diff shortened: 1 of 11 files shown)


app/assets/scss/_save-search-page.scss
```diff
@@ -0,0 +1,26 @@
+.save-search-page {
+    .search-results-summary{
+        -webkit-box-sizing: border-box;
+        -moz-box-sizing: border-box;
+        box-sizing: border-box;
+        clear: both;
+        border-left: 9px solid $border-colour;
+        padding: 0 0.78947em;
+        margin-bottom: 0.78947em;
```
DilwoarH: this comes from `.panel` govuk scss toolkit. Need to look into possibility of using this instead of copying CSS.

app/templates/direct-award/save-search.html
```diff
@@ -21,32 +21,48 @@
...
+          Save your search
+        </h1>
+      </header>
+    </div>
   </div>
+  <form method="post" action="{{ url_for('direct_award.project_create', framework_framework=framework_framework)}}" id="createProjectForm">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <div class="search-results-summary">
+          <p>{{ search_summary.markup() }}</p>
```
DilwoarH: http://alphagov.github.io/digitalmarketplace-frontend-toolkit/search-summary.html

we should use this method

galund: I'm not sure we should - it's one of the tiniest most pointless templates I've ever seen :)

app/assets/scss/_search-page.scss
```diff
@@ -2,15 +2,37 @@
...
+    padding: 5px;
+  }
+
+  .save-button {
+    display: inline-block;
+    right: 0;
+    float: right;
+
+    .button-save {
+      font-size: 16px;
```
samuelhwilliams: I feel like this is too small. We should probably be looking at `@core-19` for the button itself.

PR #596, opened 2017-09-05
Show appropriate banners for cancelled and unsuccessful briefs

CrystalPea:
Trello ticket: https://trello.com/c/sqfpwtPW/797-1-show-award-status-when-buyer-has-cancelled-procurement

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -56,7 +56,11 @@
       with
       type = "temporary-message",
       heading = "This opportunity was cancelled",
-      message = "You can't apply for this opportunity now. The buyer may publish an " + "updated version"|nbsp + " on the " + "Digital Marketplace"|nbsp
+      message = (
+        "The buyer cancelled this opportunity, for example because they no longer have the budget."
+        + "<br />"|safe
+        + "They may publish an updated version later."
+      )
     %}
       {% include "toolkit/notification-banner.html" %}
     {% endwith %}
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -49,6 +49,19 @@
     {% endwith %}
   </div>
 </div>
+{% elif brief.status == 'cancelled' %}
+<div class="grid-row">
+  <div class="column-one-whole">
+    {%
+      with
+      type = "temporary-message",
+      heading = "This opportunity was cancelled",
+      message = "You can't apply for this opportunity now. The buyer may publish an " + "updated version"|nbsp + " on the " + "Digital Marketplace"|nbsp
+    %}
+      {% include "toolkit/notification-banner.html" %}
+    {% endwith %}
+  </div>
+</div>
 {% elif brief.status == 'awarded' %}
 <div class="grid-row">
   <div class="column-one-whole">
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -23,7 +23,7 @@
 
 {% block main_content %}
 
-{% if brief.status in ['closed', 'cancelled', 'unsuccessful'] %}
+{% if brief.status == 'closed' %}
 <div class="grid-row">
   <div class="column-one-whole">
     {%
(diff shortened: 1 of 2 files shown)


app/templates/brief.html
```diff
@@ -49,6 +49,40 @@
...
+{% elif brief.status == 'cancelled' %}
+<div class="grid-row">
+  <div class="column-one-whole">
+    {%
+      with
+      type = "temporary-message",
+      heading = "This opportunity was cancelled",
+      message = (
+        "The buyer cancelled this opportunity, for example because they no longer have the budget."
+        + "<br />"|safe
```
TheDoubleK: I recently (yesterday!) learned that you can avoid the need for `|safe` by setting the value beforehand using something like 
```
{% set message %}
The first line<br />
is above the second line.
{% endset %}
``` 

See, e.g. https://github.com/alphagov/digitalmarketplace-supplier-frontend/commit/d8c751f33fc12641623bf725aef9b856514762b1

package-lock.json
```diff
@@ -0,0 +1,3028 @@
+{
```
TheDoubleK: Is this file meant to be committed?

tests/main/views/test_marketplace.py
```diff
@@ -977,6 +977,50 @@ def test_supplier_size_visible_on_award_banner(self):
         assert 'Company size: SME' in awarded_banner.xpath('p/text()')[2]
 
 
+class TestCancelledBriefPage(BaseBriefPageTest):
+    def setup_method(self, method):
+        super(TestCancelledBriefPage, self).setup_method(method)
+        self.brief['briefs']['status'] = 'cancelled'
+
+    def test_cancelled_banner_shown_on_cancelled_brief_page(self):
```
TheDoubleK: Might also be worth testing that the cancelled banner is not shown on not-cancelled briefs?  You might just be able to add an assertion to an existing test of non-cancelled briefs, rather than needing a whole new test.

TheDoubleK: (Also for the other new banners - be good to test they don't show up where they shouldn't)

package-lock.json
CrystalPea: Oops, it was not, I'm gonna fix it :x

tests/main/views/test_marketplace.py
CrystalPea: This is a great idea, I will do that 👍

app/templates/brief.html
```diff
@@ -43,7 +43,41 @@
       with
       type = "temporary-message",
       heading = "This opportunity was withdrawn on " + "{}".format(brief.withdrawnAt|dateformat)|nbsp,
-      message = "You can't apply for this opportunity now. The buyer may publish an " + "updated version"|nbsp + " on the " + "Digital Marketplace"|nbsp
+      message = "You can't apply for this opportunity now. The buyer may publish an " + "updated version"|nbsp + " on the " + "Digital Marketplace."|nbsp
```
katstevens: I'm curious why `"updated version|nbsp"` is separated out here? Why not have it all in one string with `|nbsp` at the end (as per the heading on the line above). Also, why do we need the `|nbsp` on the message in the first place?  I can understand wanting the header all on one row but does the same apply for the message? Or is this required for the banner macro?

tests/main/views/test_marketplace.py
katstevens: For completeness, should we then also test that the award details aren't shown on non-awarded briefs?

app/templates/brief.html
CrystalPea: That might be a question @benvand is better equipped to answer :)

tests/main/views/test_marketplace.py
CrystalPea: I made one test that checks all of that, gonna push it up in a sec :)

app/templates/brief.html
CrystalPea: Is it preferable to do it this way?

tests/main/views/test_marketplace.py
```diff
@@ -380,6 +380,17 @@ def test_dos_brief_has_correct_title(self):
         assert page_heading.xpath('h1/text()')[0] == self.brief['briefs']['title']
         assert page_heading.xpath('p[@class="context"]/text()')[0] == self.brief['briefs']['organisation']
 
+    @pytest.mark.parametrize('status', ['closed', 'unsuccessful', 'cancelled', 'awarded'])
```
CrystalPea: @katstevens @TheDoubleK what do you think?

app/templates/brief.html
katstevens: Ben did indeed answer my question 😸 Happy to keep this as it is.

app/templates/brief.html
CrystalPea: @TheDoubleK I changed that now, do you think the code is ready for merging now?

PR #597, opened 2017-09-05
End your search

DilwoarH:
As a G-Cloud buyer
I need a fixed list of services
so that I can start the shortlisting process

Ticket: https://trello.com/c/gs2I3Cj2/676-end-your-search

### DIFF
--- a/app/assets/scss/_end-your-search-page.scss
+++ b/app/assets/scss/_end-your-search-page.scss
@@ -1,0 +1,10 @@
+.end-search-page {
+  .end-your-search-button {
+    @include core-16;
+    margin-top: 1.5em;
+  }
+
+  h2 {
+    margin-bottom: 1em;
+  }
+}
--- a/app/assets/scss/_end-your-search-page.scss
+++ b/app/assets/scss/_end-your-search-page.scss
@@ -1,10 +1,10 @@
 .end-search-page {
   .end-your-search-button {
-    @include core-16;
+    @include core-19;
     margin-top: 1.5em;
   }
 
-  h2 {
+  .dmspeak .heading-xmedium {
     margin-bottom: 1em;
   }
 }
--- a/app/assets/scss/_end-your-search-page.scss
+++ b/app/assets/scss/_end-your-search-page.scss
@@ -1,9 +1,4 @@
 .end-search-page {
-  .end-your-search-button {
-    @include core-19;
-    margin-top: 1.5em;
-  }
-
   .dmspeak .heading-xmedium {
     margin-bottom: 1em;
   }
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -275,6 +275,13 @@ def search_services():
         **template_args
     )
 
+@direct_award.route('/<string:direct_award_id>/end-search', methods=['GET'])
+def end_search(direct_award_id):
+    return render_template(
+        'direct-award/end-search.html'
+    )
+
+
 
 @direct_award.route('/<string:framework_framework>', methods=['GET'])
 def saved_search_overview(framework_framework):
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -293,7 +293,7 @@ def saved_search_overview(framework_framework):
     closed_projects = []
 
     for project in projects:
-        if project['lockedAt'] is None:
+        if project['active']:
             open_projects.append(project)
         else:
             closed_projects.append(project)
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -435,7 +435,7 @@ def view_project(framework_framework, project_id):
                            customer_benefits_record_form_email=framework_urls['customer_benefits_record_form_email'])
 
 
-@direct_award.route('/<string:framework_framework>/projects/<int:project_id>/end-search', methods=['GET'])
+@direct_award.route('/<string:framework_framework>/projects/<int:project_id>/end-search', methods=['GET', 'POST'])
 def end_search(framework_framework, project_id):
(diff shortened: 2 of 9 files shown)


app/assets/scss/_end-your-search-page.scss
```diff
@@ -0,0 +1,10 @@
+.end-search-page {
```
TheDoubleK: Could this go in the buttons section of the front-end toolkit?

TheDoubleK: Perhaps with a more generic name, to allow for reuse.

app/templates/direct-award/end-search.html
```diff
@@ -0,0 +1,69 @@
...
+      {
+        "link": "/buyers",
+        "label": "Your Account"
+      },
+      {
+        "link": "/buyers/direct-award",
+        "label": "Saved Searches"
+      },
+      {
+        "link": "/buyers/direct-award/{}".format(project.id),
```
samuelhwilliams: These should be `url_for`d

app/assets/scss/_end-your-search-page.scss
galund: We're gonna go with the standard (bigger) button, anyway :) - thanks

app/templates/direct-award/end-search.html
```diff
@@ -0,0 +1,78 @@
...
+  <p>download the shortlist to help you compare services.</p>
+
+  
+  <br>
+  <form 
+    action="{{ url_for('direct_award.end_search_action', framework_framework=framework.framework, project_id=project.id) }}"
+    method="POST"  
+    >
+    <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" /> 
+    <button class="button-save end-your-searxch-button">End search and continue</button>
```
samuelhwilliams: `searxch`

app/main/views/g_cloud.py
```diff
@@ -34,6 +34,7 @@
 
 
 PROJECT_CREATED_MESSAGE = Markup("""Your new project has been created.""")
+PROJECT_ENDED_MESSAGE = Markup("""Your search has been ended. You can now download your shortlist.""")
```
samuelhwilliams: Is this real content? Seems like a weird sentence.

app/main/views/g_cloud.py
```diff
@@ -292,7 +293,7 @@ def saved_search_overview(framework_framework):
     closed_projects = []
 
     for project in projects:
-        if project['active']:
+        if project['lockedAt'] is None:
```
galund: Let's put this back - the distinction for this project list page is between open procurements and closed ones. Saved searches are still 'open' in this sense - they have neither been awarded nor archived, so they are 'active'.

if we want to show the locked status here, then that's a different thing?

app/templates/direct-award/end-search.html
```diff
@@ -0,0 +1,76 @@
...
+  </header>
+
+  <div class="dmspeak">
+    <h2 class="heading-xmedium">Before you end your search</h2>
+  </div>
+  <p>You should only continue when you have finished searching for cloud services</p>
+  <p>and are ready to start comparing the services you've found</p>
+
+  <br>
+  <p>You can not edit your search after it has ended.</p>
```
galund: "Can't" (or "cannot"), rather than "can not" I think?

app/main/views/g_cloud.py
galund: Placeholder, real thing can edit for Karl, agree first sentence should probably have been "marked as complete" or something, but somehow be consistent with "ending".. shrug :)

app/templates/direct-award/end-search.html
```diff
@@ -0,0 +1,78 @@
...
+      End search
+    </h1>
+  </header>
+
+  <div class="dmspeak">
+    <h2 class="heading-xmedium">Before you end your search</h2>
+  </div>
+  <p>You should only continue when you have finished searching for cloud services</p>
+  <p>and are ready to start comparing the services you’ve found</p>
+
```
galund: There's an errant paragraph el here

app/templates/errors/400.html
```diff
@@ -0,0 +1,24 @@
...
+  <div class="index-page grid-row">
+    <div class="column-two-thirds">
+      <header class="page-heading-smaller">
+        <h1>Sorry, there was a problem with your request</h1>
+      </header>
+      <p>
+        {% if error_message %}
+          {{ error_message }}
+        {% else %}
+          Please do not attempt the same request again.
```
galund: Assume there's only contrived ways for this to happen, but what comes to mind is multiple user concurrency, so when we add collaboration we could think about how to make the recovery more friendly (e.g. provide a link back to the project)

samuelhwilliams: This was mostly copied from the user-frontend, but yeah we should sort out the content eventually.

We should bring all the error message templates into the frontend-toolkit eventually. I started doing that but it seems like a non-trivial thing to link them all up from a central place, so shelved it for now. I dropped a ticket in the tech debt board a month or two ago about it anyway.

app/templates/direct-award/end-search.html
samuelhwilliams: Good spot; we were missing a couple of div wrappers to get the proper two-thirds styling. Have added.

app/templates/errors/400.html
samuelhwilliams: Regarding contrived ways for this to happen: if you go back in the browser you can hit this page again, and in Chrome it automatically re-sends the request so it's not that difficult to conceivably hit this 400 as a fairly average user. But yeah we could probably have something nicer once collaboration comes in.

PR #599, opened 2017-09-06
Shortlist downloaded state (for labs)

samuelhwilliams:
## Summary
We haven't been able to implement the API for downloading the shortlist in time for user research tomorrow. We **may** want to mock this behaviour out. If we decide that we do, this will meet the needs for a temporary solution and we should revert this at the end of the day once labs are finished and the real process is in place.

This PR will cause a cookie to be set when either of the shortlist links are clicked. When the search overview page is loaded, this cookie is checked and, if set, triggers the intended state on the overview.

## Example
shortlist-downloaded

## Ticket
https://trello.com/c/igRhA3Bt/711-shortlist-file-download

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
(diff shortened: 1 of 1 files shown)


TheDoubleK:
If you swear on your honour you won't forget to remove it... for today only... then OK :+1:

samuelhwilliams:
I totally promise.

PR #603, opened 2017-09-08
Adds error messages, updates to content and new tests  for end-search

DilwoarH:
Ticket: https://trello.com/c/gs2I3Cj2/676-end-your-search

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -38,7 +38,7 @@ from app import search_api_client, data_api_client, content_loader
 END_SEARCH_LIMIT = 100  # TODO: This should be done in the API.
 PROJECT_SAVED_MESSAGE = Markup("""Search saved.""")
 PROJECT_ENDED_MESSAGE = Markup("""Search ended. You can now download your search results.""")
-TOO_MANY_RESULTS_ERROR = Markup("""
+TOO_MANY_RESULTS_MESSAGE = Markup("""
     You have too many results. Choose a category or add filters to refine your search.""")
 
 
@@ -461,7 +461,7 @@ def end_search(framework_framework, project_id):
     disable_end_search_btn = False
 
     if int(search_count) > END_SEARCH_LIMIT:
-        flash(TOO_MANY_RESULTS_ERROR, 'error')
+        flash(TOO_MANY_RESULTS_MESSAGE, 'error')
         disable_end_search_btn = True
 
     if not framework or not project:
@@ -470,7 +470,7 @@ def end_search(framework_framework, project_id):
     if project['lockedAt']:
         abort(400)
 
-    if request.method == 'POST' and not int(search_count) > END_SEARCH_LIMIT:
+    if request.method == 'POST' and int(search_count) <= END_SEARCH_LIMIT:
         try:
             data_api_client.lock_direct_award_project(user_email=current_user.email_address, project_id=project_id)
         except HTTPError as e:
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -464,9 +464,6 @@ def end_search(framework_framework, project_id):
         flash(TOO_MANY_RESULTS_ERROR, 'error')
         disable_end_search_btn = True
 
-    if not is_direct_award_project_accessible(project, current_user.id):
-        abort(404)
-
     if not framework or not project:
         abort(404)
 
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -35,8 +35,11 @@ from ..exceptions import AuthException
 from app import search_api_client, data_api_client, content_loader
 
 
+END_SEARCH_LIMIT = 100  # TODO: This should be done in the API.
 PROJECT_SAVED_MESSAGE = Markup("""Search saved.""")
-PROJECT_ENDED_MESSAGE = Markup("""Your search has been ended. You can now download your shortlist.""")
+PROJECT_ENDED_MESSAGE = Markup("""Search ended. You can now download your search results.""")
+TOO_MANY_RESULTS_ERROR = Markup("""
+    You have too many results. Choose a category or add filters to refine your search.""")
 
 
 @main.route('/g-cloud')
@@ -443,8 +446,24 @@ def view_project(framework_framework, project_id):
 def end_search(framework_framework, project_id):
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
     framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
+    frameworks_by_slug = framework_helpers.get_frameworks_by_slug(data_api_client)
 
-    project = data_api_client.get_direct_award_project(project_id).get('project')
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+    if not is_direct_award_project_accessible(project, current_user.id):
+        abort(404)
+
+    searches = data_api_client.find_direct_award_project_searches(user_id=current_user.id,
+                                                                  project_id=project['id'])['searches']
+    search = list(filter(lambda x: x['active'], searches))[0]
+    search_meta = SearchMeta(search['searchUrl'], frameworks_by_slug)
+    search_count = search_meta.search_summary.count
+    disable_end_search_btn = False
+
+    if int(search_count) > END_SEARCH_LIMIT:
+        flash(TOO_MANY_RESULTS_ERROR, 'error')
+        disable_end_search_btn = True
+
     if not is_direct_award_project_accessible(project, current_user.id):
         abort(404)
 
@@ -454,7 +473,7 @@ def end_search(framework_framework, project_id):
     if project['lockedAt']:
         abort(400)
 
-    if request.method == 'POST':
+    if request.method == 'POST' and not int(search_count) > END_SEARCH_LIMIT:
         try:
             data_api_client.lock_direct_award_project(user_email=current_user.email_address, project_id=project_id)
         except HTTPError as e:
@@ -467,7 +486,8 @@ def end_search(framework_framework, project_id):
     return render_template(
         'direct-award/end-search.html',
         project=project,
-        framework=framework
+        framework=framework,
+        disable_end_search_btn=disable_end_search_btn,
     )
 
 
--- a/app/templates/direct-award/end-search.html
+++ b/app/templates/direct-award/end-search.html
@@ -65,11 +65,9 @@
         method="POST"
         >
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <button class="button-save end-your-search-button"
-          {% if disable_end_search_btn %} 
-           disabled="disabled"
-          {% endif %}
-        >End search and continue</button>
+        <button class="button-save end-your-search-button" {% if disable_end_search_btn %}disabled="disabled"{% endif %}>
+          End search and continue
+        </button>
       </form>
 
       {%
--- a/app/templates/direct-award/end-search.html
+++ b/app/templates/direct-award/end-search.html
@@ -41,7 +41,7 @@
           End search
         </h1>
       </header>
-
+      
       <div class="dmspeak">
         <h2 class="heading-xmedium">Before you end your search</h2>
       </div>
@@ -65,7 +65,11 @@
         method="POST"
         >
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <button class="button-save end-your-search-button">End search and continue</button>
+        <button class="button-save end-your-search-button"
+          {% if disable_end_search_btn %} 
+           disabled="disabled"
+          {% endif %}
+        >End search and continue</button>
       </form>
 
       {%
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -162,6 +162,12 @@ class BaseApplicationTest(object):
         return project
 
     @staticmethod
+    def _get_direct_award_lock_project_fixture():
+        fixture = BaseApplicationTest._get_fixture_data('direct_award_project_fixture.json')
+        fixture['project']['lockedAt'] = "2017-09-08T14:07:14.716154Z"
+        return fixture
+
+    @staticmethod
     def _get_direct_award_project_searches_fixture():
         return BaseApplicationTest._get_fixture_data('direct_award_project_searches_fixture.json')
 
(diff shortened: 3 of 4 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -326,3 +327,38 @@ def test_search_api_urls_convert_to_correct_frontend_urls(self, search_api_url,
         doc = html.fromstring(body)
         assert html_escape(frontend_url) in body
         assert len(doc.xpath('//a[@href="{}"]'.format(frontend_url)))
+
+
+class TestDirectAwardEndSearch(TestDirectAwardBase):
+    def setup_method(self, method):
+        super(TestDirectAwardEndSearch, self).setup_method(method)
+
+        self.g9_search_results = self._get_g9_search_results_fixture_data()
```
galund: This is already in the base class isn't it. (Looks like you copied from `TestDirectAwardURLGeneration`, but that doesn't inherit from `TestDirectAwardBase`.) Same goes for some other line in the setup method I think?

tests/main/views/test_direct_award.py
```diff
@@ -326,3 +327,38 @@ def test_search_api_urls_convert_to_correct_frontend_urls(self, search_api_url,
...
+
+        data_api_client.find_direct_award_project_searches = mock.Mock()
+        data_api_client.find_direct_award_project_searches.return_value = \
+            self._get_direct_award_project_searches_fixture()
+
+    def test_end_search_page_renders(self):
+        self.login_as_buyer()
+        res = self.client.get('/buyers/direct-award/g-cloud/projects/1/end-search')
+        assert res.status_code == 200
+        end_search_found = re.findall(r'<h1>\s*End search\s*<\/h1>', res.get_data(as_text=True))
```
samuelhwilliams: I think we generally use xpath as opposed to text/regex matches now, so it would be nice to use that if you can. Should pretty much just be as below (albeit structured more nicely)
```
from lxml import html
assert len(html.fromstring(res.get_data(as_text=True).xpath('//h1[text()="End search"]') == 1
```

tests/fixtures/direct_award_project_locked_fixture.json
```diff
@@ -0,0 +1,9 @@
+{
```
samuelhwilliams: It would be better not to add a new fixture for this but to mutate the existing fixture, e.g. by adding a `locked_at` parameter to the calling function to set the attribute to whatever the test needs. Might need to tweak implementation slightly for this, but it would be cleaner.

tests/main/views/test_direct_award.py
galund: In my head the rule is that if you want to test the structure, it's better to use xpath. In this case we're testing that an H1 is there so you can definitely argue that xpath is better, although I'm not too fussed personally ;)

samuelhwilliams: I assume you mean 'xpath _isn't better', but I'm ok with you leaving it as-is. ;)

tests/main/views/test_direct_award.py
```diff
@@ -326,3 +327,20 @@ def test_search_api_urls_convert_to_correct_frontend_urls(self, search_api_url,
...
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert len(doc.xpath('//h1[contains(normalize-space(), "End search")]')) == 1
+
+    def test_end_search_redirects_to_project_page(self):
+        self.login_as_buyer()
+        data_api_client.lock_direct_award_project = mock.Mock()
+        data_api_client.lock_direct_award_project.return_value = self._get_direct_award_lock_project_fixture()
+        res = self.client.post('/buyers/direct-award/g-cloud/projects/1/end-search')
+        assert res.status_code == 302
+        assert re.findall(r'\/buyers\/direct-award\/g-cloud\/projects\/1$', res.location)
```
galund: Can you just use the `in` operator here, if so it'd be a lot more readable than the regex.

samuelhwilliams: Or just res.location.endswith.

And if you're using raw strings you shouldn't need to escape the slashes.

tests/helpers.py
```diff
@@ -158,6 +158,12 @@ def _get_direct_award_project_fixture(**kwargs):
         return project
 
     @staticmethod
+    def _get_direct_award_lock_project_fixture(**kwargs):
+        fixture = BaseApplicationTest._get_fixture_data('direct_award_project_fixture.json')
+        fixture['project']['lockedAt'] = "2017-09-08T14:07:14.716154Z"
+        return fixture
```
samuelhwilliams: This isn't quite what I had in mind and we probably don't want to hardcode a random timestamp in for this fixture. It should probably just use the current time, so that it plays ball with freezegun for locking times during testing. This method doesn't need to take in `**kwargs` though, because it doesn't do anything with it, so let's at least drop that.

It would be better still to have a more general purpose mutator for this fixture so you could replace any set of values with custom ones, in which case it could take in `**kwargs` and mutate the fixture appropriately.

DilwoarH:
PR updated with new error message when results are too many

screen shot 2017-09-15 at 11 32 59

MuriloDalRi:
👍

samuelhwilliams:
We need to reduce the limit on this to 100 services now (from 1000). I've been looking into the data transfer overhead for building the shortlist. I've spoken to Richard B (who spoke to Lisa) and got the "signed off" at product level.

In order to raise this limit we'll need a proper job runner for the Marketplace to handle long-running and/or computationally-heavy tasks asynchronously.

@galund @DilwoarH

DilwoarH:
@samuelhwilliams @galund @karlchillmaid 

As 100 is a pretty low number, do we need to update the content on this page?

app/main/views/g_cloud.py
```diff
@@ -437,8 +439,24 @@ def view_project(framework_framework, project_id):
...
+        abort(404)
+
+    searches = data_api_client.find_direct_award_project_searches(user_id=current_user.id,
+                                                                  project_id=project['id'])['searches']
+    search = list(filter(lambda x: x['active'], searches))[0]
+    search_meta = SearchMeta(search['searchUrl'], frameworks_by_slug)
+    search_count = search_meta.search_summary.count
+    disable_end_search_btn = False
+
+    if int(search_count) > 999:
```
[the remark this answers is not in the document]
DilwoarH: will get it done

karlchillmaid:
I think we can keep the content the same:
`You have too many results. Choose a category or add filters to refine your search.`

We decided against including the actual number last week, to avoid encouraging users to play a numbers game.

Trying to explain the reason/logic behind the 100 limit in the flash message isn't really helpful for users, is it?

DilwoarH:
100 seems like a low enough number that indirectly encouraging a numbers games isn't that big of a deal in my opinion. Having said that, I don't mind if you are happy. I have updated the limit to 100.

galund:
Very interesting that when we had the same issue with people choosing categories, we ended up deciding that we _did_ have to tell them the number.

I think it might be worth straw-polling some users for "how many services makes for a good shortlist", then we could use that figure in our advice (I'd assume it's much lower than the hard limit we're putting in). Probably we won't get that info before we go live, but it'd be nice to pass it onto the user research team to drop in if they are talking to folk about g-cloud.

app/main/views/g_cloud.py
```diff
@@ -33,8 +33,11 @@
 from app import search_api_client, data_api_client, content_loader
 
 
+PROJECT_END_SEARCH_LIMIT = 100
 PROJECT_SAVED_MESSAGE = Markup("""Search saved.""")
-PROJECT_ENDED_MESSAGE = Markup("""Your search has been ended. You can now download your shortlist.""")
+PROJECT_ENDED_MESSAGE = Markup("""Search ended. You can now download your shortlist.""")
+PROJECT_TOO_MANY_RESULTS_ERROR = Markup("""
+    You have too many results. Choose a category or add filters to refine your search.""")
```
galund: FWIW I don't think we necessarily have to use the `Markup` thing if we aren't actually including any markup :)

galund: Oh, the PROJECT_ thing here wasn't supposed to be a prefix. It makes sense for 'project ended' because that's what ended - these should just be END_SEARCH_LIMIT and TOO_MANY_RESULTS_ERROR

app/main/views/g_cloud.py
```diff
@@ -33,8 +33,11 @@
 from app import search_api_client, data_api_client, content_loader
 
 
+PROJECT_END_SEARCH_LIMIT = 100
```
galund: Can you stick a comment in here that really we should be getting the limit from the API (somehow). Don't propose we do that now, but some kind of note to make sure we keep the two in sync is important for maintainability.

app/templates/direct-award/end-search.html
```diff
@@ -62,7 +62,11 @@ <h2 class="heading-xmedium">After your search has ended</h2>
         method="POST"
         >
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <button class="button-save end-your-search-button">End search and continue</button>
+        <button class="button-save end-your-search-button"
+          {% if disable_end_search_btn %} 
+           disabled="disabled" aria-disabled="true"
```
galund: Not important, but FYI https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme

app/main/views/g_cloud.py
```diff
@@ -35,8 +35,11 @@
 from app import search_api_client, data_api_client, content_loader
 
 
+END_SEARCH_LIMIT = 100  # TODO: This should be done in the API.
 PROJECT_SAVED_MESSAGE = Markup("""Search saved.""")
-PROJECT_ENDED_MESSAGE = Markup("""Your search has been ended. You can now download your shortlist.""")
+PROJECT_ENDED_MESSAGE = Markup("""Search ended. You can now download your search results.""")
```
DilwoarH: @karlchillmaid - is this fine for content?

karlchillmaid: Yes, that's the content that was signed off for the flash banners.

app/main/views/g_cloud.py
```diff
@@ -35,8 +35,11 @@
...
+TOO_MANY_RESULTS_ERROR = Markup("""
```
samuelhwilliams: This should probably be called `TOO_MANY_RESULTS_MESSAGE` rather than `ERROR`. Also the padding on the newline should be removed as it technically inserts 4 spaces at the start of the string (although some/all of these _may_ be stripped by the Markup call).

app/main/views/g_cloud.py
```diff
@@ -443,8 +446,24 @@ def view_project(framework_framework, project_id):
...
+    frameworks_by_slug = framework_helpers.get_frameworks_by_slug(data_api_client)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+    if not is_direct_award_project_accessible(project, current_user.id):
+        abort(404)
+
+    searches = data_api_client.find_direct_award_project_searches(user_id=current_user.id,
+                                                                  project_id=project['id'])['searches']
+    search = list(filter(lambda x: x['active'], searches))[0]
```
samuelhwilliams: We don't need to do this list+filter any more if you pass `active=True` to the `find_direct_award_project_searches` call (assuming you have a recent-ish apiclient on this branch).

app/main/views/g_cloud.py
```diff
@@ -454,7 +473,7 @@ def end_search(framework_framework, project_id):
     if project['lockedAt']:
         abort(400)
 
-    if request.method == 'POST':
+    if request.method == 'POST' and not int(search_count) > END_SEARCH_LIMIT:
```
samuelhwilliams: `int(search_count) <= END_SEARCH_LIMIT` is probably clearer.

app/templates/direct-award/end-search.html
```diff
@@ -65,7 +65,11 @@ <h2 class="heading-xmedium">After your search has ended</h2>
         method="POST"
         >
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <button class="button-save end-your-search-button">End search and continue</button>
+        <button class="button-save end-your-search-button"
+          {% if disable_end_search_btn %} 
+           disabled="disabled"
+          {% endif %}
```
samuelhwilliams: It would be a bit neater and more consistent with elsewhere if this was all inline, i.e.: `<button class="button-save end-your-search-button"{% if disable_end_search_btn %} disabled="disabled"{% endif %}>`

tests/helpers.py
```diff
@@ -162,6 +162,12 @@ def _get_direct_award_project_fixture(**kwargs):
         return project
 
     @staticmethod
+    def _get_direct_award_lock_project_fixture():
+        fixture = BaseApplicationTest._get_fixture_data('direct_award_project_fixture.json')
+        fixture['project']['lockedAt'] = "2017-09-08T14:07:14.716154Z"
```
samuelhwilliams: Let's clean the time up a little, eg `2017-09-08T00:00:00.000000Z`

tests/main/views/test_direct_award.py
```diff
@@ -389,3 +389,45 @@ def test_search_api_urls_convert_to_correct_frontend_urls(self, search_api_url,
...
+    def test_end_search_page_renders(self):
+        self.login_as_buyer()
+
+        res = self.client.get('/buyers/direct-award/g-cloud/projects/1/end-search')
+        assert res.status_code == 200
+
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert len(doc.xpath('//h1[contains(normalize-space(), "End search")]')) == 1
+
+    def test_end_search_page_renders_error_when_results_more_than_999(self):
```
samuelhwilliams: The limit is 100 now, so we need to rename this test and update the return value later on. Also I think it's 100 inclusive (at least that's what your earlier logic validates).

tests/main/views/test_direct_award.py
```diff
@@ -389,3 +389,45 @@ def test_search_api_urls_convert_to_correct_frontend_urls(self, search_api_url,
...
+            "links": {}
+        }
+
+        self._search_api_client_helpers_patch = \
+            mock.patch('app.main.helpers.search_save_helpers.search_api_client', new=search_api_client)
+        self._search_api_client_helpers = self._search_api_client_helpers_patch.start()
+
+        res = self.client.get('/buyers/direct-award/g-cloud/projects/1/end-search')
+        assert res.status_code == 200
+        assert "You have too many results." in res.get_data(as_text=True)
```
samuelhwilliams: You could possibly even import the content variable `TOO_MANY_RESULTS_ERROR/MESSAGE` here so that this test is proofed against any changes in that text.

PR #604, opened 2017-09-12
Format text on cancelled and unsuccessful brief banners to one line

CrystalPea:
This change is an effect of QA-ing process.

Trello: https://trello.com/c/sqfpwtPW/797-1-show-award-status-when-buyer-has-cancelled-procurement

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -52,15 +52,11 @@
 {% elif brief.status == 'cancelled' %}
 <div class="grid-row">
   <div class="column-one-whole">
-    {% set brief_cancelled_message %}
-      The buyer cancelled this opportunity, for example because they no longer have the budget.<br />
-      They may publish an updated version later.
-    {% endset %}
     {%
       with
       type = "temporary-message",
       heading = "This opportunity was cancelled",
-      message = brief_cancelled_message
+      message = "The buyer cancelled this opportunity, for example because they no longer have the budget. They may publish an updated version later."
     %}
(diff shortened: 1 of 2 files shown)


app/templates/brief.html
```diff
@@ -52,15 +52,11 @@
...
-    {% set brief_cancelled_message %}
-      The buyer cancelled this opportunity, for example because they no longer have the budget.<br />
-      They may publish an updated version later.
-    {% endset %}
     {%
       with
       type = "temporary-message",
       heading = "This opportunity was cancelled",
-      message = brief_cancelled_message
+      message = The buyer cancelled this opportunity, for example because they no longer have the budget. They may publish an updated version later.
```
benvand: Does it not need the `<br />` now?
Also If the heading is going to have quote then it would be nice if the message did too. Looks like the message does in other places.

Otherwise looks good

katstevens:
Though saying that, I think you have some failing tests according to Travis?

CrystalPea:
Yes, I forgot to put the messages as strings, I now fixed that 😬 

Also, mobile:
screen shot 2017-09-12 at 11 59 30

app/templates/brief.html
CrystalPea: Thanks, yes, I forgot the quotes at first, now fixed :).

Removing line separators is the main thing about this PR, as all message should be on one line apparently, so no <br /> ☺️

benvand: 👍

PR #608, opened 2017-09-13
709 save search layout tweaks

galund:
The small amount of layout tweaks that we said we'd do in ticket 709, before coming back to this screen in [673](https://trello.com/c/zt49gYL6/673-choose-an-existing-saved-search)

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -1,7 +1,4 @@
-// GOVUK Front-end toolkit styles
-@import "_grid_layout.scss";
-@import "_typography.scss";
-@import "_url-helpers.scss";
+@import "govuk-elements";
 
 // Path to assets for use with file-url()
 $path: "/static/images/";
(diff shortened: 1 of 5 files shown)


app/assets/scss/application.scss
```diff
@@ -1,7 +1,4 @@
-// GOVUK Front-end toolkit styles
-@import "_grid_layout.scss";
-@import "_typography.scss";
-@import "_url-helpers.scss";
+@import "govuk-elements";
```
MuriloDalRi: How do these imports work? Why has it changed?

galund: Previously we imported 'some' of govuk-elements. Now we're importing all of it. I can't really see any reason to selectively import (now that we need more than a few files) as it just makes life more complicated, although probably I should have looked at the generated application.css before and afterwards...

MuriloDalRi: 👍

PR #610, opened 2017-09-14
Update homepage link to new "Become a supplier" page

TheDoubleK:
For this Trello ticket: https://trello.com/c/kMYuzoge/75-create-the-become-a-supplier-page-2
This should NOT BE MERGED until the page in supplier app actually exists:

 - [x] https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/741

When this is merged then these *two* functional tests PRs should be merged at the same time to keep things green:
 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/364
 - [ ] https://github.com/alphagov/digitalmarketplace-functional-tests/pull/365

## New look
screen shot 2017-09-14 at 18 51 00

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -94,11 +94,10 @@
     {% if current_user.role != 'supplier' %}
       <div class="padding-bottom-small">
         <p>
-          <a href="/suppliers/create" class="top-level-link">
-            Create a supplier account
+          <a href="/suppliers/supply" class="top-level-link">
+            Become a supplier
           </a>
         </p>
-        <p>Get updates about opportunities to sell on the Digital Marketplace.</p>
       </div>
     {% endif %}
 
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -294,7 +294,7 @@ class TestHomepageSidebarMessage(BaseApplicationTest):
         sidebar_link_texts = [str(item).strip() for item in sidebar_links]
 
         assert 'View Digital Outcomes and Specialists opportunities' in sidebar_link_texts
-        assert 'Create a supplier account' in sidebar_link_texts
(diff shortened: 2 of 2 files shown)


app/templates/index.html
```diff
@@ -94,11 +94,10 @@
     {% if current_user.role != 'supplier' %}
       <div class="padding-bottom-small">
         <p>
-          <a href="/suppliers/create" class="top-level-link">
-            Create a supplier account
+          <a href="/suppliers/supply" class="top-level-link">
```
samuelhwilliams: I don't think I understand why the URL needs to change. `/suppliers/supply` feels super redundant and I don't think `supply` as a verb really encapsulates what they're doing. The URL doesn't let them supply anything (by itself). They're creating an account, and that's the content they get on the page. Is `create-account` better?

TheDoubleK: The `/suppliers/create` page still exists.  The new page will not always let them create an account - it's the "Become a supplier" page, which might only let them sign up for email notifications if there are no frameworks "coming" or "open" (eg like right now).

I'm not completely sold on the URL - if you can think of something better for this page (see screenshots at https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/741) then let us know.

But Laurence, Constance and Chelsea decided on `/supply` for the ticket, and we have no product person at the moment to OK any proposed changes.

samuelhwilliams: Ah I see! I'll have a think.

TheDoubleK:
@samuelhwilliams please approve :p

app/templates/index.html
risicle: Yeah I wasn't sure about `supply` but I couldn't think of anything better.

risicle:
@TheDoubleK , I've rebased this for you so you can just go ahead when you're available.

PR #611, opened 2017-09-18
Content update for mvp

karlchillmaid:
Content updates for the compare and decide (save your search) MVP

Tickets:
728 - https://trello.com/c/6E2XvQU8/728-content-update-task-list-page
730 - https://trello.com/c/vRgXDTjg/730-content-g-cloud-saved-searches-page
732 - https://trello.com/c/NaT5aJZV/732-content-search-results-page
733 - https://trello.com/c/5cLRkuQp/733-content-end-search-page
734 - https://trello.com/c/igqMgO5b/734-content-download-your-search-results-page


galund:
Are we completely sure about losing the word 'shortlist'. It seemed quite important as a way of distinguishing between the steps in the process, to have 'moved on' from searching to shortlisting?

app/templates/direct-award/end-search.html
```diff
@@ -45,16 +45,19 @@
       <div class="dmspeak">
         <h2 class="heading-xmedium">Before you end your search</h2>
       </div>
-      <p>You should only continue when you have finished searching for cloud services and are ready to start comparing the services you’ve found.</p>
+      <p>You should only continue when you have finished searching for cloud services.</p>
 
       <br>
-      <p>You can’t edit your search after it has ended.</p>
+      <p>You can not edit your search once it has ended.</p>
```
samuelhwilliams: I feel like we generally would opt for 'cannot' or 'can't' before; can you help me understand the difference?

karlchillmaid: Cannot or can not are both correct. The first reads better, so I'll change it.

We're avoiding can't in this instance to reduce the risk of users misreading the negative contraction as can. There's a lot of chat about this in content and accessibility communities at the moment.

karlchillmaid:
@galund Yes, we're dropping shortlist. I discussed with Laura and Jane has 2i-ed.

The definition of what qualifies as a shortlist (or a longlist too) caused some confusion in user testing and drew a lot of feedback internally too. This is tied up with what we call the action that users perform at certain points in the process (review, compare, long or shortlist). We've definitely struggled to get a consensus on the language and process. One outcome of this mission is that the guidance needs to be thoroughly reviewed.

If feedback and research shows that this is more confusing without the word shortlist than it was with it, we can always revisit.

karlchillmaid:
Updated heading content, lines 184-186 in test_direct_award.py to fix broken test.

DilwoarH:
@karlchillmaid - the urls needs to change as well right? https://www.preview.marketplace.team/buyers/direct-award/g-cloud/projects/128/download-shortlist

right now it has `download-shortlist` in the path?

DilwoarH:
rebased branch

DilwoarH:
@karlchillmaid - i've updated the urls

PR #612, opened 2017-09-18
Choose an existing saved search

DilwoarH:
As a G-Cloud buyer

Ticket: https://trello.com/c/zt49gYL6/673-choose-an-existing-saved-search

Still to do:
- ~~Write Tests~~

screen shot 2017-09-18 at 15 20 59

### DIFF
--- a/app/main/helpers/direct_award_helpers.py
+++ b/app/main/helpers/direct_award_helpers.py
@@ -1,2 +1,31 @@
+from app import data_api_client
+from operator import itemgetter
+
+
 def is_direct_award_project_accessible(project, user_id):
     return any([user['id'] == user_id for user in project['users']])
+
+
+def get_direct_award_projects(user_id, return_type="all", sort_by_key=None):
+    projects = data_api_client.find_direct_award_projects(user_id).get('projects', [])
+
+    res = {
+        "open_projects": [],
+        "closed_projects": [],
+    }
+
+    for project in projects:
+        if project['lockedAt'] is None:
+            res['open_projects'].append(project)
+        else:
+            res['closed_projects'].append(project)
+
+    if return_type == "all":
+        if sort_by_key:
+            res['open_projects'].sort(key=itemgetter(sort_by_key))
+            res['closed_projects'].sort(key=itemgetter(sort_by_key))
+        return res
+    else:
+        if sort_by_key:
+            res[return_type].sort(key=itemgetter(sort_by_key))
+        return res[return_type]
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -27,7 +27,7 @@ from ..helpers import framework_helpers
 from ..helpers.search_save_helpers import SearchMeta
 from ..forms.direct_award_forms import CreateProjectForm
 
-from ..helpers.direct_award_helpers import is_direct_award_project_accessible
+from ..helpers.direct_award_helpers import is_direct_award_project_accessible, get_direct_award_projects
 
 from ..exceptions import AuthException
 from app import search_api_client, data_api_client, content_loader
@@ -288,20 +288,12 @@ def saved_search_overview(framework_framework):
     content_loader.load_messages(framework['slug'], ['descriptions', 'urls'])
     framework_short_description = content_loader.get_message(framework['slug'], 'descriptions', 'framework_short')
 
-    projects = data_api_client.find_direct_award_projects(current_user.id).get('projects', [])
-    open_projects = []
-    closed_projects = []
-
-    for project in projects:
-        if project['lockedAt'] is None:
-            open_projects.append(project)
-        else:
-            closed_projects.append(project)
+    projects = get_direct_award_projects(current_user.id)
 
     return render_template(
         'direct-award/index.html',
-        open_projects=open_projects,
-        closed_projects=closed_projects,
+        open_projects=projects['open_projects'],
+        closed_projects=projects['closed_projects'],
         framework=framework,
         framework_short_description=framework_short_description
     )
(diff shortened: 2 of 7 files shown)


app/main/helpers/direct_award_helpers.py
```diff
@@ -1,2 +1,22 @@
+from app import data_api_client
+
+
 def is_direct_award_project_accessible(project, user_id):
     return any([user['id'] == user_id for user in project['users']])
+
+
+def get_direct_award_projects(user_id):
```
galund: This should be called `get_direct_award_projects_by_status` or something like that, and the keys can just be `open` and `locked`?

I also think it's important that we at least acknowledge the `active` flag. I think it would be fine for inactive projects to go into a separate status bucket here e.g. `archived`, i.e. for them to not appear anywhere for now. We don't have a way of unsetting that 'active' flag anyway, so it's not like there should be any, but I can imagine someone deciding to set that flag by hand because 'reasons' in a support scenario :)

app/main/views/g_cloud.py
```diff
@@ -340,54 +332,67 @@ def save_search(framework_framework):
         **build_search_query(request.args, filters.values(), content_manifest, lots_by_slug)
     )
 
+    projects = get_direct_award_projects(current_user.id)['open_projects']
+    projects.sort(key=lambda x: x['name'])
```
galund: `key=itemgetter('name')` is perhaps slightly neater? never quite convinced either way :)

app/main/views/g_cloud.py
```diff
@@ -340,54 +332,67 @@ def save_search(framework_framework):
...
 @direct_award.route('/<string:framework_framework>/projects/create', methods=['POST'])
 def project_create(framework_framework):
     form = CreateProjectForm()
-    if form.validate_on_submit():
-        name = form.name.data
+    name = form.name.data
+
+    if not form.validate_on_submit() or not request.form['save_search_selection']:
+        projects = get_direct_award_projects(current_user.id)['open_projects']
+        projects.sort(key=lambda x: x['name'])
```
galund: There's some duplicate code here; I think I'm planning to roll `project_create` and `save_search` into one view, so don't necessarily worry about doing that before this lands as I can pick that up.

app/templates/direct-award/save-search.html
```diff
@@ -33,36 +33,67 @@
   </div>
   <form method="post" action="{{ url_for('direct_award.project_create', framework_framework=framework_framework)}}" id="createProjectForm">
     <div class="grid-row">
-      <div class="column-two-thirds dmspeak">
+      <div class="column-two-thirds">
```
galund: Are you sure about taking the dmspeak of the container? I thought it needed to be applied to pretty much everything otherwise our reset css just clobbers all the styles on all the elements :(

app/templates/direct-award/save-search.html
```diff
@@ -33,36 +33,67 @@
...
+
+              {% if projects %}
+              
+              <p class="form-block">or</p>
+
+              {% for project in projects %}
+
+              <div class="multiple-choice">
+                <input type="radio" name="save_search_selection" value="{{ project.id }}" {% if request.form.save_search_selection == project.id %}checked="checked"{% endif %}>
+                <label for="exisitng-search">{{ project.name or "No Project Name" }}</label>
```
galund: There's a typo here, but in any event the label needs to be associated with the _specific_ input element, otherwise it won' work.

Potentially the easiest solution is to drop the `for` attribute altogether, and just wrap the `input` inside the `label`.

app/main/views/g_cloud.py
```diff
@@ -340,54 +332,72 @@ def save_search(framework_framework):
...
 
 @direct_award.route('/<string:framework_framework>/projects/create', methods=['POST'])
 def project_create(framework_framework):
     form = CreateProjectForm()
-    if form.validate_on_submit():
-        name = form.name.data
+    name = form.name.data
+    name_is_invalid = request.form['save_search_selection'] == "new_search" and not name
+
+    if not form.validate_on_submit() or not request.form['save_search_selection'] or name_is_invalid:
```
galund: Something about this isn't working for me, in that if I submit the form without doing anything, I'm just getting a 400 but without the form being rendered again.

DilwoarH: Hey - I added HTML5 validation to stop the frontend from posting without the radio button selected but the issue was that the request did not have the radio button set, therefore, it threw the 400 error.

DilwoarH: If you pull down - it should work now. @galund

galund: Yep, the HTML 5 validation works. Is there some validation server-side for browsers that don't support that? I think the idea of putting all of these fields into one WTForm is probably sound, if that's do-able.

app/templates/direct-award/save-search.html
DilwoarH: not sure what the reset css does, but it doesn't seem to affect the styling of the page here.

galund: Sure, but adding a dmspeak wrapper around just the heading doesn't seem particularly more correct. Actually we may not need the dmspeak thing at all, I'm playing around with how the heading styles come out if we stop importing govuk-elements' `reset` _anyway_, so watch this space kinda ting...

app/main/views/g_cloud.py
```diff
@@ -340,54 +332,77 @@ def save_search(framework_framework):
...
+    name_is_invalid = save_search_selection == "new_search" and not name
+
+    if not form.validate_on_submit() or not save_search_selection or name_is_invalid:
+        projects = get_direct_award_projects(current_user.id, 'open_projects', 'name')
+        projects.sort(key=lambda x: x['name'])
+        frameworks_by_slug = framework_helpers.get_frameworks_by_slug(data_api_client)
+        search_meta = SearchMeta(request.form['search_api_url'], frameworks_by_slug)
+
+        if not name:
+            form.name.errors = ["Names must be between 1 and 100 characters"]
```
galund: Presumably this string comes from somewhere in the forms system, and it'd have been vaguely nice to re-use that rather than hard-code it here, but I guess let's not worry for now.

app/templates/direct-award/save-search.html
```diff
@@ -33,36 +33,69 @@
...
+
+              {% endif %}
+        </fieldset>
 
-        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
-        <input type="hidden" name="search_api_url" value="{{ search_api_url }}"/>
-        <input type="submit" class="button-save" name="return_to_overview" value="Save and continue">
+        <br>
+        
+        <fieldset>
```
galund: I don't think it means anything to put a fieldset around a single field (which is what this is, given that two of the fields are hidden!) I'd just remove that.

app/templates/direct-award/save-search.html
```diff
@@ -33,36 +33,69 @@
...
+              {% for project in projects %}
+
+              <div class="multiple-choice">
+                <input type="radio" name="save_search_selection" value="{{ project.id }}" {% if request.form.save_search_selection == project.id %}checked="checked"{% endif %}>
+                <label for="save_search_selection">{{ project.name or "No Project Name" }}</label>
+              </div>
+
+              {% endfor %}
+
+              {% endif %}
```
galund: Too much indentation, please adjust editor :)

tests/main/views/test_direct_award.py
```diff
@@ -108,6 +116,11 @@ def test_save_search_submit_success(self):
         assert res.status_code == 303
         assert urlparse(res.location).path == '/buyers/direct-award/g-cloud/projects/1'
 
+    def test_save_existing_search_submit_success(self):
+        res = self._create_project('some name " foo bar \u2016', '1')
+        assert res.status_code == 303
+        assert urlparse(res.location).path == '/buyers/direct-award/g-cloud/projects/1'
```
galund: Interestingly I think actually this case should be an error: the user shouldn't submit anything in the 'name' field if they are choosing an existing project?

(So there should actually be two test cases here, and an additional validation we forgot about?)

DilwoarH: oh ye that is true

DilwoarH: updated the test to not send name

app/templates/direct-award/save-search.html
DilwoarH: fixed

app/main/views/g_cloud.py
DilwoarH: I couldn't find a way to force it to throw an error. If you know a way I'd be happy to try it out

app/templates/direct-award/save-search.html
```diff
@@ -33,36 +33,68 @@
...
+        <fieldset>
+          <div class="multiple-choice" data-target="new-search">
+            <input type="radio" name="save_search_selection" value="new_search" 
+              {% if request.form.save_search_selection == 'new_search' or not projects %}checked="checked"{% endif %} 
+              required="required">
+            <label for="save_search_selection">Create a new search</label>
+          </div>
+          <div class="panel panel-border-narrow js-hidden" id="new-search">
+            {% if form.name.errors %}
+            <div class="validation-wrapper">
```
galund: The approach I tend to prefer for this is to leave the `div` in place regardless of whether there are errors, and wrap just the attribute/class name in the `if`. That means you don't have to worry about having another `if` at some random further point down the page, which makes the indentation look all horrible too :)

DilwoarH: This part was copied over from the existing code. Can refactor if it's an issue?

PR #614, opened 2017-09-20
722 refactor save search

galund:
* Refactor save search form to avoid exposing Search API URLs;
* fix problem with the initial Save Search button in IE8, which does not support the form-action attribute;
* fix label elements on save search page.

The first of these should have no functional impact; the other two are bug-fixes.

As well as now using a frontend-side (i.e. buyer app) URL (chunk) to represent the search on the 'save search' page, the other refactoring of interest is that the 'create project' view has been combined into the 'save search' view. It already did more than just create projects, and the two views needed to have quite a large chunk of code in common anyway.

https://trello.com/c/ubKd3ztP/709-save-search-and-name-your-procurement-design-and-content
https://trello.com/c/jjVWnK1e/722-security-fixes-for-save-search-feature

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -7,7 +7,12 @@ from flask import url_for
 
 
 def get_lot_from_request(request, all_lots):
-    lot = request.args.get('lot', None)
+    # TODO prefer get_lot_from_args - get rid of this
+    return get_lot_from_args(request.args, all_lots)
+
+
+def get_lot_from_args(args, all_lots):
+    lot = args.get('lot', None)
     return lot if (not lot or lot in all_lots) else None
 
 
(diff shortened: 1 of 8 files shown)


app/main/helpers/search_helpers.py
```diff
@@ -7,7 +7,12 @@
 
 
 def get_lot_from_request(request, all_lots):
-    lot = request.args.get('lot', None)
+    # TODO prefer get_lot_from_args - get rid of this
```
samuelhwilliams: Given this is a simple function, it feels like it would be nice to just rip out the 'old' one rather than put in a TODO that will be quickly ignored.

galund: Agree, I was just in a rush. I can still do that before this lands :)

PR #615, opened 2017-09-22
Download search results

samuelhwilliams:
## Summary
When a G-Cloud buyer has ended their search and locked in their result set, we want them to be able to download a list of the services that match their specifications so they can start to compare those services offline. This PR adds that endpoint and supports CSV and ODS filetypes.

We pull in an updated version of utils, which now contains a generic `DownloadFileView` that can be  (and has been) subclassed in the buyer-frontend to generate the data required for these files.

We also update the SearchMeta class to prevent HTML tags being embedded in the search summary, as we want to present this in the downloaded file.

The URL for the search results download preface page has been changed to allow a better structure. We have `/projects/x/results` for the preface page, and `/projects/x/results/download` for the download endpoint.

## Example file downloads
https://drive.google.com/open?id=0Bwv8xRW8JeF7QXplWm1kQWRldFE (csv)
https://drive.google.com/open?id=0Bwv8xRW8JeF7ZExyNVN0WVhkTEU (ods)

## Blocked by
* [x] Must write tests for DownloadResultsView
* [x] Re-run `freeze-requirements` after the utils PR goes in

## Depends upon
https://github.com/alphagov/digitalmarketplace-api/pull/656
~~https://github.com/alphagov/digitalmarketplace-apiclient/pull/98~~
https://github.com/alphagov/digitalmarketplace-utils/pull/332
~~https://github.com/alphagov/digitalmarketplace-frameworks/pull/473~~

## Other associated PRs
https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/32

## Ticket
https://trello.com/c/igRhA3Bt/711-shortlist-file-download

### DIFF
--- a/app/main/presenters/search_summary.py
+++ b/app/main/presenters/search_summary.py
@@ -1,3 +1,4 @@
+from lxml.html import document_fromstring
 import os
 import yaml
 from collections import defaultdict
@@ -31,7 +32,9 @@ class SearchSummary(object):
             formatted_conjunction = " {} ".format(final_conjunction)
             return formatted_conjunction.join([u', '.join(start), end])
 
-    def __init__(self, results_total, request_args, filter_groups, lots_by_slug):
+    def __init__(self, results_total, request_args, filter_groups, lots_by_slug, include_markup=True):
+        self.include_markup = include_markup
+
         self._lots_by_slug = lots_by_slug
         self._set_initial_sentence(results_total, request_args)
 
@@ -50,7 +53,8 @@ class SearchSummary(object):
                     SummaryFragment(
                         group_id=group_id,
                         filters=filters,
-                        rules=group_rules)
+                        rules=group_rules,
+                        include_markup=self.include_markup)
                 )
 
     def _set_initial_sentence(self, results_total, request_args):
@@ -89,7 +93,13 @@ class SearchSummary(object):
                 map(_get_fragment_string, self.filters_fragments))
             parts.append(SearchSummary.write_list_as_sentence(
                 fragment_strings, u"and"))
-        return Markup(u" ".join(parts))
+
+        marked_up_text = Markup(u" ".join(parts))
+
+        if not self.include_markup:
+            marked_up_text = document_fromstring(marked_up_text).text_content()
+
+        return marked_up_text
 
     def get_starting_sentence(self):
         return u"{}{}{} {}".format(
@@ -188,7 +198,8 @@ class SummaryFragment(object):
     POST_TAG = u'</em>'
     FINAL_CONJUNCTION = u'and'
 
-    def __init__(self, group_id, filters, rules):
+    def __init__(self, group_id, filters, rules, include_markup=True):
+
         self.id = group_id
         self.rules = rules
         self.form = 'singular'
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -1,12 +1,19 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from werkzeug.urls import url_encode, url_decode
+from datetime import datetime
 from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
 from flask_login import current_user
+import inflection
+from werkzeug.urls import url_encode, url_decode
 
-from dmutils.formats import dateformat
+from dmcontent.formats import format_service_price
+from dmcontent.questions import Pricing
+
+from dmutils.ods import A as AnchorElement
+from dmutils.formats import dateformat, DATETIME_FORMAT, utcdatetimeformat
 from dmutils.filters import capitalize_first
+from dmutils.views import SimpleDownloadFileView
 from dmapiclient import HTTPError
 
 from ...main import main, direct_award
@@ -27,6 +34,7 @@ from ..helpers.search_helpers import (
 )
 from ..helpers import framework_helpers
 from ..helpers.search_save_helpers import SearchMeta
+from ..helpers.shared_helpers import get_fields_from_manifest, get_questions_from_manifest_by_id
 from ..forms.direct_award_forms import CreateProjectForm
 
 from ..helpers.direct_award_helpers import is_direct_award_project_accessible, get_direct_award_projects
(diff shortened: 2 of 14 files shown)


app/main/presenters/search_summary.py
```diff
@@ -31,7 +24,16 @@ def write_list_as_sentence(input_list, final_conjunction):
             formatted_conjunction = " {} ".format(final_conjunction)
             return formatted_conjunction.join([u', '.join(start), end])
 
-    def __init__(self, results_total, request_args, filter_groups, lots_by_slug):
+    def __init__(self, results_total, request_args, filter_groups, lots_by_slug, markup_tags=True):
+        self.markup_tags = markup_tags
```
galund: In my head this name isn't obviously boolean, can we have `include_markup` instead maybe?

app/main/presenters/search_summary.py
```diff
@@ -31,7 +24,16 @@ def write_list_as_sentence(input_list, final_conjunction):
...
+
+        self.COUNT_PRE_TAG = '<span class="search-summary-count">' if markup_tags else ''
+        self.COUNT_POST_TAG = '</span>' if markup_tags else ''
+        self.KEYWORDS_PRE_TAG = '<em>' if markup_tags else ''
+        self.KEYWORDS_POST_TAG = '</em>' if markup_tags else ''
+        self.LOT_PRE_TAG = '<em>' if markup_tags else ''
+        self.LOT_POST_TAG = '</em>' if markup_tags else ''
```
galund: This lot feel like they break DRY. To be honest I think I'd have preferred an approach that stripped the tags off when we wanted a non-HTML version (e.g. using `lxml`), rather than having to pepper the existing code with tests.

app/main/views/g_cloud.py
```diff
@@ -488,32 +496,189 @@ def end_search(framework_framework, project_id):
...
+                                                                      project_id=kwargs['project_id'])
+
+    def determine_filetype(self, file_context=None, **kwargs):
+        file_type = self.request.args.get('filetype', '').lower()
+
+        if file_type == 'csv':
+            return SimpleDownloadFileView.FILETYPES.CSV
+
+        elif file_type == 'odf':
+            return SimpleDownloadFileView.FILETYPES.ODS
```
MuriloDalRi: Is it ODF or ODS?

samuelhwilliams: :man_facepalming: ODS.

app/main/presenters/search_summary.py
samuelhwilliams: Currently we only have lxml as a dev dependency. Do we really want to include that in the main app?

galund: Personally I'd be fine with that but I guess you'd better ask around. And if other people disagree that this would be better then by all means ignore me on this.

TheDoubleK:
We added lxml to admin front-end recently. If it's useful for what needs to be done then it's legit I think.

app/main/presenters/search_summary.py
samuelhwilliams: Done. 👍

app/main/presenters/search_summary.py
```diff
@@ -31,7 +25,16 @@ def write_list_as_sentence(input_list, final_conjunction):
             formatted_conjunction = " {} ".format(final_conjunction)
             return formatted_conjunction.join([u', '.join(start), end])
 
-    def __init__(self, results_total, request_args, filter_groups, lots_by_slug):
+    def __init__(self, results_total, request_args, filter_groups, lots_by_slug, include_markup=True):
+        self.include_markup = include_markup
+
+        self.COUNT_PRE_TAG = '<span class="search-summary-count">'
```
galund: Could these go back to being class-level? Makes a bit more sense for constants really.

app/main/views/g_cloud.py
```diff
@@ -490,32 +499,189 @@ def end_search(framework_framework, project_id):
...
 
-        # Indexes for our search API are named by the framework slug they relate to.
-        framework_slug = search_api_client.get_index_from_search_api_url(search['searchUrl'])
-        framework = data_api_client.get_framework(framework_slug)['frameworks']
+    try:
+        search = data_api_client.find_direct_award_project_searches(user_id=current_user.id,
+                                                                    project_id=project['id'],
+                                                                    only_active=True)['searches'][0]
+    except (KeyError, IndexError):
+        abort(400)
```
galund: I guess indexerror could have been a 404 but who cares :)

app/templates/direct-award/results.html
```diff
@@ -55,9 +55,9 @@
     <div class="marketplace-paragraph">
       <p>
         <br>
-        <a href="{{ url_for('direct_award.download_shortlist', framework_framework=framework.framework, project_id=project.id, filetype='odf') }}">Download search results spreadsheet (ODF)</a>
+        <a href="{{ url_for('direct_award.download_results', framework_framework=framework.framework, project_id=project.id, filetype='ods') }}">Download search results spreadsheet (ODS)</a>
         <br>
-        <a href="{{ url_for('direct_award.download_shortlist', framework_framework=framework.framework, project_id=project.id, filetype='csv') }}">Download search results spreadsheet (CSV)</a>
+        <a href="{{ url_for('direct_award.download_results', framework_framework=framework.framework, project_id=project.id, filetype='csv') }}">Download search results spreadsheet (CSV)</a>
```
galund: Still think it should have been a "list" of some sort to make it clear this is _not_ (except indirectly) a set of search results. Mutter mutter :-)

tests/fixtures/direct_award_project_services_fixture.json
```diff
@@ -0,0 +1,181 @@
...
+      "supplier": {
+        "contact": {
+          "email": "test@fixture.com",
+          "name": "Supplier contact name",
+          "phone": "01234 000 000"
+        },
+        "name": "Supplier name"
+      }
+    },
+    {
```
galund: did we really need this many?

samuelhwilliams: Probably not. But I am at the end of my tether with mocks and testing now... please don't make me touch it again for a bit.

samuelhwilliams: Probably not. But I am at the end of my tether with mocks and testing now... please don't make me touch it again for a bit.

PR #616, opened 2017-09-22
736 bug styling on dos opportunities page

galund:
Use a class name common across both G-Cloud and DOS searches for style - fixes record count size.

No longer import all of gov-uk elements - fixes heading size by ensuring that we no longer load the govuk-elements `_reset.scss` stylesheet. (Note that we do still load our own _reset.scss file, which is a bit different!)

https://trello.com/c/DG6GGfib/736-bug-styling-on-dos-opportunities-page


DilwoarH:
Looks good.

PR #617, opened 2017-09-22
Update search result download links

samuelhwilliams:
## Summary
Our links on the download search results page are not consistent with document download links elsewhere on the site. This brings them in line and removes the `(ODF/CSV)` suffix which is now brought before the link name. I've also slightly tweaked the content on what I think might work well, but @karlchillmaid to decide. @JeremyHHY FYI, these are our file download links.

## Before
screen shot 2017-09-22 at 21 35 01

## After
screen shot 2017-09-22 at 21 32 35

### DIFF
--- a/app/templates/direct-award/results.html
+++ b/app/templates/direct-award/results.html
@@ -53,13 +53,26 @@
       </ul>
     </div>
     <div class="marketplace-paragraph">
-      <p>
-        <br>
-        <a href="{{ url_for('direct_award.download_results', framework_framework=framework.framework, project_id=project.id, filetype='ods') }}">Download search results spreadsheet (ODS)</a>
-        <br>
-        <a href="{{ url_for('direct_award.download_results', framework_framework=framework.framework, project_id=project.id, filetype='csv') }}">Download search results spreadsheet (CSV)</a>
-      </p>
       <br>
+      {%
+      with
+      items = [
+          {
+              "title": "Download search results as a spreadsheet",
+              "link": url_for('direct_award.download_shortlist', framework_framework=framework.framework, project_id=project.id, filetype='odf'),
+              "file_type": "ODS",
+              "download": "True"
+          },
+          {
+              "title": "Download search results as comma-separated values",
+              "link": url_for('direct_award.download_shortlist', framework_framework=framework.framework, project_id=project.id, filetype='csv'),
+              "file_type": "CSV",
+              "download": "True"
+          }
+      ]
+      %}
+          {% include "toolkit/documents.html" %}
+      {% endwith %}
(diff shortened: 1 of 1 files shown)


TheDoubleK:
I think the only publicly shown document downloads are on the G-Cloud service pages, e.g. on the right-hand side here: https://www.digitalmarketplace.service.gov.uk/g-cloud/services/436997837732555

There are several more in the supplier account for suppliers on a framework, but tricker to get a look at right away.

TheDoubleK:
Seems there may be some inconsistency around this...

During applications open the file links have the file type thing:
screen shot 2017-09-25 at 09 51 57

screen shot 2017-09-25 at 09 53 05

But then once framework is live and we show framework documents to suppliers we don't have it:
screen shot 2017-09-25 at 09 51 41

TheDoubleK:
FWIW, I think the new version here is better than the previous version :+1:

app/templates/direct-award/download-search-results.html
```diff
@@ -53,13 +53,26 @@
...
+      with
+      items = [
+          {
+              "title": "Download search results as a spreadsheet",
+              "link": url_for('direct_award.download_shortlist', framework_framework=framework.framework, project_id=project.id, filetype='odf'),
+              "file_type": "ODS",
+              "download": "True"
+          },
+          {
+              "title": "Download search results as comma-seperated values",
```
galund: Spelling: sep*a*rated

karlchillmaid: 👍

galund:
FWIW I think "Download search results as" is repetitive and therefore obscures the difference between the links. The context / heading means the user knows what these links will be, if they are obvious, so I'd be tempted to just have "Spreadsheet" and "Comma-separated values data file" or something like that.

samuelhwilliams:
This is where I saw the inconsistency (because I was using this page to make sure I wasn't breaking the briefs download format 🙂)

screen shot 2017-09-25 at 10 14 11

app/templates/direct-award/download-search-results.html
samuelhwilliams: :man_facepalming: I get that wrong all the time.

samuelhwilliams:
Thoughts @galund @karlchillmaid @TheDoubleK?

samuelhwilliams:
Thoughts @JeremyHHY?

JeremyHHY:
This looks good.
Are these icons contain alt text for screen readers?

samuelhwilliams:
Yes, the filetypes get read out by screenreaders. @JeremyHHY

PR #619, opened 2017-09-26
Feedback forms reset

galund:
Take 2 feedback forms. The same, but with CSRF.

screen shot 2017-09-26 at 10 38 26

screen shot 2017-09-26 at 10 40 03

Depends on https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/377

https://trello.com/c/Uak7y047/706-feedback-forms


config.py
```diff
@@ -74,6 +74,15 @@ class Config(object):
         'g-cloud-5': 'g-cloud-6',
     }
 
+    DM_FEEDBACK_FORM = {
```
samuelhwilliams: Assuming the feedback form is available across the site, do we have a shared config place we could put this?

galund: AFAIK, we don't have any config that's shared between the apps.

But in this case only the buyer app will need that particular config, because there's only one submission endpoint, and it's here. Only the template will end up being shared.

samuelhwilliams: Oh of course. Cool.

app/main/views/feedback.py
```diff
@@ -0,0 +1,32 @@
...
+@main.route('/feedback', methods=["POST"])
+def send_feedback():
+    feedback_config = current_app.config['DM_FEEDBACK_FORM']
+    form_data = MultiDict()
+    for field, google_form_field in feedback_config['fields'].items():
+        form_data.setlist(google_form_field, request.form.getlist(field))
+
+    result = requests.post(feedback_config['uri'], list(form_data.items(multi=True)))
+    if result.status_code != 200:
+        raise ServiceUnavailable('Google forms submission problem (status %d)'.format(result.status_code))
```
samuelhwilliams: Are we definitely happy using Google forms as the submission method?

Either way I'd probably take Google out of the error message and have it just say something more generic like `There was a problem submitting your feedback. Please try again later.` (i.e. not calling out Google).

galund: Yeah I think Google Forms is the MVP for this. Once we understand volumes and who tends to need to deal with these things then we look at alternatives.

But good point re the message.

- [ ] update error message

tests/main/views/test_feedback.py
```diff
@@ -0,0 +1,38 @@
...
+    @mock.patch('requests.post')
+    def test_feedback_submission(self, external_requests_post):
+        external_requests_post.return_value.status_code = 200
+        response = self._post()
+
+        assert response.status_code == 303
+        new_page = self.client.get(response.location)
+        assert "Thank you for your message" in new_page.get_data(as_text=True)
+
+    def test_feedback_form_has_csrf(self):
```
samuelhwilliams: If this is from the frontend-toolkit, we shouldn't really test for this here, should we?

galund: Hmm I wondered if you'd notice that. I might try to argue that where the form happens to come from is an implementation detail. But then you'd probably make me duplicate the test into all the frontend apps.

Ho hum. I'll have a look and see if there's anywhere in the toolkit that this test could meaningfully go, but if not I'd still rather have it here than not at all?

samuelhwilliams: Yep, if there's nowhere reasonable for it to go in the toolkit, this is indeed better than nothing. 👍

config.py
```diff
@@ -74,6 +74,15 @@ class Config(object):
...
+        'uri': 'https://docs.google.com/a/digital.cabinet-office.gov.uk/forms/d/e/1FAIpQLSc-uXv-4VqGBipDwPhJFDpET5UrHnJpsJ5FFTn4-MBAfKeOPg/formResponse',  # noqa
```
TheDoubleK: Is there just one of these forms for all environments?  Or is this config overridden somewhere for prod?

MuriloDalRi: I think we had just one form for all environments, which we should probably change.

TheDoubleK:
Yeah we should probably have at least one other form exclusively for Prod, and store its URI as a "secret".  I just visited this form directly and put some random stuff in.

But for the time being this looks good, all working for me OK.
:+1:

config.py
samuelhwilliams: _MVP_ :trollface: 

But I kinda agree it would be nice to have a separate form for prod and not have the URL baked into the frontends.

tests/main/views/test_feedback.py
risicle: I think it's a good habit to test for things you "expect" from an underlying component. The only reason not to is if it's superfluously laborious. I don't believe in this strict classification of tests into "unit tests" and "integration tests". It's important to check that when using a lib/API, you've actually got the right end of the stick over how things work.

config.py
galund: Yeah, we can do that if we really prefer. Bear in mind the page URI is sent in the form response, so it's super-easy for us to filter by environment if we want to. Easier - dare I say it - than the Notify configuration *by some good measure* :-D

galund:
@TheDoubleK I don't think these URLs are secret, really. I mean, anyone will be able to go online and submit the production form anyway, so in some ways the preview form is actually more secret than the production one... at which point you can see why I don't think it makes much difference!

(Bear in mind these URLs definitely don't allow people to read others' posts, just submit a new copy of the form.)

tests/main/views/test_feedback.py
risicle: (it also will alert you to breakage through api changes)

app/main/views/feedback.py
```diff
@@ -0,0 +1,35 @@
...
+        raise ServiceUnavailable('There was a problem submitting your feedback. Please try again later.')
+
+    came_from = url_parse(request.form['uri'])
+    # strip netloc and scheme as we should ignore attempts to make us redirect elsewhere
+    replaced = came_from._replace(scheme='', netloc='')
+
+    flash(Markup(
+        """Thank you for your message. If you have more detailed feedback, please email
+        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
+        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
```
risicle:
```
https://airtable.com/shrkFM8L6Wfenzn5Q
```
Not worth making this configurable?

galund: I reckon when it changes the content here will change too. Certainly the content should at least be reviewed if the URL changes, so having them in the same spot seems ok to me.

PR #622, opened 2017-10-02
Adds in validation message for older browsers which do not have HTML5…

DilwoarH:
Adds in validation message for older browsers which do not have HTML5 validation turned on.

Ticket: https://trello.com/c/zt49gYL6/673-choose-an-existing-saved-search

screen shot 2017-10-02 at 12 43 38

### DIFF
--- a/app/templates/direct-award/save-search.html
+++ b/app/templates/direct-award/save-search.html
@@ -43,8 +43,16 @@
         <div class="dmspeak">
           <h2 class="heading-xmedium">Choose where to save your search</h2>
         </div>
+        {% if save_search_selection_error %}
+        <p class="validation-message" id="error-save-search-selection">
+          {{ save_search_selection_error }}
+        </p>
+        {% endif %}
         
         <fieldset>
+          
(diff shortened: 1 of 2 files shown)


app/templates/direct-award/save-search.html
```diff
@@ -43,8 +43,16 @@
...
         </div>
+        {% if save_search_selection_error %}
+        <p class="validation-message" id="error-save-search-selection">
+          {{ save_search_selection_error }}
+        </p>
+        {% endif %}
         
         <fieldset>
+          {% if save_search_selection_error %}
+          <div class="validation-wrapper">
```
galund: Can we have the `<div>` always be in the document, but the class added only if the condition is met? This means you only need one `{% if %}` block.

CrystalPea: Ahh, that could be quite smart if possible. Do you think it would impact on reliability?

DilwoarH: @galund - this has been changed for you.

PR #624, opened 2017-10-02
Layout issues of save button when using mobile

DilwoarH:
Ticket: https://trello.com/c/vWThokqK/740-layout-issues-of-save-button-when-using-mobile

Old:
screen shot 2017-10-02 at 16 09 38

New:
screen shot 2017-10-02 at 16 09 57

### DIFF
--- a/app/assets/scss/_search-page.scss
+++ b/app/assets/scss/_search-page.scss
@@ -16,7 +16,8 @@
   margin: 0px 0px 20px;
 
   .save-summary-text {
-    @include core-16($line-height: (36/16));
+    @include core-16;
+    padding: 7px 15px;
     @include media(mobile) {
       margin-bottom: 10px;
     }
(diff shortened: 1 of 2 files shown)


app/assets/scss/_search-page.scss
```diff
@@ -12,23 +12,20 @@
...
-  min-height: 34px;
-  margin-bottom: 20px;
+  padding-top: 20px;
+  margin: 0px 0px 20px;
 
   .save-summary-text {
-    display: inline-block;
-    font-size: 16px;
-    padding: 5px;
+    @include core-16($line-height: (36/16));
```
samuelhwilliams: This line height looks super weird with a smaller resolution.

screen shot 2017-10-04 at 13 36 15

DilwoarH: This is caused by the `column-three-quarters` and `column-one-quarter` classes. Not sure how much we can do about that. Do you have any suggestions?

DilwoarH: I can add additional CSS for that size screen?

DilwoarH: I have added in padding instead.

PR #626, opened 2017-10-04
Prevent access to supplier pages if no services on framework

samuelhwilliams:
## Summary
It's possible to get to a supplier page via `/g-cloud/suppliers/<supplier_id>` even if the supplier is not on the G-Cloud framework. This will prevent accessing supplier pages from the supplier A-Z if said supplier doesn't have any services live on G-Cloud.

I've hardcoded the framework name 'g-cloud' because the URL already hardcodes it, and it's not necessarily a simple thing to make the URL take any known framework.

## Ticket
https://trello.com/c/a78tWZAC/19-vulnerability-in-g-cloud-supplier-a-z

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -60,15 +60,22 @@ def suppliers_list_by_prefix():
 def suppliers_details(supplier_id):
     supplier = data_api_client.get_supplier(supplier_id=supplier_id)["suppliers"]
 
-    first_character_of_supplier_name = supplier["name"][:1]
-    if is_alpha(first_character_of_supplier_name):
-        prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
-    else:
-        prefix = u"other"
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    live_framework_names = [f['name'] for f in all_frameworks if f['status'] == 'live' and f['framework'] == 'g-cloud']
 
-    return render_template(
-        'suppliers_details.html',
-        supplier=supplier,
-        prefix=prefix,
-        gcloud_framework_description=get_framework_description(data_api_client, 'g-cloud'),
-    )
+    if any(supplier.get('service_counts', {}).get(framework_name, 0) > 0 for framework_name in live_framework_names):
+        first_character_of_supplier_name = supplier["name"][:1]
+        if is_alpha(first_character_of_supplier_name):
+            prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
+        else:
+            prefix = u"other"
+
+        return render_template(
+            'suppliers_details.html',
+            supplier=supplier,
+            prefix=prefix,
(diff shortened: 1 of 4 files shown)


app/main/views/suppliers.py
```diff
@@ -60,15 +60,24 @@ def suppliers_list_by_prefix():
...
+    live_framework_names = [f['name'] for f in filter(lambda f: f['status'] == 'live' and f['framework'] == 'g-cloud',
+                                                      all_frameworks)]
+
+    for framework_name in live_framework_names:
+        if supplier.get('service_counts', {}).get(framework_name, 0) > 0:
+            first_character_of_supplier_name = supplier["name"][:1]
+            if is_alpha(first_character_of_supplier_name):
+                prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
+            else:
+                prefix = u"other"
```
MuriloDalRi: I think we can drop the u with python 3

app/main/views/suppliers.py
```diff
@@ -60,15 +60,23 @@ def suppliers_list_by_prefix():
...
-    first_character_of_supplier_name = supplier["name"][:1]
-    if is_alpha(first_character_of_supplier_name):
-        prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
-    else:
-        prefix = u"other"
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    live_framework_names = [f['name'] for f in all_frameworks if f['status'] == 'live' and f['framework'] == 'g-cloud']
+
+    for framework_name in live_framework_names:
+        if supplier.get('service_counts', {}).get(framework_name, 0) > 0:
```
allait: I think we can replace the external loop with:
`if any(supplier.get('service_counts', {}).get(framework_name, 0) > 0 for framework_name in live_framework_names)`

It's functionally the same, but less indentation and makes the intent clearer and highlights that we only ever expect to take one action here.

app/main/views/suppliers.py
```diff
@@ -60,15 +60,22 @@ def suppliers_list_by_prefix():
 def suppliers_details(supplier_id):
     supplier = data_api_client.get_supplier(supplier_id=supplier_id)["suppliers"]
 
-    first_character_of_supplier_name = supplier["name"][:1]
-    if is_alpha(first_character_of_supplier_name):
-        prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
-    else:
-        prefix = u"other"
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    live_framework_names = [f['name'] for f in all_frameworks if f['status'] == 'live' and f['framework'] == 'g-cloud']
```
TheDoubleK: We could have parameterised the route by framework-framework to give legit DOS suppliers a page.

samuelhwilliams: We could have, and I considered doing that initially (going so far as to implement it (bar tests)), but it does swell the scope of this ticket quite a bit.

PR #634, opened 2017-10-10
Open airtable link in new window per Jeremy.

galund:
https://trello.com/c/Uak7y047/77-feedback-forms

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -29,7 +29,7 @@ def send_feedback():
     flash(Markup(
         """Thank you for your message. If you have more detailed feedback, please email
         <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
-        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
+        <a target="_blank" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
         """))
 
     return redirect(replaced, code=303)
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -29,7 +29,8 @@ def send_feedback():
     flash(Markup(
         """Thank you for your message. If you have more detailed feedback, please email
         <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
-        <a target="_blank" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
+        <a target="_blank" rel="external noopener noreferrer" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take
+        part in our research</a>.
         """))
 
     return redirect(replaced, code=303)


app/main/views/feedback.py
```diff
@@ -29,7 +29,7 @@ def send_feedback():
     flash(Markup(
         """Thank you for your message. If you have more detailed feedback, please email
         <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
-        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
+        <a target="_blank" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
```
allait: Since this is an external link it should also have `rel="noopener noreferrer"`

app/main/views/feedback.py
```diff
@@ -29,7 +29,8 @@ def send_feedback():
     flash(Markup(
         """Thank you for your message. If you have more detailed feedback, please email
         <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
-        <a href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our research</a>.
+        <a target="_blank" rel="noopener noreferrer" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take part in our
```
samuelhwilliams: We should also have `rel=external`?

TheDoubleK:
There's a frontend toolkit component for that...
http://alphagov.github.io/digitalmarketplace-frontend-toolkit/external-link.html

TheDoubleK:
The "Example in new window"

galund:
@TheDoubleK yeah that toolkit pattern is a bit weird because it wraps the <a> in a <div>, which is definitely not markup we can use in the middle of a paragraph.

TheDoubleK:
Oh that's weird. Why would we have that? Does the Jinja version definitely give us a `<div>`? We should fix the pattern rather than having a random set of attributes on our external links.

galund:
Where it was just 'external' then I'm not sure it's worth having an entire macro for that. If we are now saying that consistently all external links need all three link types, then yeah I'm more up for encapsulating those in a macro.

Obviously ripping out the div is very much not-backward-compatible though :/

TheDoubleK:
Or could be backward-compatible if, say, it isn't actually used anywhere at the moment, which is a possibility.

galund:
Yeah, it's being used, in places where some block-level el will be required...
```
digitalmarketplace-admin-frontend/app/templates/suppliers/view_signed_agreement.html:71
digitalmarketplace-buyer-frontend/app/templates/_service_meta.html:12,37
digitalmarketplace-buyer-frontend/app/templates/direct-award/end-search.html:82
```
The most recent of those (end-search.html) doesn't even seem to be using it for a genuinely external link, so I guess I need to raise a ticket about that :-(

galund:
(Have re-opened <https://trello.com/c/gs2I3Cj2/88-end-your-search>)

PR #635, opened 2017-10-11
Fixes breadcrumbs and content

DilwoarH:
As a G-Cloud buyer

Ticket: https://trello.com/c/zt49gYL6/74-choose-an-existing-saved-search

screen shot 2017-10-11 at 12 27 18

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -365,7 +365,7 @@ def save_search(framework_framework):
             form.name.errors = ["Names must be between 1 and 100 characters"]
 
         if not save_search_selection and request.method == "POST":
-            save_search_selection_error = "Please choose where to save your search."
+            save_search_selection_error = "Please choose where to save your search"
         else:
             save_search_selection_error = None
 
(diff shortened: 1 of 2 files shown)


app/main/views/g_cloud.py
```diff
@@ -365,7 +365,7 @@ def save_search(framework_framework):
             form.name.errors = ["Names must be between 1 and 100 characters"]
 
         if not save_search_selection and request.method == "POST":
-            save_search_selection_error = "Please choose where to save your search."
+            save_search_selection_error = "Please choose where to save your search"
```
samuelhwilliams: Can we refactor this so that the content is in the template? Ideally as much content as possible should be there rather than in code files.

samuelhwilliams: Can we refactor this so that the content is in the template? Ideally as much content as possible should be there rather than in code files.

DilwoarH: yeah we can do

DilwoarH: I have reviewed this and in another PR, this error message has been used in multiple places inside the template. Therefore I think this is still the best place to have the content.

DilwoarH: https://projects.invisionapp.com/share/QUDX63O3K#/screens/258002852

samuelhwilliams: Hmm, okay, that's probably sensible enough then. 👍

PR #636, opened 2017-10-11
Add opportunity data download to the opportunity list page

CrystalPea:
- Add 'Opportunity Data' section to opportunities list page under the filtering section
- Style it using existing front-end patterns
- Add an G-Analytics event to track the downloads

Trello ticket: https://trello.com/c/CN4daMQA/23-add-opportunity-data-download-to-the-opportunity-list-page
screen shot 2017-10-12 at 10 48 01

### DIFF
--- a/app/assets/javascripts/analytics/_events.js
+++ b/app/assets/javascripts/analytics/_events.js
@@ -92,12 +92,13 @@
   GOVUK.GDM.analytics.events = {
     'supplierListDownload': function (e) {
       var linkClick = new LinkClick(e);
-
-      if ((linkClick.category() === 'download') && (linkClick.fileType() === 'csv')) {
-        GOVUK.analytics.trackEvent('download', 'csv', {
-          'label': downloadLinkLabel(linkClick),
-          'transport': 'beacon'
-        });
+      if (downloadLinkLabel(linkClick)) {
+        if ((linkClick.category() === 'download') && (linkClick.fileType() === 'csv')) {
+          GOVUK.analytics.trackEvent('download', 'csv', {
+            'label': downloadLinkLabel(linkClick),
+            'transport': 'beacon'
+          });
+        }
       }
     },
     trackEvent: function (e) {
--- a/app/templates/search/_opportunity_data.html
+++ b/app/templates/search/_opportunity_data.html
@@ -1,0 +1,5 @@
+<h2 class='opportunity-data-header'>Opportunity data</h2>
+<p class='opportunity-data-description'>Download data about closed opportunities including who won and how much the contract was worth or if the contract was never awarded and why.<p>
+<a class='opportunity-data-link' href="https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/test.csv">
+  You can download new data every week.
+</a>
--- a/app/templates/search/_opportunity_data.html
+++ b/app/templates/search/_opportunity_data.html
@@ -1,5 +1,21 @@
-<h2 class='opportunity-data-header'>Opportunity data</h2>
-<p class='opportunity-data-description'>Download data about closed opportunities including who won and how much the contract was worth or if the contract was never awarded and why.<p>
-<a class='opportunity-data-link' href="https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/test.csv">
-  You can download new data every week.
-</a>
+<div class="dmspeak">
+  <h2 class="heading-small" id="opportunity-data-header">Opportunity data</h2>
+  <p id="opportunity-data-description">Download data about closed opportunities including award status. This will be updated daily.<p>
+</div>
+{%
+  with
+  items = [
+      {
+          "title": "Download data",
+          "link": "https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/opportunity-data.csv",
+          "file_type": "CSV",
+          "download": "True",
+          "analytics": "trackEvent",
+          "analytics_category": "opportunity-data-csv",
+          "analytics_action": "download CSV",
+          "analytics_label": "Opportunity Data CSV"
+      }
+    ]
+%}
+    {% include "toolkit/documents.html" %}
+{% endwith %}
--- a/spec/javascripts/unit/AnalyticsSpec.js
+++ b/spec/javascripts/unit/AnalyticsSpec.js
@@ -109,6 +109,19 @@ describe("GOVUK.Analytics", function () {
       }]);
     });
 
+    it('sends an event requested via html attributes on example of Oportunity Data download', function() {
+      $(document.body).append('<a id="opportunity-data" data-analytics="trackEvent" data-analytics-category="opportunity-data-csv" data-analytics-action="download CSV" data-analytics-label="Opportunity Data CSV" href="https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/opportunity-data.csv">Download data</a>');
+      GOVUK.GDM.analytics.events.init();
+      $('#opportunity-data').click();
+      expect(window.ga.calls.first().args).toEqual(['send', {
+        'hitType': 'event',
+        'eventCategory': "opportunity-data-csv",
+        'eventAction': "download CSV",
+        'eventLabel': "Opportunity Data CSV",
+        'transport': 'beacon'
+      }]);
+    });
+
     it('sends the right event when a list of user research labs download link is clicked', function() {
       spyOn(GOVUK.GDM.analytics.location, "pathname")
         .and
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -1303,7 +1303,7 @@ class TestCatalogueOfBriefsPage(BaseApplicationTest):
         assert res.status_code == 200
         document = html.fromstring(res.get_data(as_text=True))
 
-        header = document.xpath("//h2[@class='opportunity-data-header']")[0].text
+        header = document.xpath("//h2[@id='opportunity-data-header']")[0].text
 
         assert "Opportunity data" in header
 
@@ -1312,32 +1312,29 @@ class TestCatalogueOfBriefsPage(BaseApplicationTest):
         assert res.status_code == 200
         document = html.fromstring(res.get_data(as_text=True))
 
-        description = document.xpath("//p[@class='opportunity-data-description']")[0].text
-        expected_description = (
-        "Download data about closed opportunities including who won"
-        + " and how much the contract was worth or if the contract was never awarded and why."
-        )
+        description = document.xpath("//p[@id='opportunity-data-description']")[0].text
+        expected_desc = "Download data about closed opportunities including award status. This will be updated daily."
 
-        assert expected_description in description
+        assert expected_desc in description
 
     def test_opportunity_data_link_text_visible_on_catalogue_page(self):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities')
         assert res.status_code == 200
         document = html.fromstring(res.get_data(as_text=True))
 
-        link_text = document.xpath("//a[@class='opportunity-data-link']")[0].text
+        link_text = document.xpath("//a[@class='document-link-with-icon']")[0].text_content()
 
-        assert "You can download new data every week." in link_text
+        assert "Download data" in link_text
 
     def test_opportunity_data_file_url_correct_on_catalogue_page(self):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities')
         assert res.status_code == 200
         document = html.fromstring(res.get_data(as_text=True))
 
-        link = document.xpath("//a[@class='opportunity-data-link']")[0].values()
+        link = document.xpath("//a[@class='document-link-with-icon']")[0].values()
         expected_link = (
             "https://assets.digitalmarketplace.service.gov.uk"
-            + "/digital-outcomes-and-specialists-2/communications/data/test.csv"
+            + "/digital-outcomes-and-specialists-2/communications/data/opportunity-data.csv"
         )
 
         assert expected_link in link
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -1298,6 +1298,50 @@ class TestCatalogueOfBriefsPage(BaseApplicationTest):
         ss_elem = document.xpath("//p[@class='search-summary']")[0]
         assert self._normalize_whitespace(self._squashed_element_text(ss_elem)) == "6 results"
 
+    def test_opportunity_data_header_visible_on_catalogue_page(self):
+        res = self.client.get('/digital-outcomes-and-specialists/opportunities')
+        assert res.status_code == 200
+        document = html.fromstring(res.get_data(as_text=True))
+
+        header = document.xpath("//h2[@class='opportunity-data-header']")[0].text
+
+        assert "Opportunity data" in header
+
+    def test_opportunity_data_description_visible_on_catalogue_page(self):
+        res = self.client.get('/digital-outcomes-and-specialists/opportunities')
+        assert res.status_code == 200
+        document = html.fromstring(res.get_data(as_text=True))
+
+        description = document.xpath("//p[@class='opportunity-data-description']")[0].text
+        expected_description = (
+        "Download data about closed opportunities including who won"
+        + " and how much the contract was worth or if the contract was never awarded and why."
+        )
+
+        assert expected_description in description
+
+    def test_opportunity_data_link_text_visible_on_catalogue_page(self):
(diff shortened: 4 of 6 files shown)


app/assets/javascripts/analytics/_events.js
```diff
@@ -97,19 +97,32 @@
         });
       }
     },
+    'opportunityDataDownload': function (e) {
```
TheDoubleK: I don't really understand everything going on in the tracking events stuff, but it looks to me like the previous `supplierListDownload` event will be fired for *any* csv download?  So if we're introducing a new one here then maybe we need to adjust the previous event to be more choosy about when it thinks the supplier list has been downloaded?

TheDoubleK: I could be wrong though.

app/templates/search/_opportunity_data.html
```diff
@@ -0,0 +1,18 @@
+<br />
```
TheDoubleK: Rather than the `<br />` here it might be better to add some `padding-top` styling to the new `search-page-opportunity-data` class of the section that wraps this (in `application.scss`).

app/templates/search/_opportunity_data.html
```diff
@@ -0,0 +1,18 @@
...
+<div class="dmspeak">
+  <h3 class="heading-small opportunity-data-header">Opportunity data</h3>
```
TheDoubleK: Is this `opportunity-data-header` class here purely for the benefit of the tests?  If so then I'd say drop it and find a different way to identify the element in the test.

spec/javascripts/unit/AnalyticsSpec.js
```diff
@@ -177,6 +177,21 @@ describe("GOVUK.Analytics", function () {
         'transport': 'beacon'
       }]);
     });
+
+    it('sends the right event when an opportunity data download link is clicked', function() {
```
TheDoubleK: Hmm, I guess this test should cover the thing I was worried about earlier.  But it only checks `window.ga.calls.first().args` - should we also check that no other calls are made to GA - e.g. ones that say the supplier list has been downloaded?
This seems like a possible drawback of many of these existing javascript tests - they thest that one thing has happened that we want, but not that a bunch of unwanted things haven't happened along with the one good thing.

TheDoubleK: But I have no real idea about Javascript, so maybe this is all fine.

tests/main/views/test_marketplace.py
```diff
@@ -1298,6 +1298,47 @@ def test_catalogue_of_briefs_page_filtered_all_options_selected(self):
         ss_elem = document.xpath("//p[@class='search-summary']")[0]
         assert self._normalize_whitespace(self._squashed_element_text(ss_elem)) == "6 results"
 
+    def test_opportunity_data_header_visible_on_catalogue_page(self):
```
TheDoubleK: These four tests are really only checking one thing i.e. that the `_opportunity_data` partial has rendered OK.  So I'd be tempted to have these as a single test with four assertions about the different bits of content we care about.

But as things stand there is no way that one test could fail and the other pass - they either all fail or all pass, which to me makes it feel like a test if just one thing.

I don't really think we need to assert every bit of content - we should be able to tweak wording in the template and not have unit tests fail.  I'm fine with all the assertions if you like, but I'd say the one thing we rally care about testing is that the link is there and points to the right place.

TheDoubleK:
As an aside, it's generally nice to have a screenshot of the new thing in the pull-request message :)

tests/main/views/test_marketplace.py
CrystalPea: I had one test initially, but then I refactored it into four tests, because when you have one assertion per test it is much easier to see what broke/changed once something does break or change. And I disagree about all having to either pass or fail. When I was fiddling with various parts of content and html elements yesterday following reviews from content/design and product manager, some tests broke but some didn't, confirming exactly what has changed and what remained the same. I feel that this was very helpful and would be diminished if I bunched them all together.

spec/javascripts/unit/AnalyticsSpec.js
CrystalPea: This is a good point, I will add a test for not receiving more calls than needed and see what happens :).

CrystalPea: My Javascript is also super rusty 😬 . @katstevens was my hero yesterday and and helped me get back to it after 8-9 months break 😆

app/templates/search/_opportunity_data.html
CrystalPea: Ahh, that opens a good question: what is a good reason to make a class? Is giving a specific name to an element a good reason? Or does it need to introduce some new .scss?

app/templates/search/_opportunity_data.html
CrystalPea: Are you ok with explaining why it's preferable for my benefit? Also, would that also need to be done in frontend-toolkit?

app/assets/javascripts/analytics/_events.js
CrystalPea: I will test it :)

CrystalPea: Ok, wrote tests, it turned out the old event tried to fire for opportunity data download and threw an error because the correct path was needed to create event label. To avoid this I have added and URL check before the event tries to fire, the same as new event has. This passes the tests and hopefully should solve the problem :).

app/templates/search/_opportunity_data.html
TheDoubleK: Someone more frontend-y might be able to explain better, but I think the argument is that HTML markup should always be semantic - i.e. relate to the logical structure of the page, and not just how it's displayed - so `<br>` is used when a line needs to break in the middle of a paragraph - like when displaying an address, or poetry, something like that.  Spacing between paragraphs or sections on the page is a purely visual thing - there is no "meaning" to this space other than the way it should look on a particular display.  So in that case it is preferable to use CSS styles.

This is partly to do with accessibility - for example, screen readers do their best to interpret  HTML structure on the page but will (I think) generally ignore CSS styles.  So it's best to keep the HTML structure as neat and logical as possible, rather than clutter it with elements that are only there for page layout.

app/templates/search/_opportunity_data.html
TheDoubleK: I don't have strong opinions on this particularly, again some deep front-end philosophy is involved...   I'm pretty sure that if it's giving a name to an element then it should be an ID rather than a class.  Classes should be applicable to possibly multiple elements on a single page.  I also think there's no point in creating a class unless it has some use - so if there's no associated style then it's a waste of everyone's bandwidth to download the unnecessary markup.

tests/main/views/test_marketplace.py
TheDoubleK: It's just as easy to see what broke when you have multiple assertions in a test - you can add an error message for each assertion, e.g. 
```
assert "this" in some_list, "This was not in the list"
assert x > 20, "x was less than 20"
assert expected_desc in description, "Could not find expected description in the page"
```

And then you get a nice message telling you exactly what has gone wrong.  And this will save precious milliseconds of test run time loading exactly the same page four times over :)

But fine if you want to leave it as four.

app/assets/javascripts/analytics/_events.js
```diff
@@ -88,13 +88,31 @@
...
-      var linkClick = new LinkClick(e);
+      var currentURL = GOVUK.GDM.analytics.location.pathname();
 
-      if ((linkClick.category() === 'download') && (linkClick.fileType() === 'csv')) {
-        GOVUK.analytics.trackEvent('download', 'csv', {
-          'label': downloadLinkLabel(linkClick),
-          'transport': 'beacon'
-        });
+      // check we're on the right page
+      if (currentURL.match('\/buyers\/frameworks\/([a-z\-]+)\/requirements\/([a-z\-]+)\/*(\d+)*')) {
```
TheDoubleK: I'm not sure this regex will work e.g. `([a-z\-]+)` will not match `digital-outcomes-and-specialists-2`.

TheDoubleK: And with a long regex like this it would be good to have a comment next to it that explains exactly what the regex is supposed to match, so that anyone looking at it later doesn't have to figure it out for themselves.

e.g. 
```
// check we're on a "start" page for building requirements
// eg. /buyers/frameworks/digital-outcomes-and-specialists-2/requirements/user-research-participants
...
```

Actually, is that the thing this is supposed to track?
The regex you have has digits at the end... is it the csv of brief responses that suppliers download?

TheDoubleK: We should double check exactly what `supplierListDownload` means.  Unless you're sure?

tests/main/views/test_marketplace.py
TheDoubleK: Looking at the file history I think it is what I originally thought - the DOS list of suppliers csv downloads from the requirements start pages.  I think..

app/assets/javascripts/analytics/_events.js
```diff
@@ -88,13 +88,31 @@
 
   GOVUK.GDM.analytics.events = {
     'supplierListDownload': function (e) {
-      var linkClick = new LinkClick(e);
+      var currentURL = GOVUK.GDM.analytics.location.pathname();
```
DilwoarH: you should use:

`var currentURL = window.location.pathname;`

app/assets/javascripts/analytics/_events.js
```diff
@@ -88,13 +88,31 @@
...
+          GOVUK.analytics.trackEvent('download', 'csv', {
+            'label': downloadLinkLabel(linkClick),
+            'transport': 'beacon'
+          });
+        }
+      }
+    },
+    'opportunityDataDownload': function (e) {
+
+      var currentURL = GOVUK.GDM.analytics.location.pathname();
```
DilwoarH: you should use:

`var currentURL = window.location.pathname;`

app/assets/javascripts/analytics/_events.js
```diff
@@ -118,10 +136,10 @@
       });
     },
     'init': function () {
-      // TODO: this should be within a specific class and only initialised on the supplier list download page
+      // TODO: this should be within a specific class and only initialised on the supplier list download page (for supplierListDownload)
+      // and the opportunities list page (for opportunityDataDownload)
       // Currently this is running on click of any <a> link!
-      $('body').on('click', 'a', this.supplierListDownload);
-      $('body').on('click', '[data-analytics=trackEvent]', this.trackEvent);
+      $('body').on('click', 'a', this.supplierListDownload).on('click', 'a', this.opportunityDataDownload).on('click', '[data-analytics=trackEvent]', this.trackEvent);
```
DilwoarH: Might look better like this for readability:
```
$('body')
    .on('click', 'a', this.supplierListDownload)
    .on('click', 'a', this.opportunityDataDownload)
    .on('click', '[data-analytics=trackEvent]', this.trackEvent);
```

app/templates/search/_opportunity_data.html
```diff
@@ -0,0 +1,18 @@
...
+{%
+with
+items = [
+    {
+        "title": "Download data",
+        "link": "https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/opportunity-data.csv",
+        "file_type": "CSV",
+        "download": "True"
+    }
+]
```
DilwoarH: It might be better to add analytics using the following example rather than adding a new function:

https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/639/files#diff-59a410eae02a55237a24954eef533e1fR75

app/assets/javascripts/analytics/_events.js
samuelhwilliams: Also, it can be a lot clearer if you use raw Python strings when working with regex patterns, which stop you having to escape things.

`'\/buyers\/frameworks\/([a-z\-]+)\/requirements\/([a-z\-]+)\/*(\d+)*'`
can be
`r'/buyers/frameworks/([a-z-]+)/requirements/([a-z-]+)/*(\d+)*'`

But as Kev suggests, this regex looks malformed towards the end. Maybe we can go over it?

CrystalPea: Thanks for spotting this, we just fixed it now with Kat :)

CrystalPea:
@DilwoarH 
So if I put below code in my template I do not need 'opportunityDataDownload' function in _events.js?

Could you please confirm? 
 
```
items = [
    {
        "title": "Download data",
        "link": "https://assets.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists-2/communications/data/opportunity-data.csv",
        "file_type": "CSV",
        "download": "True",
        "analytics": "trackEvent",
        "analytics_category": "opportunity-data-csv",
        "analytics_action": "download-csv",
        "analytics_label": "DOS Opportunity Data CSV"
    }
]
```

app/assets/javascripts/analytics/_events.js
CrystalPea: Could you maybe explain why, please? It would be much more useful if you did that, thanks!

katstevens: I'm curious too - we use the `GOVUK.GDM.analytics.location` in several places here and in the briefs FE. There must be some benefit (filtering out test events maybe)?

app/templates/search/_opportunity_data.html
CrystalPea: That was a good explanation, thanks a lot, I discussed it with Sam and Laurence and will add bottom padding to filter section wrapper class "search-page-filters".

app/templates/search/_opportunity_data.html
CrystalPea: In principle I agree with you that id should be used for unique element and this is how I have been doing stuff in my personal projects. Yet since I started working at DM I noticed we don't really use id's, just classes. 

As to giving elements unique names, I find it helps me navigate around the template better. 

Also, this probably is not a good argument, but we do have empty classes already. For example now I wanted to add padding to "search-page-filters" class, but it turned out such class does not exist in frontend-toolkit 😬 .
[edit: I found the class in buyer-frontend in "digitalmarketplace-buyer-frontend/app/assets/scss/_search-page.scss". Is it ok that there is a whole lot of things that is not in frontend-toolkit? 😕 

Very open to changing class names to id names, or alternatively to changing class names to different class names but not sure what those alternative class names could be.

app/templates/search/_opportunity_data.html
CrystalPea: Done now :)

app/assets/javascripts/analytics/_events.js
CrystalPea: Cool, will move this like you suggested, thanks! :)

app/assets/javascripts/analytics/_events.js
DilwoarH: The reason why I would suggest using the native location functions is because you would always get a consistent response. If someone decides to change the function to return something other than what you expect, it would cause your function to break. 

Also this it's just a wrapper class unless you want to modify the value returned. I would recommend just using the native functionality. There are no additional benefits that I can see from the function below. If you want to continue to use it due to the consistency it's fine I guess.

``` 
(function (root) {
  "use strict";

  root.GOVUK.GDM = root.GOVUK.GDM || {};
  root.GOVUK.GDM.analytics = {
    'register': function () {
      var cookieDomain = (root.document.domain === 'www.digitalmarketplace.service.gov.uk') ? '.digitalmarketplace.service.gov.uk' : root.document.domain;
      var universalId = 'UA-49258698-1';

      GOVUK.Analytics.load();
      GOVUK.analytics = new GOVUK.Analytics({
        universalId: universalId,
        cookieDomain: cookieDomain
      });
    },

    // wrapper around access to window.location
    'location': {
      'hostname': function () {
        return root.location.hostname;
      },
      'pathname': function () {
        return root.location.pathname;
      },
      'protocol': function () {
        return root.location.protocol;
      },
      'search': function () {
        return root.location.search;
      }
    }
  };
})(window);
```

DilwoarH:
@CrystalPea 

^^^ that is correct, no function needed.

CrystalPea:
@DilwoarH 
^^^
All changes have now been made. If you could maybe just have a last look at the code to see if it looks right now it would be great.

Also, is there a way for us to write a test to check if the google analytics event for data opportunity download will be fired now?

TheDoubleK:
It would be nice to get that test for the GA event in if possible.  I'm not the best person to help with that.  Other changes all look good to me.

spec/javascripts/unit/AnalyticsSpec.js
```diff
@@ -109,6 +109,19 @@ describe("GOVUK.Analytics", function () {
       }]);
     });
 
+    it('sends an event requested via html attributes on example of Oportunity Data download', function() {
```
CrystalPea: @TheDoubleK Is this the test you were looking for? :)

PR #639, opened 2017-10-12
Data tracking for new flow [Google Analytics]

DilwoarH:
As a member of the Digital Marketplace team
I need to measure how people use the save search feature
so that I can make decisions to iterate and improve it

We want to measure:

- **conversions**: users who save a search and continue to download search results. Goal = 65%

- **number of results**: number of results when users save search vs number of results when search is ended

Ticket: https://trello.com/c/WKt5QsW9/73-data-tracking-for-new-flow-google-analytics
Download file type now recorded as event

Required: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/383

### DIFF
--- a/app/templates/direct-award/results.html
+++ b/app/templates/direct-award/results.html
@@ -71,11 +71,32 @@
           }
       ]
       %}
-          {% include "toolkit/documents.html" %}
+        <ul class='document-list'>
+          {% for item in items %}
+            <li class='document-list-item'>
+              {% if item.file_type %}
+                <a href="{{ item.link }}" class='document-link-with-icon' {% if item.download %}download {% endif %}
+                  data-analytics="trackEvent"
+                  data-analytics-category="Direct Award"
+                  data-analytics-action="Download search results format"
+                  data-analytics-label="{{ item.file_type }}"
+                >
+                  <span class='document-icon'>{{ item.file_type }}<span> document:</span></span>
+                  {{ item.title }}
+                </a>
+              {% else %}
+                <a href="{{ item.link }}" class='document-link'>
+                  {{ item.title }}
+                </a>
+              {% endif %}
+            </li>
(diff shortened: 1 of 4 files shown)


app/templates/direct-award/results.html
```diff
@@ -71,11 +71,32 @@
...
+          {% for item in items %}
+            <li class='document-list-item'>
+              {% if item.file_type %}
+                <a href="{{ item.link }}" class='document-link-with-icon' {% if item.download %}download {% endif %}
+                  data-analytics="trackEvent"
+                  data-analytics-category="Direct Award"
+                  data-analytics-action="Download search results format"
+                  data-analytics-label="{{ item.file_type }}"
+                >
+                  <span class='document-icon'>{{ item.file_type }}<span> document:</span></span>
```
galund: what's the extra ```span``` here for?

DilwoarH: this was from the template, https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/documents.html

Should we just update the toolkit instead with the GA options?

DilwoarH: I didn't update it, just in case it was used in other places (ie. not in Digital Marketplace)

galund: Nothing other than the Digital Marketplace frontends should be using our toolkit.

We ought to be using/updating our toolkit things rather than duplicating them out, if they seem like good patterns and we aren't changing them in significant ways. Do have to check which of our apps are using them and where, before making changes tho.

As to whether the span should be there or not: no unless you can find some CSS that relies on it. (In which case I'd argue it still needs a `class`.)

DilwoarH:
REQUIRED: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/383

PR #640, opened 2017-10-12
Choose an existing saved search - Change Layout and Content

DilwoarH:
As a G-Cloud buyer

Ticket: https://trello.com/c/zt49gYL6/74-choose-an-existing-saved-search

### DIFF
--- a/app/templates/direct-award/save-search.html
+++ b/app/templates/direct-award/save-search.html
@@ -27,10 +27,31 @@
 
 <div class="single-question-page save-search-page">
   <div class="grid-row">
+    <div class="column-one-whole">
+        {% if form.name.errors or save_search_selection_error %}
+        <div class="error-summary" role="alert" aria-labelledby="saving-search-error" tabindex="-1">
+
+          <div class="dmspeak">
+            <h2 class="heading-xmedium">There was a problem saving your search</h2>
+          </div>
+
+          <ul class="error-summary-list">
+              {% if save_search_selection_error %}
+                <li><a href="#select-where-to-save-error">{{ save_search_selection_error }}</a></li>
+              {% elif form.name.errors and not save_search_selection_error %}
+                {% for error in form.name.errors %}<li><a href="#name-your-search-error">{{ error }}</a></li>{% endfor %}
+              {% endif %}
+          </ul>
+    
+        </div>
+        {% endif %}
+    </div>
+  </div>
+  <div class="grid-row">
     <div class="column-two-thirds">
       <header class="page-heading-smaller page-heading-without-breadcrumb">
         <h1>
-          Save your search
+          Choose where to save your search
         </h1>
       </header>
     </div>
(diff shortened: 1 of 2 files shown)


app/main/views/g_cloud.py
```diff
@@ -365,7 +365,7 @@ def save_search(framework_framework):
             form.name.errors = ["Names must be between 1 and 100 characters"]
 
         if not save_search_selection and request.method == "POST":
-            save_search_selection_error = "Please choose where to save your search."
+            save_search_selection_error = "Please choose where to save your search"
```
samuelhwilliams: Can we put this content in the template? Generally we try to keep copy there rather than in the code itself.

app/templates/direct-award/save-search.html
```diff
@@ -23,10 +23,32 @@
...
+
+          <div class="dmspeak">
+            <h2 class="heading-xmedium">There was a problem saving your search</h2>
+          </div>
+
+          <ul class="error-summary-list">
+              {% if save_search_selection_error %}
+                <li><a>{{ save_search_selection_error }}</a></li>
+              {% endif %}
+              {% if form.name.errors and not save_search_selection_error %}
```
samuelhwilliams: This could be an elif of the same if statement?

app/templates/direct-award/save-search.html
```diff
@@ -23,10 +23,32 @@
...
+        </div>
+        {% endif %}
+    </div>
+  </div>
+  <div class="grid-row">
     <div class="column-two-thirds">
       <header class="page-heading-smaller page-heading-without-breadcrumb">
         <h1>
-          Save your search
+            Choose where to save your search
```
samuelhwilliams: Not that it matters particularly but not sure why this has extra indentation - in templates we generally use 2 spaces (for whatever reason).

app/templates/direct-award/save-search.html
```diff
@@ -40,61 +62,63 @@
...
+
+              {% if projects %}
+
+              
+
+              {% for project in projects %}
+
+              <div class="multiple-choice">
+                <input type="radio" name="save_search_selection" value="{{ project.id }}" id="project-{{ project.id }}"{% if request.form.save_search_selection == project.id %}checked="checked"{% endif %}>
+                <label for="project-{{ project.id }}">{{ project.name or "No Project Name" }}</label>
```
samuelhwilliams: Although it should be impossible, it might be better to use `project.id` rather than "No Project Name", to cover the theoretical case where two projects don't have names and would then be indistinguishable.

But we definitely shouldn't be able to have projects without names...

galund: Agree, perhaps "Untitled {{id}}" is vaguely traditional

app/templates/direct-award/save-search.html
```diff
@@ -27,10 +27,31 @@
 
 <div class="single-question-page save-search-page">
   <div class="grid-row">
+    <div class="column-one-whole">
+        {% if form.name.errors or save_search_selection_error %}
+        <div class="error-summary" role="alert" aria-labelledby="saving-search-error" tabindex="-1">
+
+          <div class="dmspeak">
+            <h2 class="heading-xmedium">There was a problem saving your search</h2>
```
galund: I think I'd like to see how this looks, but it seems weird to have essentially a sub-heading _before_ the heading to which it relates. Maybe <h2> is not the right element. This almost feels like a flash message. If this is really the pattern from govuk then of course I'll stand corrected...

DilwoarH: https://govuk-elements.herokuapp.com/errors/#summarise-errors

Hey - this is the template we followed.

galund: Okay, yeah I suspected as much. Cheers.

PR #648, opened 2017-10-17
Fix broken category sub category section on filtering

galund:
Categories tree 'selected filters' must include parent category. This ensures the parent category (implied) selection survives when a filter is selected - we get a hidden field in the filters form.

https://trello.com/c/QPkflfgt/122-broken-category-sub-category-section-on-filtering-issue-exists-on-production

### DIFF
--- a/app/main/presenters/search_presenters.py
+++ b/app/main/presenters/search_presenters.py
@@ -288,8 +288,18 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, url
             content_manifest, framework, aggregations, keys_to_remove, parent_category=category['value'])
 
         if selected_descendants:
-            # If a parentCategory has been sent as a url query param, de-select all other parent categories.
-            if request.values.get('parentCategory', category['value']) != category['value']:
+            # If a parentCategory has been sent as a url query param, and it's this category...
+            # (Note: `get` fallback is for if there are selected descendants but no parent category is set, which is
+            # not expected, but could happen if constructing URLs by hand.)
+            if request.values.get('parentCategory', category['value']) == category['value']:
+                # ... then ensure this choice survives as a hidden field for the filters form
+                parent_category_filter = dict()
+                parent_category_filter['name'] = 'parentCategory'
+                parent_category_filter['value'] = category['value']
+                selected_category_filters.append(parent_category_filter)
+
+            else:
+                # ... but otherwise, don't include this category's children in the tree
                 selected_descendants = []
 
         if directly_selected or selected_descendants:
--- a/tests/main/presenters/test_search_presenters.py
+++ b/tests/main/presenters/test_search_presenters.py
@@ -256,7 +256,7 @@ class TestSearchFilters(BaseApplicationTest):
         with self.app.test_request_context(url):
             selection = build_lots_and_categories_link_tree(framework, lots, category_filter_group, flask.request,
                                                             g9_builder)
-            assert len(selection) == 4  # all -> software -> option2 -> option2.2
+            assert len(selection) == 5  # all -> software -> option2 -> option2.2; option2 as a parent category filter
 
             tree_root = selection[0]
             # check that only siblings of the selected sub-category are shown, and other categories
@@ -268,6 +268,9 @@ class TestSearchFilters(BaseApplicationTest):
             assert category_filters[0]['selected']
             sub_category_filters = category_filters[0]['children']
             assert len(sub_category_filters) == 2
+
+            assert [f for f in selection if f.get('name') == 'parentCategory'] == [
+                {'name': 'parentCategory', 'value': 'option 2'}]
 
     def test_filter_groups_have_correct_default_state(self):
         request = self._get_request_for_params({
--- a/tests/main/presenters/test_search_presenters.py
+++ b/tests/main/presenters/test_search_presenters.py
@@ -206,72 +206,6 @@ class TestSearchFilters(BaseApplicationTest):
         assert search_filters[0]['filters'][1]['name'] == 'booleanExample2'
         assert search_filters[0]['filters'][1]['checked'] is False
 
-    @mock.patch('app.main.presenters.search_presenters.search_api_client', autospec=True)
-    def test_show_lots_and_categories_selection(self, search_api_client):
-        search_api_client.aggregate_services.return_value = _get_g9_aggregations_fixture_data()
-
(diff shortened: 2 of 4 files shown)


app/main/presenters/search_presenters.py
```diff
@@ -288,8 +288,18 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, url
             content_manifest, framework, aggregations, keys_to_remove, parent_category=category['value'])
 
         if selected_descendants:
-            # If a parentCategory has been sent as a url query param, de-select all other parent categories.
-            if request.values.get('parentCategory', category['value']) != category['value']:
+            # If a parentCategory has been sent as a url query param, and it's this category...
+            # (Note: `get` fallback is for if there are selected descendants but no parent category is set, which is
+            # not expected, but could happen if constructing URLs by hand.)
+            if request.values.get('parentCategory', category['value']) == category['value']:
```
risicle: I'm *really* confused over werkzeug's differentiation between `args` `form` and `values` and it seems from a distance to be one of the worst designed things ever.

galund: Me too. So I just looked it up: args is from the GET query string; form is from the POST; values is a combination of the two, e.g. for if your form might be a POST or a GET and you don't much know or care.

I suppose in these presenters, `values` is the right thing, because they don't know whether the view that called them was a POST or a GET. Really, they probably shouldn't be poking around in the request, so the selected parent category should have been passed in. But this was easier at the time, and now doesn't seem like the moment to revisit whether our search presenters are sufficiently decoupled...

risicle: Yeah, the thing I can't figure out is what you're supposed to do if you want to access the POSTed parameters unsullied by any query strings. Anyway, not that relevant here...

galund: That's `request.form` - I think. In other words, that's not an actual form object (because that would introduce a dependency on wtforms or whatever, and flask doesn't do that) - it's just another multidict.

risicle: Well I would think that if the docs didn't do on to say

```
Changed in version 0.9: Previous to Werkzeug 0.9 this would only contain form data for POST and PUT requests.
```

Which I take to mean pulls in the parameters from GET requests...?

galund: For the record, we've established it won't look in the URL/GET params, but regardless of HTTP method it will look in the request body, as long as the content type is `multipart/form-data` or equivalent.

app/main/presenters/search_presenters.py
```diff
@@ -288,8 +288,18 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, url
...
+                # ... then ensure this choice survives as a hidden field for the filters form
+                parent_category_filter = dict()
+                parent_category_filter['name'] = 'parentCategory'
+                parent_category_filter['value'] = category['value']
```
katstevens: I might be missing something obvious here but is there any reason why this isn't written on one line as 
```
parent_category_filter = {'name': 'parentCategory', 'value': category['value']}
```
?

tests/main/presenters/test_search_presenters.py
```diff
@@ -256,7 +256,7 @@ def test_show_lots_and_categories_selection(self, search_api_client):
         with self.app.test_request_context(url):
```
katstevens: This long test feels like it could be split in two (with maybe some shared setup/context)?

app/main/presenters/search_presenters.py
```diff
@@ -288,8 +288,18 @@ def _annotate_categories_with_selection(lot_slug, category_filters, request, url
             content_manifest, framework, aggregations, keys_to_remove, parent_category=category['value'])
 
         if selected_descendants:
-            # If a parentCategory has been sent as a url query param, de-select all other parent categories.
-            if request.values.get('parentCategory', category['value']) != category['value']:
+            # If a parentCategory has been sent as a url query param, and it's this category...
+            # (Note: `get` fallback is for if there are selected descendants but no parent category is set, which is
+            # not expected, but could happen if constructing URLs by hand.)
```
katstevens: Is the fallback scenario unlikely enough that we don't need a test for it? Or is already covered somewhere?

galund: The fallback was already in place: I was just documenting it, as it took me a while to understand why it was there. No, I don't think we test for it - it feels like an edge case, and seeing as it can't happen (I think!), it's probably okay not to test it.

The tests already run both these code paths, anyway, regardless of that fallback.

app/main/presenters/search_presenters.py
galund: Point taken, this was just done in the same style as where we build other "filter dictionaries" - see `search_presenters.build_lots_and_categories_link_tree` for example. Some of those cases it's perhaps more justifiable to use this format as some of the values are the result of relatively long function calls (e.g. see the `root_node['link'] = search_link_builder(...)` line).

I think if I change this I should change them all, and I'm not sure the end result would necessarily be more readable?

app/main/presenters/search_presenters.py
risicle: oh and this is probably clearer if written:

```
if 'parentCategory' not in request.values or request.values.get('parentCategory') == category['value']:
```

app/main/presenters/search_presenters.py
risicle: Or even just

```
selected_category_filters.append({
    'name': 'parentCategory',
    'value': category['value'],
})
```

katstevens: Hmm. Consistency probably wins in this case? Your call @galund...

app/main/presenters/search_presenters.py
galund: You're right that the actual logic becomes more clear, but it calls out the weird case even more, which doesn't make the overall 'normal' flow more clear? Also I'd have to rewrite the comment, which already made my head hurt enough...

app/main/presenters/search_presenters.py
galund: Going with consistency for now if that's ok. I'll bear in mind that ppl prefer `{..}` notation next time I'm writing this kind of thing :)

app/main/presenters/search_presenters.py
risicle: Weird cases are just as important as the "normal" case to me but whatevs

PR #649, opened 2017-10-18
Dh 82 bug saved search overview page lists searches in the wrong chronological order

DilwoarH:
Ticket: https://trello.com/c/jtMEsYq8/82-bug-saved-search-overview-page-lists-searches-in-the-wrong-chronological-order

### DIFF
--- a/app/main/helpers/direct_award_helpers.py
+++ b/app/main/helpers/direct_award_helpers.py
@@ -6,8 +6,8 @@ def is_direct_award_project_accessible(project, user_id):
     return any([user['id'] == user_id for user in project['users']])
 
 
-def get_direct_award_projects(user_id, return_type="all", sort_by_key=None):
-    projects = data_api_client.find_direct_award_projects(user_id).get('projects', [])
+def get_direct_award_projects(user_id, return_type="all", sort_by_key=None, latest_first=None):
+    projects = data_api_client.find_direct_award_projects(user_id, latest_first=latest_first).get('projects', [])
 
     res = {
         "open_projects": [],
@@ -24,6 +24,7 @@ def get_direct_award_projects(user_id, return_type="all", sort_by_key=None):
         if sort_by_key:
             res['open_projects'].sort(key=itemgetter(sort_by_key))
             res['closed_projects'].sort(key=itemgetter(sort_by_key))
+
         return res
     else:
         if sort_by_key:
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -6,6 +6,7 @@ import pytest
 import sys
 from urllib.parse import quote_plus, urlparse
 from werkzeug.exceptions import BadRequest, NotFound
+from dateutil import parser
 
 from dmcontent.content_loader import ContentLoader
 from app import data_api_client, search_api_client, content_loader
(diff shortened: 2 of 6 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -64,6 +65,22 @@ def setup_class(cls):
...
 
+    def test_renders_saved_search_overview(self):
+        self.login_as_buyer()
+        res = self.client.get(self.SAVE_SEARCH_OVERVIEW_URL)
+        assert res.status_code == 200
+
+        result_html = res.get_data(as_text=True)
+        assert "Your saved searches" in result_html
+        assert "Searching" in result_html
+        assert "Search ended" in result_html
```
samuelhwilliams: These tests are really forgiving and don't make it clear to the reader what's been tested. We should more definitively assert that we're checking a breadcrumb and two table headings.

tests/main/views/test_direct_award.py
```diff
@@ -64,6 +65,22 @@ def setup_class(cls):
...
+
+        doc = html.fromstring(result_html)
+        open_date = "Sunday 17 September 2017 at 3:33pm GMT"
+        closed_date = "Tuesday 10 October 2017 at 4:55pm GMT"
+        assert doc.xpath('//*[@id="content"]/div/div/table[1]/tbody/tr[1]/td[2]/span')[0].text == open_date
+        assert doc.xpath('//*[@id="content"]/div/div/table[2]/tbody/tr[1]/td[2]/span')[0].text == closed_date
```
samuelhwilliams: It might be good to group the first of these two asserts under the 'searching' assert above, and the second under the 'search ended' assert - to more clearly represent associations.

I don't know if such a specific xpath assertion is good or bad. It feels a bit fragile. But it's probably fine.

app/main/helpers/direct_award_helpers.py
```diff
@@ -24,6 +24,9 @@ def get_direct_award_projects(user_id, return_type="all", sort_by_key=None):
         if sort_by_key:
             res['open_projects'].sort(key=itemgetter(sort_by_key))
             res['closed_projects'].sort(key=itemgetter(sort_by_key))
+        else:
+            res['closed_projects'].sort(key=itemgetter('lockedAt'), reverse=True)
```
samuelhwilliams: Won't this reverse=True negate the latest_first parameter? We shouldn't need this if we're letting people specify the chronological order directly through a method parameter.

DilwoarH: on the closed projects, we don't show the `Created` column therefore it needed to be reordered based on `lockedAt`.

samuelhwilliams: My bad, yes, makes sense.

samuelhwilliams: I think this method throws me off a bit altogether. Considering it's only used in two places, it feels like it's been abstracted too soon and the required logic could have just sat in the two places it's called. There's some dissonance around sorting closed_projects by lockedAt, I guess, which means it's not instantly apparent that these attributes are one and the same.

Also when we call this function in `save_search`, we pass a sort_by_key of name, and then re-sort the list on name again afterwards. We could take that out at the least. :)

DilwoarH: I could remove the additional sort from here and put it into the function itself. Other than this specific line, everything else is generic.

tests/main/views/test_direct_award.py
DilwoarH: You're right, I'll have a look at making these better.

tests/main/views/test_direct_award.py
DilwoarH: the path gets the first row of the tables so I would leave them as they are I think.

DilwoarH:
@samuelhwilliams - changes have been done

samuelhwilliams:
So I don't think there's a test that actually pins the key part of this ticket - that searches aren't currently listed in the correct chronological order. We should really make sure that's tested.

DilwoarH:
@samuelhwilliams  - i have updated the tests to check the ordering of search.

PR #654, opened 2017-10-24
BUG: Spacing issue on the compare and decide saved search task list page

DilwoarH:
The space between the numeral and the title seems to vary on different items.

1 and 7 look ok, the others look too close together.

Ticket: https://trello.com/c/LHDlR93V/123-bug-spacing-issue-on-the-compare-and-decide-saved-search-task-list-page

Before:
screen_shot_2017-10-20_at_10 09 44

After:
screen shot 2017-10-24 at 11 53 10

### DIFF
--- a/app/assets/scss/_search_overview.scss
+++ b/app/assets/scss/_search_overview.scss
@@ -18,7 +18,7 @@ ol.instruction-list-item-top li {
 
 ol.steps {
   @include bold-27;
-  padding-left: 25px;
+  padding-left: 30px;
 
   @include media( $min-width: $site-width + (2 * (27px + $gutter) ) ) {
     padding-left: 0;
@@ -30,11 +30,11 @@ ol.steps {
   .instruction-list-item-top,
   .instruction-list-item-box,
   .instruction-list-item-action {
-    margin-left: 25px;
+    margin-left: 30px;
   }
 
   .step-number {
-    min-width: 25px;
+    min-width: 30px;
     float: left;
   }
 }
--- a/app/templates/direct-award/view-project.html
+++ b/app/templates/direct-award/view-project.html
@@ -135,7 +135,7 @@
           }
         ] %}
           <li class="instruction-list-item divider">
-            <h2 class="instruction-list-item-body"><span class="step-number" role="presentation">{{ loop.index }}.</span>&nbsp;{{ step.title }}</h2>
+            <h2 class="instruction-list-item-body"><span class="step-number" role="presentation">{{ loop.index }}.</span>{{ step.title }}</h2>
             {% for item in step.content %}
               {% if item.type == 'text' %}
                 <p class="instruction-list-item-top{% for class in item.class %} {{ class }}{% endfor %}">{{ item.text|safe }}</p>


app/assets/scss/_search_overview.scss
```diff
@@ -18,7 +18,7 @@ ol.instruction-list-item-top li {
 
 ol.steps {
```
Wynndow: Why do we have this scss file here in the buyer app? Shouldn't it be in the frontend toolkit and imported?

DilwoarH: I think because we required additional styling. @samuelhwilliams could you confirm?

Wynndow:
I don't know if this is a thing we care about, but the titles for each section seem a bit weird to me on mobile. I'd expect the broken line to continue from the same indent as the beginning of the title, not underneath the number.

screen shot 2017-10-24 at 14 14 23

DilwoarH:
@Wynndow spoke to laura and jeremy - they agree that the mobile issue is a new issue and will get it's own ticket.

DilwoarH:
TICKET CREATED FOR MOBILE ISSUE: https://trello.com/c/z1RNJnQy/54-layout-tweaks-on-saved-search-task-list-page

app/templates/direct-award/view-project.html
```diff
@@ -135,7 +135,7 @@
           }
         ] %}
           <li class="instruction-list-item divider">
-            <h2 class="instruction-list-item-body"><span class="step-number" role="presentation">{{ loop.index }}.</span>&nbsp;{{ step.title }}</h2>
+            <h2 class="instruction-list-item-body"><span class="step-number" role="presentation">{{ loop.index }}.</span>{{ step.title }}</h2>
```
galund: I think whatever the CSS, there ought to be at least one space (of some sort) in the actual markup. For accessibility if nothing else.

galund: (I don't actually see why we expect anything other than the numbers to line up on the left. There shouldn't be a requirement for the text after the numbers to line up.)

app/assets/scss/_search_overview.scss
samuelhwilliams: If it's possible to get it in the frontend toolkit, that'd be great. I don't remember why it was put here specifically. It might even just have been my lack of understanding at the time, or I was in a rush. Not ideal.

If it's easy to re-centralise this, we should.

app/templates/direct-award/view-project.html
samuelhwilliams: We needed to lose the space from the step.title to get it aligned properly with the paragraph below.

We could add a space after the loop.index, but it isn't required. VoiceOver, etc will handle reading it out normally and I can't think of any other accessibility issues it might cause.

app/assets/scss/_search_overview.scss
```diff
@@ -18,7 +18,7 @@ ol.instruction-list-item-top li {
...
   @include bold-27;
-  padding-left: 25px;
+  padding-left: 30px;
```
samuelhwilliams: I'm not sure I understand why we would want to increase this. The briefs-FE is 25px (albeit approximated via `em`s) so we stay consistent with that. It also doesn't help with either of the issues for this PR: getting the step text aligned with the paragraph, or getting the step numbers to have equal spacing.

app/templates/direct-award/view-project.html
galund: hmm, I think the space is required for this to be correct English, but I'm willing to believe it makes little practical difference

DilwoarH:
@samuelhwilliams - this is something I didn't understand. We have this in the frontend toolkit, not sure why it was copied and pasted into the template. Do you know?

http://alphagov.github.io/digitalmarketplace-frontend-toolkit/instruction-list.html

samuelhwilliams:
If we definitely have exactly this in the frontend-toolkit, then you should remove it here. But what is in _search_overview.css (buyer-fe) was taken from _brief_submission.css (briefs-fe), specifically the ol/ul parts, which aren't in the FE toolkit.

If they can easily be centralised, we should do that.

app/assets/scss/_search_overview.scss
DilwoarH: due to an increased size of font used in the direct award task list, 25px is sometimes not enough to make the rows consistent.

samuelhwilliams: Should we instead bring the fontsize back down to whatever the briefs one is to be more consistent then?

galund: +1 on consistency here

PR #658, opened 2017-11-01
Removes download attribute

DilwoarH:


app/templates/direct-award/results.html
```diff
@@ -68,7 +67,6 @@
               "title": "Download search results as comma-separated values",
               "link": url_for('direct_award.download_results', framework_framework=framework.framework, project_id=project.id, filetype='csv'),
               "file_type": "CSV",
-              "download": "True",
```
galund: So if the 'documents' template has a feature that doesn't work, potentially we should be taking it out of there?

DilwoarH: the feature itself works, we may be using it for other external links which we cannot control the headers of.

galund: Hmmm, if it results in untestable code, then we should at least be sticking a comment in there warning people about that?

PR #659, opened 2017-11-02
Allow suppliers to edit submitted application

katstevens:
Trello ticket: https://trello.com/c/kQpbazwt/85-allow-suppliers-edit-submitted-application-if-opportunity-is-still-open

Cosmetic changes to the Buyer FE as part of the 'suppliers editing their Brief applications' story (see prototype https://suppliers-edit-application.cloudapps.digital/application/opportunity):
- the blue 'Apply' link button is now a green submit button (with a GET form action to trigger the link - not great but we do something similar in the Briefs FE when a buyer wants to start a new Brief).
- remove paragraph text, as per prototype 

Old version: 
apply-old
New version:
apply-new

Also some tidying:
- convert some of the hardcoded URLs in the same template to be named external routes.
- fix pytest deprecation warnings (adding `[tool:pytest]` to config and upgrading `pytest-cov`) which means test are now green instead of yellow 🎉 

This PR can be merged independently of the rest of the supplier edit flow, however some functional tests will fail (so don't merge yet!).

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -141,19 +141,15 @@
 {% elif brief.status == 'live' %}
   <div class="grid-row">
     <div class="column-two-thirds">
-      <h2 class="summary-item-heading">Apply for this opportunity</h2>
-      <p class="dmspeak">To apply, you must give evidence for all the essential and nice-to-have skills and experience you have.</p>
-    </div>
-  </div>
-  <div class="grid-row">
-    <div class="column-one-third">
-      {% with
-         url = "/suppliers/opportunities/{}/responses/start".format(brief.id),
-         label = "Apply",
-         advice = True
-      %}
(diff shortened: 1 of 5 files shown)


app/templates/brief.html
```diff
@@ -130,9 +130,8 @@
   <div class="grid-row">
     <div class="column-one-third">
         {% with
-           url = "/suppliers/opportunities/{}/responses/result".format(brief.id),
-           text = "View your application",
-           bigger = True
```
TheDoubleK: No bigger?

katstevens: Nope! It looks awful. Confirmed with Laurence that the non-big link is better.

TheDoubleK: Cool, just checking :)

PR #662, opened 2017-11-07
Update project_id types to strings in preparation for obfuscated IDs

samuelhwilliams:
## Summary
We are obscuring Direct Award project/saved search IDs, so the URL routes need to accept strings rather than integers now. We will redirect the internal IDs to the obfuscated IDs for a transitional period (a week or so).

Haven't added tests because it "just works" and we will rip it out in short order. Tell me if you think I should anyway.

## Ticket
https://trello.com/c/GEziXqcE/34-security-change-direct-award-projects-to-use-randomly-generated-ids-rather-than-sequential

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -51,6 +51,21 @@ TOO_MANY_RESULTS_MESSAGE = Markup("""
     You have too many results. Choose a category or add filters to refine your search.""")
 
 
+# TODO: Temporary redirect for direct_award internal IDs. Remove after 21/11/2017 - SW.
+@direct_award.before_request
+def redirect_internal_ids():
+    if 'project_id' in request.view_args:
+        project_id = request.view_args['project_id']
+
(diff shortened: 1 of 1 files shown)


app/main/views/g_cloud.py
```diff
@@ -51,6 +51,21 @@
...
+    if 'project_id' in request.view_args:
+        project_id = request.view_args['project_id']
+
+        project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+        if project and is_direct_award_project_accessible(project, current_user.id):
+            if project_id.isdigit() and project_id != project['id']:
+                view_args = request.view_args.copy()
+                view_args['project_id'] = project['id']
+
+                return redirect(url_for(request.endpoint, **view_args))
```
risicle: I don't understand how this works.

risicle: Ok now I do

risicle: Neat

samuelhwilliams:
PR re-opened in new branch with less functionality.

PR #663, opened 2017-11-07
Track >100 services error on end your search

DilwoarH:
We want to answer these questions:

- how many users see this error?

- what do users do after they see the error?

- how many services are in the list when users try to end their search?

Ticket: https://trello.com/c/s2Xp5iE3/168-track-100-services-error-on-end-your-search

### DIFF
--- a/app/assets/javascripts/analytics/_virtualPageViews.js
+++ b/app/assets/javascripts/analytics/_virtualPageViews.js
@@ -16,14 +16,23 @@
 
   };
 
-  VirtualPageViews.prototype.sendVirtualPageView = function( url, name ) {
+  VirtualPageViews.prototype.sendVirtualPageView = function( url, name, dimensions ) {
 
-    if ( GOVUK.analytics  && url ) {
-      var urlList = url.split("?");
-      urlList[0]  = urlList[0] + "/vpv";
-      url         = urlList.join("?");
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Looks like all of our apps have individual `app/assets/javascript/analytics` folders... we should probably centralise this to the frontend-toolkit soon. Not one for this obviously as it looks like they're already out of sync, but as an aside.

app/assets/javascripts/analytics/_virtualPageViews.js
```diff
@@ -16,15 +16,24 @@
...
 
-  VirtualPageViews.prototype.sendVirtualPageView = function( url, name ) {
+  VirtualPageViews.prototype.sendVirtualPageView = function( url, name, dimensions ) {
 
-    if ( GOVUK.analytics  && url ) {
-      var urlList = url.split("?");
-      urlList[0]  = urlList[0] + "/vpv";
-      url         = urlList.join("?");
-      GOVUK.analytics.trackPageview(url, name);
+    if ( !GOVUK.analytics || !url ) return false;
```
CrystalPea: I like how you changed it to be a guard statement :)

DilwoarH: :)

PR #664, opened 2017-11-09
Make page level custom dimension for task list

DilwoarH:
To work out if users view different information at different stages of the process

Ticket: https://trello.com/c/6FOMgGFT/138-make-page-level-custom-dimension-for-task-list

Updates to fixture to add in missing "downloadedAt" attribute

### DIFF
--- a/app/assets/javascripts/analytics/_pageViews.js
+++ b/app/assets/javascripts/analytics/_pageViews.js
@@ -17,16 +17,13 @@
           j;
 
       // page level custom dimension
-      var customDimensions = $('meta[name="ga_customDimension"]');
-      for (var i = 0; i < customDimensions.length; i++)
-      {
-        var customDimension = customDimensions[i];
-        var dimensionId = $(customDimension).attr('data-id');
-        var dimensionValue = $(customDimension).attr('data-value');
-
-        if ( !dimensionId || !dimensionValue ) continue;
-        GOVUK.analytics.setDimension(dimensionId, dimensionValue);
-      }
+      $('meta[name="ga_customDimension"]').each(function(index, customDimension){
+          var dimensionId = $(customDimension).attr('data-id');
+          var dimensionValue = $(customDimension).attr('data-value');
+          if ( dimensionId && dimensionValue ) {
+            GOVUK.analytics.setDimension(dimensionId, dimensionValue);
+          }
+      });
 
       // check that we're on the catalogue page for opportunites
       if (pathname === "/digital-outcomes-and-specialists/opportunities") {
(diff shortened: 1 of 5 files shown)


DilwoarH:
Hey @TheDoubleK @galund,

I was wondering if you guys knew a better way of doing this. I have followed the current implementation but I feel like this can be made into a generic solution.

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -16,6 +16,17 @@
           i,
           j;
 
+      // check for direct-award stage
+      if ( /\/buyers\/direct-award\/g-cloud\/projects\/[0-9]*$/.test( pathname ) )
```
galund: We reckon data attributes for this as per the events stuff

DilwoarH:
@galund - I have changed the custom dimensions to be more generic

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -16,6 +16,18 @@
           i,
           j;
 
+      // page level custom dimension
+      var customDimensions = $('meta[name="ga_customDimension"]');
+      for (var i = 0; i < customDimensions.length; i++)
```
galund: Isn't it more idiomatic to iterate using JQuery `each`?

DilwoarH:
@galund - I have updated the loop to be a jquery each

app/assets/javascripts/analytics/_pageViews.js
```diff
@@ -16,6 +16,15 @@
           i,
           j;
 
+      // page level custom dimension
+      $('meta[name="ga_customDimension"]').each(function(index, customDimension){
+          var dimensionId = $(customDimension).attr('data-id');
```
galund: Out of interest, does this still work if you remove the `$(..)`? I'd have thought so, as JQuery should already be passing through Jquery objects when you use `.each`. Doesn't much matter obviously.

One more thing: this code only seems to get called on page load? Does that mean that the custom dimensions can't be attached to virtual page views, and does that matter for our current use case? Maybe it's just a case of noting that limitation somewhere rather than fixing it.

DilwoarH: Regarding the `$(..)` I usually always use the variable around the jquery tags due to some instances it not being translated properly. 

Regarding the code only running on page load, this would be correct. For virtual page views, we have extended the virtual pageview functions already to allow passing custom dimensions through. ref: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/assets/javascripts/analytics/_virtualPageViews.js#L23

PR #666, opened 2017-11-21
Notification banner fix

katstevens:
Fixes bug where the notification banner template was unintentionally using the `content` context variable in the opportunity page (see https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/388 - this PR will fail until that is merged).

The breaking change will affect the Briefs FE only (separate PR to come shortly).


benvand:
👍 Solves 
briefs-banner-bug

PR #667, opened 2017-11-24
Redirect saved search IDs

samuelhwilliams:
## Summary
We are generating random external IDs for Direct Award projects/saved searches, so we need a transitional function to redirect buyer URLs if the ID returned by the API doesn't match the ID provided in the URL.

## Associated PRs
https://github.com/alphagov/digitalmarketplace-api/pull/690

## Ticket
https://trello.com/c/GEziXqcE/34-security-change-direct-award-projects-to-use-randomly-generated-ids-rather-than-sequential

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -51,6 +51,21 @@ TOO_MANY_RESULTS_MESSAGE = Markup("""
     You have too many results. Choose a category or add filters to refine your search.""")
 
 
+# TODO: Temporary redirect for direct_award internal IDs. Remove after external IDs have been out for a week - SW.
+@direct_award.before_request
+def redirect_internal_ids():
+    if 'project_id' in request.view_args:
+        project_id = request.view_args['project_id']
+
+        project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+        if project and is_direct_award_project_accessible(project, current_user.id):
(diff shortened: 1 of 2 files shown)


app/main/views/g_cloud.py
```diff
@@ -51,6 +51,21 @@
     You have too many results. Choose a category or add filters to refine your search.""")
 
 
+# TODO: Temporary redirect for direct_award internal IDs. Remove after external IDs have been out for a week - SW.
+@direct_award.before_request
+def redirect_internal_ids():
+    if 'project_id' in request.view_args:
+        project_id = request.view_args['project_id']
+
+        project = data_api_client.get_direct_award_project(project_id=project_id)['project']
```
idavidmcdonald: So if someone puts an internal ID in the url and we query the API for it, would we not get a 404 and this function would do nothing? Maybe I've misunderstood (in which case a slightly expanded description on the function might help). A unit test could also be a good addition to this, even though I know this is a temporary thing.

Could we po

samuelhwilliams: If someone puts an internal ID, they won't get a 404 because the API currently allows internal IDs to be used to retrieve resources.

idavidmcdonald: Ah yes, thanks

PR #668, opened 2017-11-27
Pull in v 9.3.2 of frontend toolkit

Wynndow:
For this trello card: https://trello.com/c/D6dmtXLC

This will only get merged once the new guidance has been published, as coordinated with @karlchillmaid 

It contains an updated link to buyer guidance, for the saved search task
list page.

The breaking change has no effect here. It's just for the search-api.

### DIFF
--- a/tests/fixtures/content/frameworks/g9/messages/urls.yml
+++ b/tests/fixtures/content/frameworks/g9/messages/urls.yml
@@ -10,4 +10,4 @@ customer_benefits_record_form_url: "http://ccs-agreements.cabinetoffice.gov.uk/s
 
 customer_benefits_record_form_email: "gcloud-benefits@crowncommercial.gov.uk"
 
-buyers_guide_compare_services_url: "https://www.gov.uk/guidance/g-cloud-buyers-guide#compare-services"
+buyers_guide_compare_services_url: "https://www.gov.uk/guidance/g-cloud-buyers-guide#review-and-compare-services"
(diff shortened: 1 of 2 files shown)


tests/fixtures/content/frameworks/g9/messages/urls.yml
```diff
@@ -10,4 +10,4 @@ customer_benefits_record_form_url: "http://ccs-agreements.cabinetoffice.gov.uk/s
 
 customer_benefits_record_form_email: "gcloud-benefits@crowncommercial.gov.uk"
 
-buyers_guide_compare_services_url: "https://www.gov.uk/guidance/g-cloud-buyers-guide#compare-services"
+buyers_guide_compare_services_url: "https://www.gov.uk/guidance/g-cloud-buyers-guide#review-and-compare-services"
```
galund: Having this stuff duplicated completely into the fixtures seems a bit weird. Why are we doing that?(@samuelhwilliams?)

Wynndow: Yeah, it threw me a little at first. Tests were failing until I realised the `ContentLoader` in the tests uses fixtures defined in the test repo.

galund: For me, fixtures should be something you feed _in_, so that the output is stable. If we're testing that the output that we get _anyway_ is equal to the fixture then that's not really a normal way to use fixtures, we arguably should just be loading up the referenced g9 stuff independently and checking against that.

PR #669, opened 2017-11-27
Update LICENCE

TheDoubleK:
See https://trello.com/c/nhcs2ylM/48-add-mit-header-to-the-licenc-se-files


samuelhwilliams:
There's a lot of redundancy here.

Maybe we can centralise our licence in another app and pull it into the repos that need it?

TheDoubleK:
Sounds like your Firebreak idea sorted.

PR #670, opened 2017-12-04
101 suppliers should be able to select a category es for frontend

Wynndow:
Part of this ticket: https://trello.com/c/KcqU7k20

This update is to convert the `list_opportunites` view which displays all the briefs [(see here)](https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities), to use elasticsearch instead of the database.

It uses the same patterns as the g-cloud-search page and shares a lot of the same code and templates.

It implements some new filters; location, as well as role for specialists.


DilwoarH:
needs the version for apiclient bumped to the correct version

app/__init__.py
```diff
@@ -35,12 +35,14 @@ def create_app(config_name):
...
             if framework_data['framework'] == 'g-cloud':
                 if framework_data['status'] != 'expired':
-                    content_loader.load_manifest(framework_data['slug'], 'services', 'search_filters')
+                    content_loader.load_manifest(framework_data['slug'], 'services', 'services_search_filters')
                 # we need to be able to display old services, even on expired frameworks
                 content_loader.load_manifest(framework_data['slug'], 'services', 'display_service')
                 content_loader.load_manifest(framework_data['slug'], 'services', 'download_results')
             elif framework_data['framework'] == 'digital-outcomes-and-specialists':
                 content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
+                if framework_data['slug'] == 'digital-outcomes-and-specialists-2':
```
galund: It'd be nice if there was some criteria we could use that would 'just work' assuming DOS3==DOS2. Maybe just check it's not DOS1, as a last resort.

app/main/views/marketplace.py
```diff
@@ -123,50 +132,134 @@ def get_brief_by_id(framework_framework, brief_id):
     )
 
 
-@main.route('/<framework_framework>/opportunities')
-def list_opportunities(framework_framework):
+@main.route('/<framework_family>/opportunities')
+def list_opportunities(framework_family):
```
galund: 👏

app/templates/search/_results_wrapper.html
```diff
@@ -0,0 +1,8 @@
+<div id="js-dm-live-search-results" class="js-dm-live-search-fade">
+  {% if framework_family == 'g-cloud' %}
+    {% include 'search/_services_results.html' %}
```
galund: I wonder if there's something we could do here in terms of creating a directory for each framework family, and building the string for the `include` from that. Arguably not as clear as this, not sure, hmm :)

app/__init__.py
Wynndow: Have updated to use a helper to fetch the latests DOS framework. Should keep it up to date!

app/__init__.py
```diff
@@ -31,16 +32,22 @@ def create_app(config_name):
...
                 if framework_data['status'] != 'expired':
-                    content_loader.load_manifest(framework_data['slug'], 'services', 'search_filters')
+                    content_loader.load_manifest(framework_data['slug'], 'services', 'services_search_filters')
                 # we need to be able to display old services, even on expired frameworks
                 content_loader.load_manifest(framework_data['slug'], 'services', 'display_service')
                 content_loader.load_manifest(framework_data['slug'], 'services', 'download_results')
             elif framework_data['framework'] == 'digital-outcomes-and-specialists':
                 content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
+                content_loader.load_manifest(
+                    get_latest_live_framework(frameworks, 'digital-outcomes-and-specialists')['slug'],
```
galund: We're iterating through frameworks, so this actually means we're going to load the DOS2 filters manifest _twice_.

Maybe we could split loading the search filters manifests into a separate loop (over framework families), covering both DOS and G-Cloud.

app/main/presenters/search_presenters.py
```diff
@@ -193,7 +200,11 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
 
     # Links in the tree should preserve all the filters, except those relating to this tree (i.e. lot
     # and category).
-    search_link_builder = Href(url_for('.search_services'))
+    if framework['framework'] == 'g-cloud':
```
galund: Maybe we should pass the 'link builder' in - so that we can avoid having this test altogether?

app/main/views/marketplace.py
```diff
@@ -123,50 +132,137 @@ def get_brief_by_id(framework_framework, brief_id):
...
+    )
+
+    search_summary = SearchSummary(
+        search_api_response['meta']['total'],
+        clean_request_query_params.copy(),
+        filters.values(),
+        lots_by_slug
+    )
+
+    category_filter_group = filters.pop('categories') if 'categories' in filters else None
```
galund: Feels like potentially we _do_ have some logic here that could be abstracted out between briefs and services filtering - which would be something to tackle in that other ticket, if there's time.

app/__init__.py
Wynndow: Sugar. Yep, I'll do the separate loop. It makes sense if we'll only have one _something_search_filters manifest per family.

Wynndow: Have actually just pulled out the loading of the DOS2 manifest to a separate thing. Looping over framework families didn't really work, as the criteria for which one to load is different depending on the framework family (g-cloud, if the framework isn't expired, dos, the latest framework iteration)

app/main/presenters/search_presenters.py
Wynndow: Nice. yes. done.

app/main/presenters/search_results.py
```diff
@@ -5,28 +5,28 @@ class SearchResults(object):
     """Provides access to the search results information"""
 
     def __init__(self, response, lots_by_slug):
-        self.search_results = response.get('documents') or response.get('services')
+        self.search_results = response['documents']
```
DilwoarH: is the response object now an array?

app/main/views/marketplace.py
```diff
@@ -123,50 +132,146 @@ def get_brief_by_id(framework_framework, brief_id):
...
+
+    template_args = dict(
+        briefs=search_results_obj.search_results,
+        category_tree_root=selected_category_tree_filters[0],
+        clear_filters_url=clear_filters_url,
+        current_lot=current_lot,
+        doc_type=doc_type,
+        filters=filters.values(),
+        filter_form_hidden_fields=filter_form_hidden_fields_by_name.values(),
+        framework=frameworks[-1],
```
DilwoarH: `-1` ?

DilwoarH:
Branch needs rebasing

app/main/views/marketplace.py
```diff
@@ -123,50 +132,146 @@ def get_brief_by_id(framework_framework, brief_id):
...
-                           next_link_args=next_link_args,
-                           briefs_count=api_result.get("meta", {}).get("total", None),
-                           )
+    pagination_config = pagination(
+        search_results_obj.total,
+        current_app.config["DM_SEARCH_PAGE_SIZE"],
+        get_page_from_request(request)
+    )
+
+    search_summary = SearchSummary(
```
galund: Didn't think we were having a search summary for this - am surprised this actually doesn't crash somewhere on some missing data :)

Wynndow: It gives you the '150 results found in all categories' thing. If you don't add any rules to the search summary manifest it doesn't try the rest of it. :)

app/main/views/marketplace.py
galund: In fact we already have a variable `framework` (set to the latest framework in the family) so probably we should be consistent and use that, rather than just the last one in the list?

app/templates/search/_briefs_results.html
```diff
@@ -1,46 +1,46 @@
...
         <li class="search-result-metadata-item">
-            {{ brief.content.summary(brief.data).get_question("specialistRole").value }}
+            {{ brief.specialistRole | capitalize() }}
         </li>
         {% endif %}
     </ul>
 
     <ul class="search-result-metadata">
-        {% if brief.data.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
+        {% if brief.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
```
galund: hmm, it's a pity we aren't ending up with dicts back from the Search API that are more directly compatible with those that come from the main API. Not something we can fix right now, but something to think about next time are looking at either of the two maybe? cc @risicle

app/main/views/marketplace.py
galund: 👍

app/templates/search/_briefs_results.html
risicle: Yup exactly what I said.

app/main/presenters/search_results.py
Wynndow: It's a dict, with an array of results, which are now behind the `documents` key.

tests/main/views/test_marketplace.py
```diff
@@ -1378,32 +1434,234 @@ def test_opportunity_status_and_published_date(self):
...
+                                                      })))
+    @mock.patch('app.main.views.marketplace.render_template', autospec=True)
+    def test_base_page_renders_search_services(self, render_template_patch, query_string, urls):
+        render_template_patch.return_value = '<p>some html</p>'
+
+        self.client.get('/digital-outcomes-and-specialists/opportunities{}'.format(query_string))
+
+        assert urls == set(x[0][0] for x in render_template_patch.call_args_list)
+
+    def test_form_has_js_hidden_filter_button_chris(self):
```
benvand: `_chris` ?

benvand: 😁

Wynndow: oh snap! I was using my name to target specific tests with the `-k` flag! Forgot to remove.....

app/main/presenters/search_presenters.py
```diff
@@ -182,21 +188,24 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
...
-    :param content_manifest: a ContentManifest instance for G-Cloud search_filters.
-    :return: list of selected category and lot filters, starting with the 'all categories' root node node
+    :param request: the original request for creating links for nodes
+    :param cleaned_request_args: current cleaned request args so that we can figure out selected lots and categories
+    :param content_manifest: a ContentManifest instance for the frameworks search_filters.
+    :param search_link_builder: a werkzeug Href object for the view category tree links link to
+    :return: list of selected category and lot filters, starting with the 'all categories' root node
     """
-    current_lot_slug = get_lot_from_args(request.args, [lot['slug'] for lot in lots])
+    current_lot_slug = get_lot_from_args(cleaned_request_args, [lot['slug'] for lot in lots])
```
benvand: This method name (`get_lot_from_args`) is not very descriptive.
What we're actually doing is validating the supplied lot name agains the supplied lots.

At first glance I'm like 'Why do we need a method to `get_lot_from_args` when we can just do `cleaned_request_args.get('lot')` ?'
Then digging in we see that aside from getting the lot from the args it validates the lot against the list that is passed in and defaults to `None`.

benvand: May be out of scope for this PR but it's really difficult to read what's going on in this `build_lots_and_categories_link_tree ` method

tests/main/presenters/test_search_presenters.py
```diff
@@ -301,7 +302,7 @@ def test_instance_has_correct_filter_groups_for_iaas(self):
```
benvand: We could do with better test coverage for the `build_lots_and_categories_link_tree`.
It's really heavily relied on throughout the repo and theres no test that shows exactly what its output should look like.

I still can't really tell what it should return. There's no example in the docstring or the tests unless I've missed something.

app/main/presenters/search_presenters.py
```diff
@@ -182,21 +188,24 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
...
+    :param framework: the framework from the API
     :param lots: a sequence of lot dicts applicable to the live framework(s)
     :param category_filter_group: a single filter group loaded from the framework search_filters manifest
-    :param request: current request so that we can figure out what lots and categories are selected
-    :param content_manifest: a ContentManifest instance for G-Cloud search_filters.
-    :return: list of selected category and lot filters, starting with the 'all categories' root node node
+    :param request: the original request for creating links for nodes
+    :param cleaned_request_args: current cleaned request args so that we can figure out selected lots and categories
+    :param content_manifest: a ContentManifest instance for the frameworks search_filters.
+    :param search_link_builder: a werkzeug Href object for the view category tree links link to
```
benvand: Can't work out what this `search_link_builder` does. Could do with a better doc

app/main/presenters/search_presenters.py
```diff
@@ -182,21 +188,24 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
...
-    search_link_builder = Href(url_for('.search_services'))
     keys_to_remove = _get_category_filter_key_set(category_filter_group)
     keys_to_remove.add('page')
-    preserved_request_args = MultiDict((k, v) for (k, v) in request.args.items(multi=True) if k not in keys_to_remove)
+    preserved_request_args = MultiDict(
+        (k, v) for (k, v) in request.args.items(multi=True) if k not in keys_to_remove
+    )
 
     selected_filters = list()
     # Create root node for the tree, always selected, which is the parent of the various lots.
```
benvand: Don't know why we're building this dictionary instead of just creating it.
I feel like someing along the lines of:
```
    root_node = {
        'label': 'All categories',
        'selected': True,
        'link': search_link_builder(_build_base_url_args(preserved_request_args, content_manifest, framework, None)),
        'children': []
    }
```

Would let the reader see an obvious example of a node before diving in to the construction of the child elements below.

app/main/presenters/search_presenters.py
```diff
@@ -208,13 +217,11 @@ def build_lots_and_categories_link_tree(framework, lots, category_filter_group,
...
-                                                                               framework, request)
-                           for lot in lots}
+    aggregations_by_lot = {
+        lot['slug']: _get_aggregations_for_lot_with_filters(
+            lot['slug'], content_manifest, framework, cleaned_request_args, doc_type, index
+        ) for lot in lots
+    }
 
     for lot in lots:
         selected_categories = []
```
benvand: `lot_selected = (lot['slug'] == current_lot_slug)`
extraneous parenthesis
Again slightly out of scope.

app/main/presenters/search_presenters.py
benvand: So this is the currently selected lot slug. This would also benefit from a more descriptive name. As someone who doesn't know this code well `current` adds nothing to my understanding. It may as well just be `lot_slug`. I can't tell if it's parsed from a url or what kind of operation wee might be performing. Whereas `currently_selected_lot_slug` at least indicates that it is the lot slug that has been selected for an action. Maybe something like `root_lot_slug` or `selected_root_lot_slug` I don't know. just on't think `current_lot_slug` is descriptive enough.

Wynndow: Have renamed `get_lot_from_args` to `get_valid_lot_from_args_or_none`

app/main/presenters/search_presenters.py
Wynndow: Have updated the doc string with more info.

app/main/presenters/search_presenters.py
risicle: I like them as it highlights that it's doing an `==`, not just a three-way `=`

app/main/presenters/search_presenters.py
risicle: Remember that a lot of this stuff is just inherited from the catalogue search, so any improvements made here should be reflected in that code too so the consolidation stage is easier.

app/main/presenters/search_presenters.py
Wynndow: Good shout, have updated.

app/main/presenters/search_presenters.py
Wynndow: Have looked at it both ways and I agree with Rob i think

app/main/presenters/search_presenters.py
Wynndow: Have done the same for the lot_filters just below.

tests/main/presenters/test_search_presenters.py
Wynndow: Have added some tests that show the output of `build_lots_and_categories` which shows what it should be outputting at various levels. There are probably more test cases that should be added, but a lot of it is tested implicitly by the tests on the view.

app/main/presenters/search_presenters.py
benvand: Fair enough.

I think these are good reasons not to:
 * https://google.github.io/styleguide/pyguide.html?showone=Parentheses#Parentheses
 * Extra, extraneous characters to parse
 * Brackets in python are commonly used to indicate the definition of a tuple
 * It breaks convention which reduces confidence

My 2 cents:
So this is a bit more conceptual and a personal preference thing but I feel like it discourages recognition of the action of the `lot['slug'] == current_lot_slug` part. 
This piece of code _will always_ be evaluated. It doesn't need the assignment or the brackets. Because it's actually, effectively, a function call.
In this case it's equivalent to  `str.__eq__(lot['slug'], current_lot_slug)`

If it was a ternary statement it would have an `if` which should be obvious if you've got syntax highlighting

Wynndow: You make good points Ben. I'm going to remove them.

PR #671, opened 2017-12-04
Search limit restricted to 100

mcraddock-ungp:
Could be related to this bug, unable to browse pass page 100?

Go to this page and try to browse to page 101, https://www.digitalmarketplace.service.gov.uk/g-cloud/search?page=100

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -44,7 +44,7 @@ from ..exceptions import AuthException
 from app import search_api_client, data_api_client, content_loader
 
 
-END_SEARCH_LIMIT = 100  # TODO: This should be done in the API.
+END_SEARCH_LIMIT = 196  # TODO: This should be done in the API.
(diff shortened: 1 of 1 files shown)


TheDoubleK:
Many thanks for the pull-request, Mark, but this won't fix the issue that you mention here.  The number here is the maximum number of results that can be saved by a buyer when finalising a G-Cloud "saved search", and we believe the current limit of 100 should be more than enough for a short list of services to evaluate.

The issue you mention was introduced when we recently upgraded to a newer version of Elasticsearch, and is unrelated to this constant in the config here (it's just a coincidence that 100 is the limit on both things). There is a bug ticket in our backlog to address the issue you mention - do let us know if it is causing you problems day-to-day as that will help us prioritise the issue as urgent. It is already prioritised for our next sprint, but we have assumed that users do not generally need to browse past the first 100 pages of results.

PR #672, opened 2017-12-04
Remove redundant SCSS

idavidmcdonald:
https://trello.com/c/bJKatcwz/77-remove-old-scss-buyer-front-end

- 25% reduction in minified CSS size
- Some files were being imported twice into application.scss for what we can only assume is no reason
- Some of the imports from the toolkit are not currently being used so we have removed them so the application.js only imports things it actually uses. 
- We haven't touched or really looked at _footer.scss as there are already plans to consolidate this.

One small discrepancy is introduced on the cookies page with a reduction in padding on the `.application-notice p`. This is due to non defensive CSS and the order of imports (`_legal.scss` comes before `_notice.scss` whereas before it was `_publication.scss` came after `_notice.scss`). However the cookies page is very unloved at the moment and needs a whole rewrite on the FE side of things. I've checked with Laurence and he is happy therefore for the reduction in padding to go out.

Old:
image

New:
image


app/assets/scss/_publication.scss
```diff
@@ -1,383 +1,4 @@
...
-    margin-left: 0;
-  }
-
-}
-
-.floated-children {
-  overflow: hidden;
-}
 
 .terms-and-conditions {
```
samuelhwilliams: At this point it feels like this file just needs renaming to e.g. `_terms_and_conditions.scss`, given that's all we have left here.

We need to be careful with this, I think. If we start gutting parts of files, it feels like we're going to create more splintering rather than less, with different fragments of files in each app and no easy way to compare them anymore. This feels like it could increase the mental overhead for developers when moving between apps rather than reduce it.

app/assets/scss/application.scss
```diff
@@ -1,49 +1,31 @@
...
-@import "design-patterns/alpha-beta";             // Only required if using _phase-banner.scss
-@import "design-patterns/buttons";                // Only required if using _buttons.scss
-
-// Functions
-@import "url-helpers";
+// GOVUK Front-end toolkit variables, mixins and functions
+@import "colours";
+@import "_grid_layout.scss";
+@import "_typography.scss";
+@import "_url-helpers.scss";
```
samuelhwilliams: I think I'd rather keep the inline comments next to these. Even if it's a bit redundant, the extra few words give you a bit more context (eg which are mixins, which are vars, etc).

app/assets/scss/_publication.scss
idavidmcdonald: Good point.

I think it's generally OK if a file contains two 'components' or unrelated things are you remove one but we probably want to think a bit harder about removing parts of a 'component'.

I'll have a scan through myself but are there any in this PR you think might be worth rethinking?

samuelhwilliams: This was the main one. If we have files with multiple components in, and we are only using a subset of those components for one of our apps, then clearly they aren't that closely related and could be separate files. It feels like it would be simpler if our files were the same across apps, and we just include whichever files we need. That we, we can see (eg) `publication.scss` and know that it does the same thing in the buyer app and the briefs app. If they both do different things across apps, that's surely more confusing (to me at least)..

idavidmcdonald: I get you.

I've renamed the file and also renamed it to be more appropriate for the cookies page too. Let me know what you think.

app/assets/scss/application.scss
idavidmcdonald: No problem, I added them back in

app/assets/scss/application.scss
```diff
@@ -81,18 +72,25 @@ $path: "/static/images/";
 @import "_table.scss";
 @import "_text.scss";
 
+// Misc styles
+// TODO: Move misc styling into their own appropriate files to import or the Digital Marketplace FE Toolkit
```
samuelhwilliams: We could just do this now rather than leave it as a TODO?

idavidmcdonald: I can do one or two of these now I think. However, at the moment I would only be moving the rest into a misc file (which feels bad) because the majority of these need a bit of refactoring/shouldn't really exist which would take a little bit of time (like thinking how all our typography is structured).

`padding-bottom-small` is not semantic and was some styling copied from the prototype which we shouldn't really be using. Not sure how best to replace this at the moment.
`return-to-top` I could move to the service-page.scss as I think it is just used at the bottom of service pages
`marketplace-paragraph` is probably another candidate from removal as we should be using typography for this.
`break-link` should potentially go into the toolkit as we write the HTML for it in the utils repo.
`frameworks-lots` is just used for the g cloud index page. Not sure if it's worth creating a single file just for this one at the moment.

So I will move the easy one but leave the rest. Does that sound reasonable or am I being lazy?

samuelhwilliams: I think framework lots is fine to go into a separate file (although it's actually meant to be an update to our general styling of bulleted lists so it _should_ go to the fe-toolkit eventually... as we are/were heading towards removing hanging bullets).

The others, yeah, I guess can stay. Looking into marketplace-paragraph/dmspeak usage in this app, there's clearly work to be done here...

app/assets/scss/_gcloud_browse_list.scss
```diff
@@ -0,0 +1,4 @@
+.framework-lots ul li ul {
+  list-style: disc;
+  margin: 0px 0px 0px 20px;
+}
```
samuelhwilliams: Missing EOF newline. :trollface:

idavidmcdonald: not anymore...

samuelhwilliams:
💥

PR #676, opened 2017-12-06
Accept `documents` as well as `services` in SearchResults

Wynndow:
In an upcoming change to the search-api ([see this PR](https://github.com/alphagov/digitalmarketplace-search-api/pull/124/))
we're going to be returning data structure with a top level `documents`
key instead of `services`. To allow this PR to go out without breaking
the buyer frontend, we need to be able to accept both keys for a while.

### DIFF
--- a/tests/main/presenters/test_search_results.py
+++ b/tests/main/presenters/test_search_results.py
@@ -49,9 +49,9 @@ class TestSearchResults(BaseApplicationTest):
 
     def test_search_results_accepts_top_level_documents_key_in_response_instead_of_services(self):
         self.fixture['documents'] = self.fixture.pop('services')
+        result = SearchResults(self.fixture, self._lots_by_slug)
 
-        search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
-        assert hasattr(search_results_instance, 'search_results')
+        assert result.search_results == self.fixture['documents']
 
     def test_search_results_is_set(self):
         search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
--- a/tests/main/presenters/test_search_results.py
+++ b/tests/main/presenters/test_search_results.py
@@ -47,6 +47,12 @@ class TestSearchResults(BaseApplicationTest):
             self._get_framework_fixture_data('g-cloud-6')['frameworks']
         )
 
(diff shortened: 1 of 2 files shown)


tests/main/presenters/test_search_results.py
```diff
@@ -47,6 +47,12 @@ def setup_method(self, method):
             self._get_framework_fixture_data('g-cloud-6')['frameworks']
         )
 
+    def test_search_results_accepts_top_level_documents_key_in_response_instead_of_services(self):
+        self.fixture['documents'] = self.fixture.pop('services')
+
+        search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
+        assert hasattr(search_results_instance, 'search_results')
```
galund: I don't think this test can fail - if you call the constructor, it will always have that attribute. You need to test for `None`, I think?

risicle: You can do `getattr(search_results_instance, "search_results", None) is not None` to catch either case.

Wynndow: I was testing here that you can instantiate a SearchResults object with a response that has a `documents` key rather than a `service` key. If you change the key to anything else the test will fail.

Wynndow: Maybe I could just change to `assert SearchResults(self.fixture, self._lots_by_slug)`

Wynndow: Have updated!

tests/main/presenters/test_search_results.py
```diff
@@ -47,6 +47,12 @@ def setup_method(self, method):
...
 
+    def test_search_results_accepts_top_level_documents_key_in_response_instead_of_services(self):
+        self.fixture['documents'] = self.fixture.pop('services')
+        result = SearchResults(self.fixture, self._lots_by_slug)
+
+        assert result.search_results == self.fixture['documents']
+
     def test_search_results_is_set(self):
         search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
         assert hasattr(search_results_instance, 'search_results')
```
galund: I see where you copied the hasattr thing from now :)

PR #677, opened 2017-12-06
Remove bower and upgrade dependencies

MuriloDalRi:
* Remove bower as a dependency and use yarn to manage our frontend dependencies
* Move all our bower dependencies to package.json and update them
* Delete bootstrap.sh and run_tests.sh scripts in favour of the makefile
* Bring node version into line with production
* Update makefile for new yarn commands
* Update README with yarn instructions


.travis.yml
```diff
@@ -2,9 +2,10 @@ language: python
 python:
   - "3.6"
 before_install:
-  - nvm install 4.2
+  - nvm install 6.10.3
```
idavidmcdonald: Could you explain the logic behind the bump please and why this version? Note, may be worth adding in a commit message etc

idavidmcdonald:
Oh, take a look at the README as well to see if that is all up to date or there are sections that need changing please (potentially around yarn being installed?)

.travis.yml
samuelhwilliams: This is the version of node that we use in production.

https://github.com/alphagov/digitalmarketplace-docker-base/blob/master/base.docker#L5

samuelhwilliams:
Don't forget to update the nixfile too. 😉

yarn.lock
```diff
@@ -0,0 +1,3124 @@
...
+
+jasmine@dflynn15/jasmine-npm:
+  version "2.4.1"
+  resolved "https://codeload.github.com/dflynn15/jasmine-npm/tar.gz/5f74e4336b247e67bbc509a8f82f3c4fa6b52964"
+  dependencies:
+    exit "^0.1.2"
+    glob "^3.2.11"
+    jasmine-core "~2.4.0"
+
+jquery@1.11.2:
```
samuelhwilliams: I think you might need to regenerate this lockfile as I've run this locally and jquery@1.12.0 gets installed (as specified in our package.json).

package.json
```diff
@@ -3,27 +3,30 @@
   "description": "Front end application for the Digital Marketplace",
   "version": "0.0.1",
   "private": true,
-  "engine": "node >= 4.2",
+  "engine": "node >= 6.10.3",
```
samuelhwilliams: Should this be `node == 6.10.3`?

samuelhwilliams:
We can also remove `bower_components` from our `.gitignore`

yarn.lock
MuriloDalRi: Ah yeah, will do!

package.json
MuriloDalRi: Yes, well spotted! I think we might want to start using version ranges such as ```6.1.x``` but we should discuss the pros and cons.

.travis.yml
MuriloDalRi: What Sam said.

MuriloDalRi:
I'll definitely update the readme and clean up the commit messages before merging this!

The nixfile! OMG I forgot about it!!! 😱 

I thought I'd leave ```bower_components``` in our ```.gitignore``` for now because people might forget to manually remove them.

Thanks for the comments!

samuelhwilliams:
That's the reason I think it would be good to remove it from the `.gitignore`, actually, because it will highlight that people _should_ remove the directory.

scripts/bootstrap.sh
```diff
@@ -1,11 +0,0 @@
...
-#
-# Bootstrap virtualenv environment and postgres databases locally.
-#
-# NOTE: This script expects to be run from the project root with
-# ./scripts/bootstrap.sh
-
-set -o pipefail
-
-# Install development dependencies
-make requirements-dev npm-install frontend-build
```
katstevens: Curious about this file - it mentions bootstrapping the postgres DB but I can't see any evidence that it ever did that?

scripts/run_tests.sh
```diff
@@ -1,8 +0,0 @@
-#!/bin/bash
-#
-# Run project tests
-#
-# NOTE: This script expects to be run from the project root with
-# ./scripts/run_tests.sh
-
-make test
```
katstevens: Hurray for getting rid of this 🔥

scripts/bootstrap.sh
MuriloDalRi: Maybe it was in the makefile at some point but it certainly doesn't seem useful now.

package.json
```diff
@@ -3,27 +3,30 @@
...
-  "engine": "node >= 4.2",
+  "engine": "node == 6.10.3",
   "dependencies": {
-    "bower": "1.5.2",
-    "gulp": "3.8.7",
-    "gulp-uglify": "1.4.0",
-    "gulp-sourcemaps": "1.5.2",
-    "del": "1.1.1",
+    "del": "1.2.0",
+    "digitalmarketplace-frameworks": "https://github.com/alphagov/digitalmarketplace-frameworks.git#9.3.2",
```
idavidmcdonald: Is this being downgraded from what it originally was?

package.json
```diff
@@ -3,27 +3,30 @@
   "description": "Front end application for the Digital Marketplace",
   "version": "0.0.1",
   "private": true,
-  "engine": "node >= 4.2",
+  "engine": "node == 6.10.3",
```
samuelhwilliams: Although this is what we use in production and it would be nice to lock this down, we need to think about how it might affect local development. If we lock this exact version, will it prevent using other versions - or just be silently ignored if that version isn't available?

It would be a good thing to get us all running on 6.10.3 but we need to be aware to send out a dev circular if things are suddenly going to stop working. Specifically thinking about Rob, for whom using 6.10.3 might not be as simple as it is for us on Macs. That's not to say we shouldn't do it of course, but let's be aware of the impact it'll have.

scripts/bootstrap.sh
TheDoubleK: There's nothing in the history. It's not clear to me what it was ever for, I've certainly never used it. Kill it with fire.

scripts/run_tests.sh
TheDoubleK: Why? This is how I generally run my tests. I guess someone needs to force me into the future of using `make`, I still don't really see the value in it TBH.

package.json
MuriloDalRi: Would something like ```^6``` be better?

We need to get better at updating our dependencies but you're right that we need to make sure we don't break anyone's local environment in the process.

Anyway I think we should be using at least v6.12.2 everywhere, so it would be better to decide now before I merge anything

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V6.md

@risicle thoughts?

risicle: It's not that it's "not simple" for me - it's that I practise proper package management rather than package soup. I have 6.11.5 available to me.

risicle: So... if you look at the "security issues" addressed in the later v6 releases, they're actually security vulnerabilities in their *bundled copy of openssl*. Yes. nodejs *bundles their own copy of openssl* - what do they think they're playing at? Anyway, any sensible distribution patches that out and replaces it with the distribution's openssl. Thus nix's 6.11.5 is fully up to date with all security issues 6.12.whatever nodejs claims to be.

scripts/run_tests.sh
risicle: But surely... since it's the *script* running *make*, it's the *script* you don't see value in...?

risicle:
This WFM as long as you add `pkgs.yarn` to `default.nix`'s `buildInputs`. This gives a nixos-r17.09 user yarn 1.0.1 - hope that's ok, seems to work. Might also want to change `pkgs.nodejs` to `pkgs.nodejs-6_x` if we're starting to be particular about versions.

package.json
MuriloDalRi: I'll change it to ```"node >= 6.10.3"``` so it's less strict

samuelhwilliams: We should make sure it's a v6 though. Can you do something like `6.10.3 <= node < 7.0.0`?

MuriloDalRi: Good point @samuelhwilliams

README.md
```diff
@@ -121,7 +121,7 @@ Front-end code (both development and production) is compiled using [Node](http:/
 
 ### Requirements
 
-You need Node (minimum version of 0.10.0, maximum version 0.12.x), which will also get you [NPM](npmjs.org), Node's package management tool. 
+You need 6.10.3 <= node < 7.0.0 which is what we currently use in production.
```
galund: Can we clarify why we wouldn't be able to upgrade node beyond 7.0.0 (hmm, what about security fixes?)

MuriloDalRi: So my original idea was to enforce the same version we use in production, but that might be a bit too strict. Do you think we should just specify a minimum version?

galund: I see what you mean - the ideal DRY thing would be for us to all be developing using docker - the version of node is specified in the Dockerfile after all.

Perhaps it would be clearer to just reference the base image, maybe say something like "You need Node (try to install the version we use in production -
 see the [base docker image](https://github.com/alphagov/digitalmarketplace-docker-base/blob/master/base.docker)."

scripts/run_tests.sh
galund: @TheDoubleK would much rather be using Gradle right :-D

README.md
MuriloDalRi: It would be neat if we could get the version from that file, but I couldn't think of a simple way of doing that. I agree your content is better, I'll update it.

We should have a reminder somewhere to update node before April next year: https://github.com/nodejs/Release#release-schedule

scripts/run_tests.sh
TheDoubleK: The script used to be the only way to run things - there was no Makefile.  Some bright spark made the Makefile do what the script used to do, and updated the script to just run make, which meant I never really updated my workflow to use the new Makefile directly.  It's fine.  We can kill this.  I can move on.

README.md
galund: Hijacked your existing ticket and scheduled it for "later" https://trello.com/c/YDiLYdWQ/181-update-node-to-version-8 :-)

scripts/run_tests.sh
benvand: FWIW I generally run the tests using the `py.test` command because there's a load of useful params you can pass in.

py.test 
-s (print standard out/ print statements as encountered instead of at the end of the test run, also allows pdb in tests) 
-k <name of a single test or test class if you like>
--tb native (for normal python tracebacks)
<name of a file>

Its awesome.

risicle: Yeah I only dig in & run `py.test` when I want to do proper digging. Most the time I want to run what jenkins'll run (include the QA metrics and javascripty stuff)

benvand: Yeah I do get caught out from time to time because py.test will pass then I'll push something up then flake8 will fail on jenkins :grin:

PR #680, opened 2017-12-11
Improve styling for search summary

laurenceberry:
Related PR: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/392

Originally I planned to just add a border to the summary results for the opportunities page but then realised that the search results summary needed more consideration.

This commit:
• removes a border-top from the save search
• increases the font-size for search results summary and suggestion
• increases font sizes for mobile

Before:
screen shot 2017-12-11 at 12 13 10

screen shot 2017-12-11 at 12 12 36

screen shot 2017-12-11 at 12 15 47

After:

screen shot 2017-12-11 at 11 59 46

screen shot 2017-12-11 at 12 00 03

screen shot 2017-12-11 at 11 28 29

### DIFF
--- a/app/assets/scss/_search-page.scss
+++ b/app/assets/scss/_search-page.scss
@@ -1,8 +1,6 @@
 @import "search/_search-summary.scss";
 
 .search-summary {
-  padding: 0 0 $gutter-half;
-
   em {
     font-style: normal;
     font-weight: bold;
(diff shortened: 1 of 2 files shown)


app/assets/scss/_search-page.scss
```diff
@@ -95,7 +92,7 @@
   .dm-filters {
```
TheDoubleK: I think we also have filters in briefs front-end now? So would be good to get these styles into the toolkit to ensure consistency. Not sure if the front-end mission is doing this?

PR #681, opened 2017-12-11
Upgrade API client to v13.1.0

katstevens:
Trello: https://trello.com/c/whH4DvW1/125-high-severity-vulnerability-in-python-requests-package

Upgrade the API client to fix a vulnerability in the Python requests package.

This means a couple of major version bumps, hopefully all OK:
- v13 `update_user` (not used in this app)
- v12 Search API index creation (not used in this app)
- v11 `get_direct_award_project_services` and `find_direct_award_project_searches` - used in this app but with keyword arguments, so should not be affected by the breaking change

Please double check if there's anything I've missed.


risicle:
Cool just let me run the tests locally...

PR #684, opened 2017-12-13
Catch missing file extension errors for service documents

katstevens:
Mitigates the error seen in this 2nd line ticket: https://trello.com/c/tvGUsJeL/110-crash-on-preview-bad-g-cloud-service

In theory this error should only happen with bad data on preview or staging - suppliers should be uploading valid file formats (PDF, ODF). We should still fail loudly if this happens, as other stuff will break. But now if a bad filename does somehow sneak in, then we will have a sane error message telling us about it that we can debug.

Possible issue: do we have any concerns logging the full document URL?


risicle:
Running tests locally...

PR #691, opened 2018-01-04
Adds in support page

DilwoarH:
Ticket: https://trello.com/c/Ss5PxdxQ/77-contact-filtering-for-buyer-and-supplier-contact-details-lists

Required PRs: 
- https://github.com/alphagov/digitalmarketplace-utils/pull/346

screenshot-localhost-2018-01-04-10-30-52-159

### DIFF
--- a/app/assets/scss/_support-page.scss
+++ b/app/assets/scss/_support-page.scss
@@ -1,0 +1,6 @@
+.support-page {
+  p {
+    margin-bottom: 30px;
+  }
+}
+
--- a/app/templates/_footer_categories.html
+++ b/app/templates/_footer_categories.html
@@ -5,7 +5,7 @@
     </h2>
     <ul>
       <li>
-        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">Digital Marketplace support</a>
+        <a href="{{ url_for('main.support') }}">Digital Marketplace support</a>
       </li>
     </ul>
   </div>
--- a/app/templates/_footer_categories.html
+++ b/app/templates/_footer_categories.html
@@ -5,7 +5,7 @@
     </h2>
     <ul>
       <li>
-        <a href="{{ url_for('main.support') }}">Digital Marketplace support</a>
+        <a href="{{ url_for('main.help') }}">Digital Marketplace help</a>
       </li>
     </ul>
   </div>
--- a/app/templates/support/index.html
+++ b/app/templates/support/index.html
@@ -1,0 +1,54 @@
+{% extends "_base_page.html" %}
+{% block page_title %}What do you need help with? - Digital Marketplace{% endblock %}
+
+{% block breadcrumb %}
+  {%
+    with
+    items = [
+      {
+          "link": url_for('main.index'),
+          "label": "Digital Marketplace"
+      }
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="support-page">
+  
+
+  <div class="grid">
+    
+      {% with heading = "What do you need help with?" %} {% include "toolkit/page-heading.html" %} {% endwith %}
+  
+      <div class="dmspeak">
+        <h2 class="heading-large">Selling services</h2>
+      </div>
+      <p>
+        To get support with the selling process, for example, updating your prices, email:
+        <a href="mailto:cloud_digital@crowncommercial.gov.uk">cloud_digital@crowncommercial.gov.uk</a>
+      </p>
+
+      <div class="dmspeak">
+        <h2 class="heading-large">Buying services</h2>
+      </div>      
+      <p>
+        To get support with the buying process, for example, evaluating suppliers, email:
+        <a href="mailto:cloud_digital@crowncommercial.gov.uk">cloud_digital@crowncommercial.gov.uk</a>
+      </p>
+
+      <div class="dmspeak">
+        <h2 class="heading-large">Creating and updating</h2>
+      </div>
+      <p>
+        To make changes to your user account, for example, if you have trouble logging in, email:
+        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a>
+      </p>
+    
+  </div>
+  
+</div>
+{% endblock %}
--- a/app/assets/scss/_help-page.scss
+++ b/app/assets/scss/_help-page.scss
@@ -1,4 +1,4 @@
-.support-page {
+.help-page {
   p {
     margin-bottom: 30px;
   }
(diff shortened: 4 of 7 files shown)


karlchillmaid:
@constancecerf Suggest a minor content change:

> To get help selling your services, for example...

and 

> To get help with the buying process...

Not a deal-breaker, though.

app/templates/_footer_categories.html
```diff
@@ -5,7 +5,7 @@ <h2>
     </h2>
     <ul>
       <li>
-        <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">Digital Marketplace support</a>
+        <a href="/support">Digital Marketplace support</a>
```
TheDoubleK: This link is going to need to be changed individually in all the other apps too I think 😞 

Ideally I think we want a `url_for('.support')` rather than a hard-coded path here, and an external route defined here: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/external.py so that other apps can use `url_for` to link to the page in their footers.

app/templates/support/index.html
```diff
@@ -0,0 +1,52 @@
...
+{% endblock %}
+
+{% block main_content %}
+
+<div class="support-page">
+  
+
+  <div class="grid">
+    
+      <header class="page-heading">
```
TheDoubleK: Can we use the toolkit macro for page headings here?

app/assets/scss/_support-page.scss
```diff
@@ -0,0 +1,8 @@
+.support-page {
+  @media (min-width: 641px) {
+    .page-heading-smaller {
+      margin-bottom: 20px;
```
TheDoubleK: Should this be part of the toolkit heading style?  Why would we only want this margin for this particular page and not for all headings that are this size?
  
I'm really wary of adding per-app styles, especially for something as common as a heading.  We should add styles in the toolkit to ensure consistency across the site.

app/templates/support/index.html
```diff
@@ -0,0 +1,52 @@
+{% extends "_base_page.html" %}
+{% block page_title %}What are you trying to do? - Digital Marketplace{% endblock %}
```
TheDoubleK: This question seems a weird title (and heading) for the page - the link to it is "Digital Marketplace support", I think the page title at least could be "Support", and maybe the H1 too, with the question as a lower-level heading.

app/templates/support/index.html
```diff
@@ -0,0 +1,52 @@
...
+
+  <div class="grid">
+    
+      <header class="page-heading">
+        <h1>
+          What are you trying to do?
+        </h1>
+      </header>
+  
+      {% with smaller = true, heading = "Sell services" %} {% include "toolkit/page-heading.html" %} {% endwith %}
```
TheDoubleK: This will create multiple `<h1>`s on the page, which we don't really want (it's bad for accessibility - we generally only have one top-level heading per page).  These should be `<h2>`s.  Kat recently added some styles for these in `dmspeak` (copied from GOV.UK styles - see here: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/200483ce04fe3715d0a624aedbdd0f33b90f084d), which might work here?

app/assets/scss/_support-page.scss
DilwoarH: This specific design change is just for this page, everywhere else they want to keep it the same

app/templates/support/index.html
DilwoarH: @laurenceberry - can you take a look?

constancecerf: Change

What are you trying to do?

TO

What do you need help with?

REASON

It sounds more obviously like support. We could also change the link text on the homepage to DigitalMarketplace help, support is more internal language.

Change

Sell services to

Selling services

Change

Buy services to

Buying services

CHANGE

Create and update

to

Creating and updating 

REASON

Tense needs to change to fit the new question.

app/templates/support/index.html
```diff
@@ -0,0 +1,54 @@
...
+    ]
+  %}
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="support-page">
+
```
idavidmcdonald: Super unimportant but do we want both these blank lines here?

app/templates/support/index.html
```diff
@@ -0,0 +1,54 @@
...
+    {% include "toolkit/breadcrumb.html" %}
+  {% endwith %}
+{% endblock %}
+
+{% block main_content %}
+
+<div class="support-page">
+  
+
+  <div class="grid">
```
idavidmcdonald: Just looking at another template and we seem to tend to have:

```
<div class="grid-row">
```

I haven't spent the time looking at what those classes do but it seems we do them elsewhere quite often (please double check this though). Is this something we are good to deviate from?

TheDoubleK:
@samuelhwilliams makes the point (here https://github.com/alphagov/digitalmarketplace-utils/pull/346) that this feels more like a user app page than a buyer app thing (it's certainly not specific to only buyers), in which case the URL would be `/user/help`, which seems pretty rasonable.

Should be an easy lift-and-shift to there if we agree?

idavidmcdonald:
I think that the buyer is the appropriate place for this. I know it's called the buyer app and this page isn't directly related to buyers but it's currently where we keep pages of a similar nature such as the home page and static pages like terms and conditions, cookies etc. I don't think we had to do any nginx set up for those static pages (a quick grep of the router app for `terms-and-conditions` didn't find me anything at least) - correct me if I'm wrong though.

Whilst this app is very ambigious in terms of it's use, I felt that the user app had a somewhat defined  purpose which was to handle user authentication (although always up for debate).

samuelhwilliams:
Hmm I think that's a fair point re: user-frontend having a tight purpose around authentication only.

If that's the case, `/help` or `/support` are fine (I prefer the latter).

TheDoubleK:
Yep fine with me to keep it here.  Buyer is generally our "www" open-to-all stuff.

idavidmcdonald:
@karlchillmaid @constancecerf Not sure who was leading on this but just to pass on some feedback. There was discussion about the wording on Friday at the end of the day. Some people felt strongly that `support` was the better word in general but then also with the new header, the terms `help` and `guidance` were too similar. 

Just thought I'd flag. Don't know what you think about that?

Would be good if you could just confirm you are happy as things stand?

constancecerf:
Help is consistent with GOV.UK also it still says support in the body so for SEO it'll still work.

karlchillmaid:
@DilwoarH 

A couple of content changes agreed with @constancecerf:

1. Please replace:
~To get support with the selling process, for example, updating your prices, email:~
with:
`For support selling your services, for example, updating your prices, email:`

2. Please update the third h2:
~Creating and updating~
so it reads:
`Creating and updating an account`

constancecerf:
@karlchillmaid @DilwoarH 

We should also change:

To get support with the buying process, for example...

TO

For support buying services, for example...

Just so it's consistent and easier to scan.

TheDoubleK:
I think before merging this it would be good to have corresponding PRs ready to go on the other frontend apps to update the footer links everywhere.  Otherwise we're going to have links that do one thing from the buyer app but something completely different in the other apps.

PR #695, opened 2018-01-08
Update header to make it easier to find support guidance and where to log in

DilwoarH:
This is part of a larger piece of work to change the homepage styling so it is easier to find things on the page and consistent with other services in the service toolkit.

Examples of other pages that use this pattern can be found under the components heading on the service toolkit. https://www.gov.uk/service-toolkit

The code for these pages can be found here: https://github.com/alphagov/product-page-example

As part of this work we will make support easier to find and remove the beta phase banner

The guidance link would go to: https://www.gov.uk/government/collections/digital-marketplace-buyers-and-suppliers-information

The GOV.UK Digital Marketplace logo would go to the digital marketplace home page.

Ticket ID: https://trello.com/c/mityCieG/106-update-header-to-make-it-easier-to-find-support-guidance-and-where-to-log-in

screen shot 2018-01-12 at 11 15 26


app/templates/index.html
```diff
@@ -4,7 +4,6 @@
 
 {% block top_header %}
 <header class="marketplace-homepage-heading">
-  {% include "toolkit/phase-banner.html" %}
```
idavidmcdonald: We can also remove the import for the corresponding scss file in application.scss

DilwoarH: unfortunately, the CSS is still required for the beta logo. As part of the removal of phase banner, I will separate these

PR #700, opened 2018-01-18
Change datetime to reflect localised time zone

brenetic:
We were able to use a pre existing filter to achieve this.

https://trello.com/c/BBgL8WtW

Paired with @katstevens

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -11,7 +11,7 @@ from dmcontent.formats import format_service_price
 from dmcontent.questions import Pricing
 
 from dmutils.ods import A as AnchorElement
-from dmutils.formats import dateformat, DATETIME_FORMAT, utcdatetimeformat
+from dmutils.formats import dateformat, DATETIME_FORMAT, datetimeformat
 from dmutils.filters import capitalize_first
 from dmutils.views import SimpleDownloadFileView
 from dmapiclient import HTTPError
@@ -617,7 +617,7 @@ class DownloadResultsView(SimpleDownloadFileView):
 
         locked_at = datetime.strptime(project['lockedAt'], DATETIME_FORMAT)
         filename = '{}-{}-results'.format(locked_at.strftime('%Y-%m-%d'), inflection.parameterize(project['name']))
-        locked_at = utcdatetimeformat(locked_at)
+        formatted_locked_at = datetimeformat(locked_at)
 
         services = self.data_api_client.get_direct_award_project_services_iter(project_id=project['id'],
                                                                                user_id=current_user.id,
@@ -641,7 +641,7 @@ class DownloadResultsView(SimpleDownloadFileView):
             'services': all_services,
             'filename': filename,
             "sheetname": "Search results",
-            'locked_at': locked_at,
+            'locked_at': formatted_locked_at,
             'search_summary': search_meta.search_summary.text_content(),
         }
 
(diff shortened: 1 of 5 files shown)


samuelhwilliams:
I feel like we should be moving towards showing the timezone everywhere rather than away from it, for the sake of removing ambiguity and keeping things clear for our international suppliers. There's always room for doubt, particularly around the time when we flip between GMT/BST, even if you're in the UK.

galund:
FWIW I agree with Sam, although I'm aware that Laura went the other way when you asked her. It would be good to justify why we don't want to show the TZ, so that we don't end up flip-flopping as different people's views are sought.

app/main/views/g_cloud.py
```diff
@@ -615,9 +615,9 @@ def get_file_context(self, **kwargs):
         manifest_questions = get_questions_from_manifest_by_id(download_results_manifest)
         price_question = list(filter(lambda x: type(x) == Pricing, manifest_questions.values()))
 
-        locked_at = datetime.strptime(project['lockedAt'], DATETIME_FORMAT)
-        filename = '{}-{}-results'.format(locked_at.strftime('%Y-%m-%d'), inflection.parameterize(project['name']))
-        locked_at = utcdatetimeformat(locked_at)
+        utc_locked_at = datetime.strptime(project['lockedAt'], DATETIME_FORMAT)
+        filename = '{}-{}-results'.format(utc_locked_at.strftime('%Y-%m-%d'), inflection.parameterize(project['name']))
```
galund: In the edge case that the filename would be different depending on whether we are using local time or UTC, it seems a bit confusing for the user to use UTC for the date. Why not go with the local date for consistency?

katstevens: We did wonder about that and thought it would be unlikely the buyer would be ending a search between 12am-1am, but happy to change for consistency. I'm not 100% where this file gets used - is it just for the buyer's own reference?

galund: Yeah this is just the download CSV of the buyer's shortlisted services? Yes it's not uploaded again anywhere else, so the naming is just to help the buyer be organised.

katstevens: Cool. Hopefully they will never notice any change (unless they are/have been working very late!)

katstevens:
We've caved in and made a new date filter in the utils: https://github.com/alphagov/digitalmarketplace-utils/pull/350

Will update this PR to pull it in.

brenetic:
@samuelhwilliams @galund Have updated to use the bumped utils package.

requirements-app.txt
```diff
@@ -6,6 +6,7 @@ Flask-Login==0.2.11
 Flask-WTF==0.12
 lxml==3.8.0
 
-git+https://github.com/alphagov/digitalmarketplace-utils.git@29.0.0#egg=digitalmarketplace-utils==29.0.0
+git+https://github.com/alphagov/digitalmarketplace-utils.git@31.2.1#egg=digitalmarketplace-utils==31.2.1
 git+https://github.com/alphagov/digitalmarketplace-content-loader.git@4.4.1#egg=digitalmarketplace-content-loader==4.4.1
 git+https://github.com/alphagov/digitalmarketplace-apiclient.git@14.0.0#egg=digitalmarketplace-apiclient==14.0.0
+git+https://github.com/alphagov/odfpy.git@ee4482a#egg=odfpy==1.3.6dev
```
katstevens: (Just an FYI for reviewers that we had to pull the `odfpy` in due to a dependency in the utils package.)

galund: Yeah I feel bad about that a bit still :)

PR #705, opened 2018-01-29
search page: add aria-live attributes

risicle:
Trello https://trello.com/c/qUViIfu9/172-dynamic-content-on-the-search-pages-which-updates-without-a-page-reload-is-marked-up-using-wai-aria-live-regions

We don't want the whole loaded results section re-read on reloading, instead just a shortened "N results found" summary.

The element referenced by an `aria-controls=` seems to need to be persistent, meaning we probably can't use a fragment of the existing search summary text, as it gets removed & replaced wholesale when a new set of results is fetched.

Instead, we return the shortened summary as its own html fragment under the summary-accessible-hint key in the live results request. This html fragment gets placed inside a dedicated (and more importantly, persistent) `search-summary-accessible-hint-wrapper` element, which itself can be the target of the `aria-controls=`

This PR has fixed the tests to *pass* but hasn't added any new tests for the new functionality. I thought adding new tests would actually be best done in functional tests to ensure the ajax procedure doesn't screw things up.

Depends on https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/401 - ~~~~TODO fix dependency reference before merge.~~~~

### DIFF
--- a/app/templates/search/_summary_accessible_hint.html
+++ b/app/templates/search/_summary_accessible_hint.html
@@ -1,0 +1,3 @@
+<div id="js-dm-live-search-summary-accessible-hint">
+  {{ total }} {{ pluralize(total, "result", "results") }} found
+</div>
--- a/app/templates/search/briefs.html
+++ b/app/templates/search/briefs.html
@@ -61,6 +61,18 @@
     </section>
   </form>
   <section class="column-two-thirds" aria-label="Search results">
+    {#
+      the element referenced by an `aria-controls=` seems to need to be persistent -
+      at least I wasn't able to get it to work by just re-using a fragment of the
+      existing search summary text, as it gets removed and replaced wholesale by
+      the javascript when a new set of results is fetched.
+
+      instead, we have this dedicated (and more importantly, persistent) wrapper
+      element, which itself can be the target of `aria-controls=`
+    #}
+    <div id="search-summary-accessible-hint-wrapper" class="search-summary-accessible-hint-wrapper" aria-atomic="true" aria-live="polite" aria-relevant="all">
+      {% include 'search/_summary_accessible_hint.html' %}
+    </div>
     {% include 'search/_summary.html' %}
     {% include 'search/_results_wrapper.html' %}
   </section>
--- a/app/templates/search/services.html
+++ b/app/templates/search/services.html
@@ -50,6 +50,18 @@
     </section>
   </form>
   <section class="column-two-thirds" aria-label="Search results">
+    {#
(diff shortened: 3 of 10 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "gulp-jasmine-phantom": "3.0.0",
     "jquery": "1.12.0",
     "hogan.js": "3.0.2",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v27.0.0",
+    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#ris-search-summary-accessible-hint",
```
katstevens: I'm assuming this is waiting for the new utils version to be merged?

risicle: Yip

risicle:
Duh - I've forgotten to include the accessible hint information in the briefs ajax response...

app/templates/search/briefs.html
```diff
@@ -61,6 +61,9 @@
     </section>
   </form>
   <section class="column-two-thirds" aria-label="Search results">
+    <div id="search-summary-accessible-hint-wrapper" class="search-summary-accessible-hint-wrapper" aria-atomic="true" aria-live="polite" aria-relevant="all">
+      {% include 'search/_summary_accessible_hint.html' %}
```
galund: 2 space indents pls

app/templates/search/_summary_accessible_hint.html
```diff
@@ -0,0 +1,3 @@
+<div id="js-dm-live-search-summary-accessible-hint">
+    {{ total }} {{ pluralize(total, "result", "results") }} found
```
galund: This is where I wanted the two-space indent(!)

risicle: huh yeah not sure why my editor didn't. sometimes it doesn't quite detect the syntax properly.

app/templates/search/services.html
```diff
@@ -50,6 +50,9 @@ <h1>Search results</h1>
     </section>
   </form>
   <section class="column-two-thirds" aria-label="Search results">
+    <div id="search-summary-accessible-hint-wrapper" class="search-summary-accessible-hint-wrapper" aria-atomic="true" aria-live="polite" aria-relevant="all">
```
galund: Comment here pls :)

risicle:
Just pushed a small extra commit which gets causes the screen reader to say "Clear filters button" rather than "Clear filters visited link".

risicle:
Sigh. Overly-literal tests...

PR #710, opened 2018-02-07
Also make form actions absolute in /404 page

TheDoubleK:
Final fix for this ticket: https://trello.com/c/2UzWy7qK/4-404-page-on-assets-has-header-links-that-are-broken

The feedback form action `/feedback` also needs to be made absolute for it to work properly when this page is proxied from a domain other than the main frontend domain.

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -95,8 +95,13 @@ def external_404():
     """
     document = html.fromstring(render_template('errors/404.html'))
     relative_links = document.xpath('//a[starts-with(@href, "/")]')
+    forms_with_relative_actions = document.xpath('//form[starts-with(@action, "/")]')
+
     for link in relative_links:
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
Is this something that is fundamentally fragile? 🤔  I suspect this does, as you say, work fine for everything we have now, but could it break in the future?

https://stackoverflow.com/questions/2725156/complete-list-of-html-tag-attributes-which-have-a-url-value

TheDoubleK:
I talked with Chris about this a bit yesterday.  Somehow the stylesheets are loading OK at the moment, and when I applied the "full URL" patch to them things broke.

If we were doing this in a lot of places I'd be worried.  But it isn't a pattern that should spread, and is only for this one error page, which I suspect isn't going to have iframes, applets, audio or video  on it any time soon.

PR #714, opened 2018-02-15
Update Dockerfile to use v3.1.1, which pins top-level dependencies.

samuelhwilliams:
## Ticket
https://trello.com/c/Y6H3QWlM/315-uwsgi-version-should-be-pinned


TheDoubleK:
Also needs a rebase already 😞

PR #722, opened 2018-03-07
Add link to MISO spend data on the homepage

brenetic:
We would like to make it easier for users to see spend data about the marketplace.

https://trello.com/c/M331O7MN/370-add-link-to-govuk-miso-spend-data-on-dm-homepage

## Before

screen shot 2018-03-07 at 14 25 16

## After

screen shot 2018-03-08 at 13 52 15

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -100,6 +100,13 @@
       </div>
     {% endif %}
 
+      <div class="padding-bottom-small">
+        <p>
+          <a href="https://www.gov.uk/government/collections/digital-marketplace-sales" class="top-level-link">
+            See Digital Marketplace sales figures
+          </a>
+        </p>
+      </div>
     </aside>
   </div>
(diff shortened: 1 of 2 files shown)


app/templates/index.html
```diff
@@ -100,6 +100,13 @@ <h2>Sell services</h2>
       </div>
     {% endif %}
 
+      <div class="padding-bottom-small">
+        <p>
+          <a href="https://www.gov.uk/government/collections/digital-marketplace-sales" class="top-level-link" target="_blank">
```
katstevens: Looking at other external links on the homepage, we don't tend to open in a new tab. I guess it could be less accessible?

brenetic: Okey dokey, have updated

brenetic:
@karlchillmaid Have updated 🚀

PR #723, opened 2018-03-07
Show award status in DOS search results

brenetic:
We now display the status of opportunities in DOS search results with something more accurate than "closed". This should make it more obvious to suppliers what has happened with the brief.

## Before

screen shot 2018-03-07 at 16 47 35

## After

screen shot 2018-03-07 at 16 44 28

screen shot 2018-03-07 at 16 45 07

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -288,6 +288,12 @@ def list_opportunities(framework_family):
         framework_family=framework['framework'],
         framework_family_name='Digital Outcomes and Specialists',
         lot_names=tuple(lot['name'] for lot in lots_by_slug.values() if lot['allowsBrief']),
+        outcome={
+            'awarded': 'awarded',
+            'cancelled': 'cancelled',
+            'closed': 'awaiting outcome',
+            'unsuccessful': 'no suitable suppliers'
+        },
         pagination=pagination_config,
         search_keywords=get_keywords_from_request(request),
         search_query=search_query,
--- a/app/templates/search/_briefs_results.html
+++ b/app/templates/search/_briefs_results.html
@@ -27,7 +27,7 @@
     <ul class="search-result-metadata">
         {% if brief.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
             <li class="search-result-metadata-item">
-                {{ brief.status | capitalize() }}
+                Closed: {{ outcome[brief.status] }}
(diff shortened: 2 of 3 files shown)


brenetic:
Anyone have a suggestion for a better name than `embellished_statuses`?

karlchillmaid:
`award_status` – or has that been used already?

katstevens:
I'd say `outcome` or something like that - the brief may not end in an award if it's cancelled/unsuccessful.

app/templates/search/_briefs_results.html
```diff
@@ -27,7 +27,7 @@ <h2 class="search-result-title">
     <ul class="search-result-metadata">
         {% if brief.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
             <li class="search-result-metadata-item">
-                {{ brief.status | capitalize() }}
+                Closed: {{ outcome.get(brief.status, brief.status) }}
```
katstevens: Having the `Closed: foo` prefix feels a bit weird to me? (Disclaimer: I am not a content person)

app/main/views/marketplace.py
```diff
@@ -288,6 +288,10 @@ def list_opportunities(framework_family):
         framework_family=framework['framework'],
         framework_family_name='Digital Outcomes and Specialists',
         lot_names=tuple(lot['name'] for lot in lots_by_slug.values() if lot['allowsBrief']),
+        outcome={
+            'closed': 'awaiting outcome',
+            'unsuccessful': 'no suitable suppliers'
```
katstevens: I know it's a bit verbose to have the `awarded` and `cancelled` ones here too, but if we want to change the content in the future it might be good to have all the strings in one place? Then we can just use `outcome[brief.status]` in the template, without the `.get`.

katstevens: Better yet, we could set the content dict in the `_briefs_results.html` template rather than the view 😸

app/templates/search/_briefs_results.html
karlchillmaid: Yeah, it's there so there's some consistency between the Open/Closed filter and the listings.

If we change the filter to include all the statuses at some point, I plan to review this content and probably drop the `Closed:` prefix.

brenetic:
@katstevens Updated 🚀

PR #726, opened 2018-03-09
fix reference to data_api_client.get_direct_award_project_services_iter to work with digitalmarketplace-apiclient 15.0.0

risicle:
This was originally missed because of a mistake in the apiclient's version bumping

Also uncovered a problem with the mocking of `data_api_client` used in `tests/main/views/test_direct_award.py` which needs fixing but don't have time to now. Tech debt ticket for this: https://trello.com/c/imymWr7B/387-insufficient-mocking-of-dataapiclient-in-buyer-frontend-tests-could-be-dangerous


brenetic:
🚀

PR #728, opened 2018-03-13
Use frameworks version 10.8.0

brenetic:
Version 10.8.0 of the frameworks contains the new mapping transformations for the briefs search filtering on 'open' and 'closed'.

The reason for this change is to allow the original pre transformed status of a brief to be stored in Elasticsearch while maintaining the ability to filter on 'open' and 'closed'. A future PR will surface the
actual status in the search results for users.

The parameters for filtering have changed:

## Before

`/opportunities?q=&doc_type=briefs&status=live`

`/opportunities?q=&doc_type=briefs&status=closed`

## After

`/opportunities?q=&doc_type=briefs&statusOpenClosed=open`

`/opportunities?q=&doc_type=briefs&statusOpenClosed=closed`

Related frameworks PR: https://github.com/alphagov/digitalmarketplace-frameworks/pull/506

### DIFF
--- a/app/main/views/marketplace.py
+++ b/app/main/views/marketplace.py
@@ -195,8 +195,11 @@ def list_opportunities(framework_family):
     doc_type = 'briefs'
     updated_request_args = None
 
-    if 'status' not in clean_request_query_params.keys():
-        updated_request_args = MultiDict([('status', 'live'), ('status', 'closed')])
+    # This will exclude anything with a 'withdrawn' status
+    if 'statusOpenClosed' not in clean_request_query_params.keys():
+        updated_request_args = MultiDict(
+            [('statusOpenClosed', 'open'), ('statusOpenClosed', 'closed')]
+        )
         updated_request_args.update(clean_request_query_params)
 
     search_api_response = search_api_client.search(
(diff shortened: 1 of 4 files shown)


katstevens:
Some failing unit tests here (will the functional tests need to be updated too?)

brenetic:
@galund @katstevens Prematurely tagged you for review, soz! Tests are fixed up now.

app/main/views/marketplace.py
```diff
@@ -195,8 +195,10 @@ def list_opportunities(framework_family):
     doc_type = 'briefs'
     updated_request_args = None
 
-    if 'status' not in clean_request_query_params.keys():
-        updated_request_args = MultiDict([('status', 'live'), ('status', 'closed')])
+    if 'statusOpenClosed' not in clean_request_query_params.keys():
```
galund: Can we add a comment here about the reason for this being to exclude "withdrawn" briefs? (cc @idavidmcdonald @Wynndow in case they know any other reasons, but I think we decided that seems to be the main point).

idavidmcdonald: I believe not showing withdrawn briefs was a conscious product decision (as they are no long relevant and could be misleading). However we still let the URL for a withdrawn brief exist because otherwise things would just disappear with no context

PR #729, opened 2018-03-13
Button too close to links on mobile

DilwoarH:
The 'Apply for this Opportunity' button (Buyer FE) is too close to the 'Log in to ask a question' link on mobile.

This bug may be repeated elsewhere in the FE apps!

(This happens when a form is created with just one element [submit button] )

https://trello.com/c/OYFHcH77/139-button-too-close-to-links-on-mobile

Before:
screenshot_20171219-170531

After:
screen shot 2018-03-13 at 16 07 48

### DIFF
--- a/app/assets/scss/_brief-page.scss
+++ b/app/assets/scss/_brief-page.scss
@@ -1,0 +1,7 @@
+.brief-page {
+  .brief-action {
+    @include media(mobile) {
+      margin-top: 1.5em;
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
It feels like it would be nice to investigate this a bit more and discover the problem/come up with a solution. This feels like a stopgap patch to cover one particular scenario. Or do we know this is the only occurrence of this issue?

PR #730, opened 2018-03-14
Bugfix 142 search clear filters behaviour with keyword search

benvand:
https://trello.com/c/RIlQOOAm/142-search-clear-filters-behaviour-with-keyword-search

In g-cloud search, if you initially search for "apricot" in the keyword search and then on the search screen change the keyword search to "banana" and hit return, the "Clear filters" link still points to a url with "apricot" in it. However if instead of hitting return you click the search button it does update the href. Which is presumably the desired behaviour.

The problem is:
Searching via the form on the page disables full page reloads and instead replaces specific sections of the page dynamically. The ‘clear filters’ link was not included as one of the parts of the page that was reloaded, meaning that the URL was not updating after a live search

The fix is this:
The way to reload part of the content  inside the `dm-filters` div is to create a new div (`js-dm-live-filter-title`) which _is_ repopulated on `live_search` reload. To do this we specify the new div name and its content in the view. The new content is then injected into the `js-dm-live-filter-title` div on each `live_search` reload.

### DIFF
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -1775,12 +1775,13 @@ class TestCatalogueOfBriefsFilterOnClick(BaseApplicationTest):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities?live-results=true')
         data = json.loads(res.get_data(as_text=True))
 
-        assert set(data.keys()) == {
+        assert sorted(data.keys()) == sorted((
             'results',
             'summary',
             'summary-accessible-hint',
             'categories',
-        }
+            'filter-title'
+        ))
 
         for k, v in data.items():
             assert set(v.keys()) == {'selector', 'html'}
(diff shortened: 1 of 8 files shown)


tests/main/views/test_marketplace.py
```diff
@@ -1775,33 +1775,42 @@ def test_live_results_returns_valid_json_structure(self):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities?live-results=true')
         data = json.loads(res.get_data(as_text=True))
 
-        assert set(data.keys()) == {
+        assert sorted(data.keys()) == sorted((
```
risicle: is this just because it gives better pytest output on failure?

benvand: Na I was thinking it probably _shouldn't_ pass if it ends up with something like:
```
set(('foo', 'bar', 'foo')) == {'foo', 'bar'}
```
I'm guessing `set` was used because of the implicit sorting but it could result in false positives.

risicle: Oh I get it! It's about repetitions. Cool.

benvand:
https://github.com/alphagov/digitalmarketplace-functional-tests/pull/482
;)

PR #732, opened 2018-03-14
Add award status details to dos search

brenetic:
Once more with feeling. This was originally done in PR https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/723 but changes needed to happen to the fields that were being indexed by the search API first.

We now display the status of opportunities in DOS search results with something more accurate than "closed". This should make it more obvious to suppliers what has happened with the brief.

## Before

screen shot 2018-03-07 at 16 47 35

## After

screen shot 2018-03-07 at 16 44 28

screen shot 2018-03-07 at 16 45 07

### DIFF
--- a/app/templates/search/_briefs_results.html
+++ b/app/templates/search/_briefs_results.html
@@ -25,9 +25,9 @@
     </ul>
 
     <ul class="search-result-metadata">
-        {% if brief.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
+        {% if brief.status in outcomes.keys() %}
             <li class="search-result-metadata-item">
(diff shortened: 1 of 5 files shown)


app/templates/search/_briefs_results.html
```diff
@@ -25,9 +25,9 @@ <h2 class="search-result-title">
     </ul>
 
     <ul class="search-result-metadata">
-        {% if brief.status in ['closed', 'awarded', 'cancelled', 'unsuccessful'] %}
+        {% if brief.status in outcomes.keys() %}
```
galund: The `.keys` thing here isn't actually necessary, I think

PR #736, opened 2018-03-22
Update https://www.digitalmarketplace.service.gov.uk/cookies

DilwoarH:
User research feature added another field on the cookie.
We need to add this to the cookies page

Ticket: https://trello.com/c/V8UI0k8H/146-update-https-wwwdigitalmarketplaceservicegovuk-cookies

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -103,7 +103,7 @@
             </tr>
             <tr>
                 <td>seen_cookie_message</td>
-                <td>Saves a message to let us know that you have seen our cookie message</td>
+                <td>Tells us that you have seen our cookie message</td>
                 <td>1 month&nbsp;</td>
             </tr>
             </tbody>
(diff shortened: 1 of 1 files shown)


app/templates/content/cookies.html
```diff
@@ -106,6 +106,11 @@ <h3>Introductory message cookie</h3>
                 <td>Saves a message to let us know that you have seen our cookie message</td>
                 <td>1 month&nbsp;</td>
             </tr>
+            <tr>
+                <td>seen_user_research_message</td>
```
idavidmcdonald: Should this be put in it's section rather than a new row added to this section?

karlchillmaid: Suggested content change (would need to be checked by the lawyer):

```Tells us that you have seen our cookie message```

DilwoarH:
screen shot 2018-03-26 at 11 47 05

@constancecerf @karlchillmaid - could you have a read.

This is how the banner looks:
http://alphagov.github.io/digitalmarketplace-frontend-toolkit/user-research-consent-banner.html

constancecerf:
Change 'User research banner cookie' to 'User research recruitment banner cookie'

REASON

It's for recruitment rather than user research itself

Change 'When you first use our service' to 'When you use our service'

REASON

It may not be the first time they use the service.

Change 'you may see a pop up 'User research banner' message' to 'you may see a pop up banner about user research recruitment'

REASON

There is no specific name for this banner so we don't need it in inverted commas

DilwoarH:
@constancecerf - updated

DilwoarH:
@idavidmcdonald @Wynndow - AWS cookies removes
@idavidmcdonald - the user banner is separated now into it's own section
@karlchillmaid - updated cookies purpose

PR #737, opened 2018-03-26
bump digitalmarketplace-utils dependency to 35.4.0, set DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS for production & staging

risicle:
Trello https://trello.com/c/YQB2HkiD/225-functional-tests-send-emails-to-invalid-email-addresses

Did I remember correctly that we decided we should set this for production and staging but preview will probably be fine as it should have a sandbox token anyway...?

### DIFF
--- a/config.py
+++ b/config.py
@@ -142,6 +142,12 @@ class Live(Config):
     DM_LOG_PATH = '/var/log/digitalmarketplace/application.log'
     DM_HTTP_PROTO = 'https'
 
+    # use of invalid email addresses with live api keys annoys Notify
+    DM_NOTIFY_REDIRECT_DOMAINS_TO_ADDRESS = {
+        "example.com": "simulate-delivered@notifications.service.gov.uk",
+        "example.gov.uk": "simulate-delivered-2@notifications.service.gov.uk",
+    }
+
 
 class Preview(Live):
     FEATURE_FLAGS_NEW_SUPPLIER_FLOW = enabled_since('2017-02-06')
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -6,6 +6,6 @@ Flask-Login==0.2.11
 Flask-WTF==0.12
 lxml==3.8.0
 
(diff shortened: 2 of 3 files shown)


config.py
```diff
@@ -96,6 +96,13 @@ def init_app(app):
         app.jinja_loader = jinja_loader
 
 
+class _NotifyRedirectDomainsMixin(object):
```
samuelhwilliams: This mixin feels like a fair candidate for utils, as presumably it'll be in all the frontend apps?

risicle: Iiiiinteresting - hadn't considered that...

risicle: Then again, there are *numerous* settings in here for which we could say that about, but we don't do any of the like for them.

risicle: (I was more concerned people wouldn't like the mixin approach - the first way I was doing it was setting it on `Live` and then explicitly re-setting it back to `None` on `Preview`. Don't know which is better)

samuelhwilliams: I guess it's a bit weird to have a single class just to pull in one variable.

I'm not opposed to _more_ of the config going to utils I guess. It's probably all pretty similar.

But I guess for now you're right that it's odd and it's probably just easier to put it in the config itself.

Wait. Don't we only want it _set_ on Preview, and don't really need it set on staging/production? It's our test environments where we want to suppress emails.

risicle: Arguably, we don't *ever* want to try sending emails to `example.com` or `example.gov.uk` domains. And wouldn't it be good to be able to test notify integration on live?

And preview, shouldn't it have a sandbox key anyway so it shouldn't matter?

samuelhwilliams: This whole thing started because we don't use a sandbox key on preview and Notify are having issues with us trying to send these emails to invalid addresses on a regular basis. We need to stop doing that, like, soon ideally.

samuelhwilliams: Doesn't matter to me if we change over to a sandbox key or if we implement this - but my understand was that this code was for our preview/staging environments, not as much for production.

risicle: Whatever stage it is, we have the necessity to have a stage sending out *real* notify emails, but still be able to use the functional tests on that stage without pissing off notify.

PR #738, opened 2018-03-27
Remove hidden doc_type field from search results

brenetic:
We are able to use the 'Destination Page' dimension can be used against the 'Search Term' dimension by filtering by either '/g-cloud/' or '/digital-outcomes-and-specialists/' which provides more reliable data.

https://trello.com/c/8RzdZPsZ/393-buyer-frontend-search-page-doctype-parameter-is-broken-and-should-be-removed


tests/main/views/test_search.py
```diff
@@ -571,7 +571,6 @@ def test_filter_form_given_category_selection(self):
             'lot': 'cloud-software',
             'serviceCategories': 'content management system (cms)',
             'parentCategory': 'electronic document and records management (edrm)',
-            'doc_type': 'services',
```
katstevens: 😿 that we don't have the equivalent test for DOS (that's an ongoing sadface of mine)

app/main/views/g_cloud.py
```diff
@@ -246,15 +245,10 @@ def search_services():
...
-            # (If a sub-category is selected, we also need the parent category id, so that the correct part
-            # of the category tree is displayed when the sub-category appears under multiple parents.)
-            selected_category_tree_filters[1:],
-            # add "doc_type" to query string for analytics disambiguation
-            ({"name": "doc_type", "value": doc_type},),
-        )
+        # Filter form should also filter by lot, and by category, when any of those are selected.
+        # (If a sub-category is selected, we also need the parent category id, so that the correct part
+        # of the category tree is displayed when the sub-category appears under multiple parents.)
+        f["name"]: f for f in selected_category_tree_filters[1:]
```
katstevens: Nice to get rid of `chain` and simplify things 👍

PR #745, opened 2018-04-17
Layout tweaks on saved search task list page

DilwoarH:
The main heading on the task list page has too much vertical space at its top margin. This is especially bad when you first create the search, where the flash message directly above seems to have too much space at its bottom margin.

Ticket: https://trello.com/c/z1RNJnQy/18-layout-tweaks-on-saved-search-task-list-page

Required PR: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/433

Before:
screen_shot_2018-01-15_at_18 18 45

After:
screen shot 2018-04-17 at 13 59 40

### DIFF
--- a/app/assets/scss/_index-page.scss
+++ b/app/assets/scss/_index-page.scss
@@ -2,6 +2,10 @@
 
 .index-page {
 
+  .flash-message-container {
+    margin-bottom: 40px; // adds in specific margin for index page due to layout of page
+  }
+
   h2 {
     @include bold-27;
     margin-bottom: $gutter;
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -12,9 +12,11 @@
 
 {% block phase_banner %}{% endblock %}
 
+ {% block body_classes %}index-page{% endblock %}
+
 {% block main_content %}
 
-<div class="index-page grid-row">
+<div class="grid-row">
   <div class="column-two-thirds">
     <h2>Find technology or people for digital projects in the public sector</h2>
       {% set items = [] %}
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "gulp-jasmine-phantom": "3.0.0",
     "jquery": "1.12.0",
     "hogan.js": "3.0.2",
-    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v28.10.2",
+    "digitalmarketplace-frontend-toolkit": "https://github.com/alphagov/digitalmarketplace-frontend-toolkit.git#v28.13.2",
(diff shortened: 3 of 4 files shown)


app/templates/direct-award/view-project.html
```diff
@@ -36,6 +36,8 @@
   {% endwith %}
 {% endblock %}
 
+{% block body_classes %}direct-award-project-overview-page{% endblock %}
+
```
CrystalPea: Ahh, interesting, what does this block do, how does it work?

DilwoarH: this block adds a class to the body tag, have a look at this template:
https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L38

I needed this as the flash message which I was trying to target was inside the body not the content block. As I did not want the change to affect every flash message, I had to add a specific class to keep the scope to that page. 

Hope this makes sense. :)

app/assets/scss/_search_overview.scss
```diff
@@ -1,4 +1,7 @@
 .direct-award-project-overview-page {
+  .flash-message-container {
+    margin-bottom: 0px;
```
galund: For 0, don't use any units

app/templates/direct-award/view-project.html
galund: Is there really no more general case that needs tackling here?

For example, maybe: - other pages that have headings? Other pages that have task lists?

It just seems like there isn't / shouldn't be much that's special about this page. All pages may have flash messages.

DilwoarH: Hey george - I was thinking of doing something more generic but didn't because of the following reasons:
- Removing the margin-bottom completely might affect pages where there isn't a header at the top.
- There may be already specific css to handle the issue on other pages.
- Not sure how we would apply specific CSS for only when headers elements are after the flash message.
- A lot of effort for a small styling issue.

DilwoarH: Hey George, 

Looked through the apps, the margin-bottom isn't really needed anywhere apart from the homepage.

I will therefore probably remove it from the template and add some custom margin for the homepage.

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/commit/3d2af0544db2c0d77c0788d128d49ed4adcff751

screen shot 2018-04-20 at 16 12 41

galund: Sounds sensible. The wins for CSS simplification work are generally small wins eh :-)

DilwoarH:
@galund - adds margin on index page

Related PR:
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/433

DilwoarH:
NEED TO YARN INSTALL AFTER TOOLKIT CHANGES HAVE BEEN MERGED ⚠️⚠️⚠️

PR #747, opened 2018-04-18
Dh 110 instrument ga tracking to capture clicks made on downloads and external links

DilwoarH:
A lot of GA tracking comes 'out of the box' but a notable exception is tracking to capture clicks on downloads and external links. Additional code is required in order to capture these.

Currently, we are tracking some downloads and external links on the DM site (those that were specifically required for previous mission work).

The proposal is to have all downloads and external links tracked automatically across the whole site, similar to how GOV.UK have done.

The advantage of doing this is that we have key data available straight away at the start of mission teams rather than having to prioritise the work, use up dev resource and then wait for enough data to populate before being able to carry out analysis.

---> includes making sure all links on the homepage are tracked, so we can measure which links are used the most and least.

Ticket: https://trello.com/c/i912cDBe/110-instrument-ga-tracking-to-capture-clicks-made-on-downloads-and-external-links

### DIFF
--- a/app/assets/javascripts/analytics/_init.js
+++ b/app/assets/javascripts/analytics/_init.js
@@ -7,5 +7,6 @@
     this.events.init();
     this.virtualPageViews.init();
     this.scrollTracking.init();
+    this.trackExternalLinks.init();
   };
 })(window);
--- a/spec/javascripts/manifest.js
+++ b/spec/javascripts/manifest.js
@@ -6,6 +6,7 @@ var manifest = {
     '../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/google-analytics-universal-tracker.js',
     '../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/analytics/analytics.js',
     '../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/analytics/_register.js',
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/analytics/_init.js
```diff
@@ -7,5 +7,6 @@
     this.events.init();
     this.virtualPageViews.init();
     this.scrollTracking.init();
+    this.trackExternalLinks.init();
```
katstevens: If we add the new module to the `_init.js` we also need to add it to the spec `manifest.js` (though there may well be a better way of doing this?) otherwise the JS tests will fail.

DilwoarH: hey @katstevens  - I have updated the tests

galund:
Is there not now some link-tracking code that we could now lose as a result of this being automatic?

DilwoarH:
yeah - there are other stats for other links - this is just a generic tracking for everything (external)

katstevens:
It would be great if we could delete any (or all!) of the custom download tracking in `_events.js`. `supplierListDownload` maybe?

DilwoarH:
@galund - Spoke to ash, he wants to leave all the other ones as they are. 
@katstevens - moving the events code is another ticket I plan to start working on. It doesn't really fit into this ticket.

PR #749, opened 2018-04-18
Buyer FE test refactor

katstevens:
Part of the ongoing test refactor (see Briefs FE equivalent for details https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/93)

- Removes unnecessary `app.app_context` managers
- Python3 `super()` syntax
- Standardising `setup()` and `teardown()` methods across the tests
- Patch the `data_api_client` in the test class setup method. This is split into per-file commits for ease of reviewing.

This also required refactoring one of the helper functions `get_direct_award_projects` to accept the `data_api_client` instance as an argument, to avoid importing it twice  (and thus having to patch it twice). This is consistent with patterns used on the other helper functions.

### DIFF
--- a/tests/main/presenters/test_search_results.py
+++ b/tests/main/presenters/test_search_results.py
@@ -39,13 +39,16 @@ class TestSearchResults(BaseApplicationTest):
         return False
 
     def setup_method(self, method):
-        super(TestSearchResults, self).setup_method(method)
+        super().setup_method(method)
 
         self.fixture = _get_fixture_data()
         self.multiple_pages_fixture = _get_fixture_multiple_pages_data()
         self._lots_by_slug = framework_helpers.get_lots_by_slug(
             self._get_framework_fixture_data('g-cloud-6')['frameworks']
         )
+
+    def teardown_method(self, method):
+        super().teardown_method(method)
 
     def test_search_results_is_set(self):
         search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
--- a/tests/main/presenters/test_search_results.py
+++ b/tests/main/presenters/test_search_results.py
@@ -47,9 +47,6 @@ class TestSearchResults(BaseApplicationTest):
             self._get_framework_fixture_data('g-cloud-6')['frameworks']
         )
 
-    def teardown_method(self, method):
-        super().teardown_method(method)
-
     def test_search_results_is_set(self):
         search_results_instance = SearchResults(self.fixture, self._lots_by_slug)
         assert hasattr(search_results_instance, 'search_results')
--- a/tests/main/presenters/test_service_presenters.py
+++ b/tests/main/presenters/test_service_presenters.py
@@ -32,7 +32,7 @@ def test_chunk_string():
 
 class TestService(BaseApplicationTest):
     def setup_method(self, method):
-        super(TestService, self).setup_method(method)
+        super().setup_method(method)
 
         self.fixture = _get_fixture_data()
         self.fixture = self.fixture['services']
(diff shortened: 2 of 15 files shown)


tests/main/presenters/test_service_presenters.py
```diff
@@ -90,7 +93,7 @@ def test_service_properties_available_via_summary_manifest(self):
         assert len(service.summary_manifest.sections[0].questions) == 5
 
 
-class TestMeta(object):
+class TestMeta:
```
benvand: Nice catch

tests/main/presenters/test_search_results.py
```diff
@@ -39,14 +39,17 @@ def _get_service_result_by_id(self, results, id):
...
-        super(TestSearchResults, self).setup_method(method)
+        super().setup_method(method)
 
         self.fixture = _get_fixture_data()
         self.multiple_pages_fixture = _get_fixture_multiple_pages_data()
         self._lots_by_slug = framework_helpers.get_lots_by_slug(
             self._get_framework_fixture_data('g-cloud-6')['frameworks']
         )
 
+    def teardown_method(self, method):
```
benvand: These aren't required to cause `BaseApplicationTest.teardown_method` to run. If it's not defined here `TestSearchResults.teardown_method` will be `BaseApplicationTest.teardown_method`.

```
class A:
    
    def foo():
        pass
    
class B(A):
    pass
    
print(B.foo)

> <function A.foo at 0x1054f2950>

```

tests/main/views/test_marketplace.py
```diff
@@ -13,7 +13,10 @@
 
 class TestApplication(BaseApplicationTest):
     def setup_method(self, method):
```
benvand: Both of these can go

tests/main/views/test_marketplace.py
```diff
@@ -374,19 +401,20 @@ def test_external_404_makes_all_links_absolute(self):
...
-        self._data_api_client = self._data_api_client_patch.start()
+        self.data_api_client_patch = mock.patch('app.main.views.marketplace.data_api_client', autospec=True)
+        self.data_api_client = self.data_api_client_patch.start()
 
         self.brief = self._get_dos_brief_fixture_data()
         self.brief_responses = self._get_dos_brief_responses_fixture_data()
         self.brief_id = self.brief['briefs']['id']
-        self._data_api_client.get_brief.return_value = self.brief
-        self._data_api_client.find_brief_responses.return_value = self.brief_responses
+        self.data_api_client.get_brief.return_value = self.brief
```
benvand: 💯

tests/main/views/test_search.py
```diff
@@ -443,8 +450,11 @@ def test_search_results_show_aggregations_by_parent_category(self):
             assert expected_lot_counts[category_name] == int(number_of_services)
 
     def test_search_results_sends_aggregation_request_without_page_filter(self):
-        data_api_client.find_frameworks.return_value = {'frameworks': [self._get_framework_fixture_data('g-cloud-9')
-                                                                       ['frameworks']]}
+        self.data_api_client.find_frameworks.return_value = {
+            'frameworks': [
+                self._get_framework_fixture_data('g-cloud-9')['frameworks']
```
benvand: Not sure there's much value in this having its own line. Could be 
```
self.data_api_client.find_frameworks.return_value = {
            'frameworks': [self._get_framework_fixture_data('g-cloud-9')['frameworks']]
        }
```

tests/main/presenters/test_search_results.py
katstevens: Ah, good spot - I had blanket-added equivalent `teardown`s where there was a `setup`, but this one didn't end up having any API client patching. Will remove

tests/main/views/test_search.py
katstevens: Yeah... I only did it because I dislike having the `]]` on the end 😸 Happy to change.

PR #752, opened 2018-04-26
Update cookies page to include new GA cookie in line with gov.uk

benvand:
https://trello.com/c/RNzdFVME/188-update-dm-cookies-with-gid-from-ga-inline-with-govuk

It looks like GA now has an _gid key for it's cookies. We need to add that to https://www.digitalmarketplace.service.gov.uk/cookies in line with
https://www.gov.uk/help/cookies

(Yes, it's the same as the one above, see here: https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage)

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -74,6 +74,12 @@
                 </td>
                 <td>2 years</td>
             </tr>
+            <tr>
+                <td>_gid</td>
+                <td>Helps us count how many people visit the Digital Marketplace by tracking if you’ve visited the service before
+                </td>
+                <td>24 hours</td>
+            </tr>
              <tr>
                 <td>_gat</td>
                 <td>Used to manage the rate at which page view requests are made


app/templates/content/cookies.html
```diff
@@ -73,6 +73,12 @@ <h3>Universal Analytics</h3>
                 <td>Helps us count how many people visit the Digital Marketplace by tracking if you’ve visited the service before
                 </td>
                 <td>2 years</td>
+            </tr>
+            <tr>
+                <td>_gid</td>
+                <td>Helps us count how many people visit the Digital Marketplace by tracking if you’ve visited the service before
```
CrystalPea: Should we maybe let users know more about how `_ga`cookie is different from `_gid`? How are they different generally? I mean, purpose stated for each is the same, but expiry time is different, why is that?

benvand: Google doesn't distinguish between them so we have no way of knowing how they're different (see the link in description). Looks like the accepted solution has been to keep the entries identical.
https://github.com/alphagov/govuk-design-system/blob/master/src/cookies.md.njk
https://github.com/alphagov/paas-product-page/blob/master/views/cookies.erb
https://www.gov.uk/help/cookies

PR #754, opened 2018-04-27
Fixes issue with index page css changing font size of report a problem

DilwoarH:
Related PR: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/745/files
Ticket: https://trello.com/c/0q1xiDvF/47-move-briefs-fe-flash-templates-into-base-template

### DIFF
--- a/app/templates/index.html
+++ b/app/templates/index.html
@@ -12,11 +12,11 @@
 
 {% block phase_banner %}{% endblock %}
 
- {% block body_classes %}index-page{% endblock %}
+ {% block body_classes %}index-page-body{% endblock %}
 
 {% block main_content %}
 
-<div class="grid-row">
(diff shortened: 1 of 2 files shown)


app/templates/index.html
```diff
@@ -12,11 +12,11 @@ <h1>
 
 {% block phase_banner %}{% endblock %}
 
- {% block body_classes %}index-page{% endblock %}
+ {% block body_classes %}index-page-body{% endblock %}
```
Wynndow: what does this block do? I can't find it anywhere else in the buyer frontend.

DilwoarH: @Wynndow - https://github.com/alphagov/govuk_template/blob/master/source/views/layouts/govuk_template.html.erb#L39

PR #755, opened 2018-05-11
FE-1 Did you award a contract for 'Buying some hosting'? page details

lfdebrux:
[Trello](https://trello.com/c/yIUrcLLo/73-fe-1-did-you-award-a-contract-for-buying-some-hosting-page-details)

1. Added "Did you award a contract?" page
2. Created a `DidYouAwardAContractForm` class for validating the radio input
3. Created a test suite `TestDirectAwardAwardContract` for the page, HTML form, and redirects
4. Created a route function `award_contract` that passes the tests

Currently as there are no routes for `which-service-won-contract` and `why-didnt-you-award-contract` the tests for those have been passed by sliming.

Dealing with the "We are still assessing services" flow is in ticket [#110](https://trello.com/c/5Lrp1lOK/110-fe-8-we-are-still-assessing-flow)

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -1,5 +1,6 @@
 from flask_wtf import FlaskForm
-from wtforms.validators import Length, Optional
+from wtforms.fields import RadioField
+from wtforms.validators import Length, Optional, InputRequired
 
 from dmutils.forms import StripWhitespaceStringField
 
@@ -12,5 +13,17 @@ class CreateProjectForm(FlaskForm):
                    max=100,
                    message="Names must be between 1 and 100 characters"),
             Optional()
+        ]
+    )
+
+
+class DidYouAwardAContractForm(FlaskForm):
+    did_you_award_a_contract = RadioField(
+        "Did you award a contract?",
+        validators=[InputRequired(message="You need to answer this question.")],
+        choices=[
+            ('yes', 'Yes'),
+            ('no', 'No'),
+            ('still-assessing', 'We are still assessing services')
         ]
     )
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -17,7 +17,7 @@ from dmutils.views import SimpleDownloadFileView
 
 from app import search_api_client, data_api_client, content_loader
 from ..exceptions import AuthException
-from ..forms.direct_award_forms import CreateProjectForm
+from ..forms.direct_award_forms import CreateProjectForm, DidYouAwardAContractForm
 from ..helpers.search_helpers import (
     get_keywords_from_request, pagination,
     get_page_from_request, query_args_for_pagination,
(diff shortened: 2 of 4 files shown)


lfdebrux:
Thanks @DilwoarH, I've addressed all your changes except the content change

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,62 @@ def end_search(framework_framework, project_id):
...
+            return redirect(url_for('.which_service_won_contract',
+                                    framework_framework=framework_framework,
+                                    project_id=project_id))
+        elif form.did_you_award_a_contract.data == 'no':
+            return redirect(url_for('.why_didnt_you_award_contract',
+                                    framework_framework=framework_framework,
+                                    project_id=project_id))
+        else:
+            # this else should never be reached
+            abort(500, "Unexpected answer to 'Did you award contract?'")
```
galund: I know Dilwoar mentioned this already, I think if you want to leave it in I'd be find with not bothering to have a message, `abort(500)` would be fine.

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,62 @@ def end_search(framework_framework, project_id):
...
+@direct_award.route(
+    '/<string:framework_framework>/projects/<int:project_id>/which-service-won-contract',
+    methods=['GET']
+)
+def which_service_won_contract(framework_framework, project_id):
+    pass
+
+
+@direct_award.route(
+    '/<string:framework_framework>/projects/<int:project_id>/why-didnt-you-award-contract',
```
galund: So I know it's kinda late to be mentioning this, but I really don't like URI's that look like they are missing punctuation. `.../non-award-reason` perhaps? cc @karlchillmaid

lfdebrux: Response from @karlchillmaid on Slack

> @george.lund @dilwoar.hussain re the URL question on GitHub — the URLs follow the standards for GOV.UK: https://insidegovuk.blog.gov.uk/url-standards-for-gov-uk/
The guidelines don't say anything about contractions, but when we discussed this at content crit we found other examples that do, eg. https://www.gov.uk/expenses-if-youre-self-employed
I agree that contractions look a bit funny without the apostrophe, but I think they're still readable, and it's probably more important to keep the URLs as close as possible to the page title and use logical, consistent, predictable patterns for URLs.

app/main/forms/direct_award_forms.py
```diff
@@ -14,3 +15,14 @@ class CreateProjectForm(FlaskForm):
             Optional()
         ]
     )
+
+class DidYouAwardAContractForm(FlaskForm):
+    did_you_award_a_contract = RadioField(
+        "Did you award a contract?",
+        validators=[InputRequired(message="You need to answer this question.")],
```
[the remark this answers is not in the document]
lfdebrux: I talked to @karlchillmaid about this this morning, he said that since the phrase is copied from another page he was okay with us keeping it as is for now for consistency.

However, he also said that it was a good question, and was going to bring it up at content crit to see if people had any thoughts on error messages, and possibly we will revisit this question across Digital Marketplace at a later time.

tl;dr let's merge it!

app/templates/direct-award/did-you-award-contract.html
```diff
@@ -0,0 +1,70 @@
...
+<div class="did-you-award-contract-page">
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <header class="page-heading">
+        <h1>
+          Did you award a contract for '{{project.name}}'?
+        </h1>
+      </header>
+
+      <form method="POST" action="{{ url_for('direct_award.did_you_award_contract', framework_framework=framework.framework, project_id=project.id) }}">
```
galund: It may be possible to use `action=""` and if so that would probably be preferable?

lfdebrux: According to [SO](https://stackoverflow.com/questions/1131781/is-it-a-good-practice-to-use-an-empty-url-for-a-html-forms-action-attribute-a) `action=""` _can_ cause issues.

I agree it is nicer though not to have a url in there. Is there any reason not just to leave the `action` attribute out altogether? Thoughts @DilwoarH?

DilwoarH: https://stackoverflow.com/a/9401608

Have a read of the above, I personally would leave it as it is as there may be some browsers which behave differently without an action. It's also standard to have an action target - removing it won't bring any additional benefit.

lfdebrux: Okay, I've left it as it is, but also added a test `test_award_contract_form_action_url_is_award_contract_url` that checks that the `action` url is the same as the GET url

benvand:
Routes that don't return a HTTP response cause 500s in flask. We should back this out. It's blocking the pipeline because it can't go to production as is.
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/760

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,61 @@ def end_search(framework_framework, project_id):
...
+)
+def did_you_award_contract(framework_framework, project_id):
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+
+    form = DidYouAwardAContractForm()
+    form.did_you_award_a_contract.options = [{"label": label, "value": value}
```
benvand: Dynamically editing the form object like this should be done in the `__init__` method of the form. The `.options` (or historically `.toolkit_macro_options`) attribute is something we're going to need every time we instantiate this form

PR #758, opened 2018-05-16
FE-2. Which service won the contract? page

DilwoarH:
Ticket: https://trello.com/c/hHyDVwFZ/72-fe-2-which-service-won-the-contract-page

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -1,7 +1,6 @@
 from flask_wtf import FlaskForm
 from wtforms.fields import RadioField
 from wtforms.validators import Length, Optional, InputRequired
-
 from dmutils.forms import StripWhitespaceStringField
 
 
@@ -39,3 +38,23 @@ class DidYouAwardAContractForm(FlaskForm):
             (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
+
+
+class WhichServiceWonTheContractForm(FlaskForm):
+    which_service_won_the_contract = RadioField(
+        "Which service won the contract?",
+        validators=[InputRequired(message="Please select the supplier which won the contract")],
+    )
+
+    def __init__(self, services, *args, **kwargs):
+        super(WhichServiceWonTheContractForm, self).__init__(*args, **kwargs)
+
+        self.which_service_won_the_contract.choices = [
+            (service["id"], service["data"]["serviceName"])
+            for service in services['services']]
+
+        self.which_service_won_the_contract.options = [{
+            "label": service["data"]["serviceName"],
+            "value": service["id"],
+            "description": service["supplier"]["name"]
+        } for service in services['services']]
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -17,7 +17,7 @@ from dmutils.views import SimpleDownloadFileView
 
 from app import search_api_client, data_api_client, content_loader
 from ..exceptions import AuthException
-from ..forms.direct_award_forms import CreateProjectForm, DidYouAwardAContractForm
+from ..forms.direct_award_forms import CreateProjectForm, DidYouAwardAContractForm, WhichServiceWonTheContractForm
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -77,3 +77,6 @@ app/static
 
 # local environment customizations
 local.nix
+
+# IDE files
+.vscode/
```
lfdebrux: This should probably be part of your global gitignore

app/templates/direct-award/which-service-won-contract.html
```diff
@@ -0,0 +1,82 @@
...
+          type = "save",
+          label = "Save and continue"
+        %}
+          {% include "toolkit/button.html" %}
+        {% endwith %}
+      </form>
+
+      {% else %}
+        <div class="dmspeak">
+          <p>You cannot award this contract as there were no services matching your requirements.</p>
```
lfdebrux: If there are no services, should we allow the user to select 'Yes' to 'Did you award a contract?'?

tests/main/views/test_direct_award.py
```diff
@@ -471,8 +471,11 @@ def test_end_search_redirects_to_project_page(self):
 class TestDirectAwardAwardContract(TestDirectAwardBase):
     def setup_method(self, method):
         super().setup_method(method)
-        self.data_api_client.get_direct_award_project.return_value['project']['lockedAt'] = \
-            "2017-08-30T07:49:26.677778Z"
+        self.data_api_client.get_direct_award_project.return_value = self._get_direct_award_lock_project_fixture()
```
lfdebrux: 👍 I should have guessed there would be an already existing fixture!

app/templates/direct-award/which-service-won-contract.html
```diff
@@ -0,0 +1,82 @@
...
+        "link": url_for('direct_award.saved_search_overview', framework_framework=framework.framework),
+        "label": "Your saved searches"
+      },
+      {
+        "link": url_for('direct_award.view_project', framework_framework=framework.framework, project_id=project.id),
+        "label": project.name
+      },
+      {
+        "link": url_for('direct_award.did_you_award_contract', framework_framework=framework.framework, project_id=project.id),
+        "label": "Did you award a contract"
```
lfdebrux: Might be nice to have a question mark here? i.e. `"Did you award a contract?`

app/main/views/g_cloud.py
```diff
@@ -595,10 +595,57 @@ def did_you_award_contract(framework_framework, project_id):
...
+        project_id=project['id'],
+        user_id=current_user.id,
+        fields=required_fields,
+    )
+
+    form = WhichServiceWonTheContractForm(services)
+
+    if request.method == "POST" and form.validate_on_submit():
+        flash('Contract awarded.')
+        return redirect(url_for('.view_project', framework_framework=framework_framework, project_id=project['id']))
```
lfdebrux: Why does this redirect to `.view_project`? Was the flash message part of the prototype?

app/templates/direct-award/which-service-won-contract.html
DilwoarH: You can still access the page without completing the first bit

app/main/views/g_cloud.py
DilwoarH: no but currently there's no route for me to hit, once your stuff is, it will go there.

app/main/forms/direct_award_forms.py
```diff
@@ -39,3 +38,19 @@ class DidYouAwardAContractForm(FlaskForm):
             (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
+
+
+class WhichServiceWonTheContractForm(FlaskForm):
+    which_service_won_the_contract = RadioField(
```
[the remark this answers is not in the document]
DilwoarH: Can't use DMRadioFields as it doesn't support description property

DilwoarH:
@samuelhwilliams @galund @lfdebrux - updated

PR #761, opened 2018-05-17
WIP Restore "FE-1 Did you award a contract for 'Buying some hosting'? page details""

samuelhwilliams:
Reverts alphagov/digitalmarketplace-buyer-frontend#760, which restores the original changes in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/755

---

We temporarily reverted the changes introduced by https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/755 in order to release the new Privacy Policy to production without adding routes that could potentially 500 (albeit only through URL hacking).

This WIP PR reintroduces the original changes and will need to add a FeatureFlag to disable the new routes on (at least) production before getting merged back in.


lfdebrux:
There was a comment from @benvand that I'd like to implement fixes for, and also a conversation with @risicle made me think about whether we should 404 if the search is not yet locked, so this PR might have some extra changes made to it.

lfdebrux:
I've pushed a change that addresses https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/755#discussion_r188861746.

I'm not quite happy with this fix though, I feel that actually the `DMRadioField` class belongs in `dmutils`. Thoughts?

risicle:
It's acceptable to initially keep something in the frontend repo with the idea that a later variant of it gets upstreamed to utils if you'd rather not get distracted by the faff of bumping versions and dependencies right this second...

risicle:
There's a larger area we all should address at some point which is "bridging the gap" between the frontend toolkit and something sane (and, I would say, "modern") like wtforms.

lfdebrux:
Yes, this is something I've been thinking about for firebreak

Laurence de Bruxelles
Developer for Digital Marketplace
07825377786
Government Digital Service

galund:
I'm happy for this to land without the feature flag. As soon as we introduce anything a (non hacky-type) user would see, we will introduce a feature flag (or be ready to go live, whichever comes first).

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,59 @@ def end_search(framework_framework, project_id):
...
+    '/<string:framework_framework>/projects/<int:project_id>/did-you-award-contract',
+    methods=['GET', 'POST']
+)
+def did_you_award_contract(framework_framework, project_id):
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+
```
DilwoarH: This needs a check to not show this page before the search has been locked.

```
project = data_api_client.get_direct_award_project(project_id=project_id)['project']
    if not is_direct_award_project_accessible(project, current_user.id):
        abort(404)

    if not project['lockedAt']:
        abort(400)
```

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,59 @@ def end_search(framework_framework, project_id):
...
+    form = DidYouAwardAContractForm()
+
+    if form.validate_on_submit():
```
DilwoarH: This should have an additional check of the request method

```
    if request.method == "POST" and form.validate_on_submit():
```

app/main/views/g_cloud.py
DilwoarH: It may also need a test to go with this

DilwoarH: Please have a look at https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/763/files#diff-74c43b483b6b57229728b027e18f534eR502 for reference

app/main/views/g_cloud.py
lfdebrux: The `validate_on_submit` method is from Flask-WTF, it includes the check for a `POST` already https://flask-wtf.readthedocs.io/en/stable/api.html#flask_wtf.FlaskForm.validate_on_submit

app/main/views/g_cloud.py
galund: It's possible to over-engineer this: if there's no UI for something, and @risicle is also implementing a test on the back-end POST to check that, then having a 400 isn't massively valuable.

The argument goes something like this:

 - we need to check this in the back end anyway;
 - we should assume that users _aren't_ trying to hack our URL's around, and that if this somehow happens then it's _our_ mistake in wiring up the front-end, not the user's mistake;
 - if we implement the check in the back end, we will therefore get 500's from the front-end if the back-end generates an unexpected 4xx
 - we will then get to find out if we did something wrong in the front-end, rather than just keep telling the user that they did something wrong but not really _what_ (when _we should assume_ it's actually _our_ problem).

lfdebrux: So I just did some tests on local development, and there seems to be a discrepancy.

If I am logged in as a buyer, and try to access `/buyers/direct-award/g-cloud/projects/314159/did-you-award-contract`, then I get a 404, because 314159 is not a valid project id.

When I run a test that accesses `/buyers/direct-award/g-cloud/projects/314159/did-you-award-contract`, I get a 200, even though 314159 is not a valid project id.

I think the problem here is with the patch on `data_api_client.get_direct_award_project`; it returns even if the project id is invalid, whereas the real method raises a 404 if the project id does not exist.

Obviously the patch can be changed to 404 for an individual test, but this might lead to believe an additional check is needed in the implementation, rather than realising that an extra step is needed in the test.

tl;dr I think we need to change the test so that `data_api_client.get_direct_award_project` so that it checks the `project_id` argument

DilwoarH: @galund  - The reason why I would argue otherwise regarding blocking the user from accessing the screen without being on that stage is because of the following:
- For consistency - this check is done on the other functions (https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/761/files#diff-c97e3489337a6c42520133356a12e560R604)
- User's should not be able to view a page which they cannot do anything with. I found that user behaviour has dramatically changed over the years and users will try to take a given url and change the ID to try and bypass stages.
- If for any reason the backend checks ever do fail, we have an additional check.

I'm happy to approve if you still think it's fine.

galund: Sure - consistency argument probably wins out, would be interested to hear what @risicle thinks tho.

There's definitely something amiss with an architecture that requires every piece of business logic to be checked in two or even three places tho :-)

lfdebrux: I've added tests and the code @DilwoarH suggested that address this comment

DilwoarH: Hey @lfdebrux - it seems like this check is missing:

`if not is_direct_award_project_accessible(project, current_user.id):
        abort(404)`

(not sure why we do this, it should be done on the project call level)

lfdebrux: I think this goes back to what @galund was saying.

I'll add the logic for consistency, but it does seem a bit odd; especially because the only way I can think of testing this is to deliberately have the api call return bad data.

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,65 @@ def end_search(framework_framework, project_id):
...
+    if not is_direct_award_project_accessible(project, current_user.id):
+        abort(404)
+
+    if project['lockedAt']:
+        abort(400)
+
+    form = DidYouAwardAContractForm()
+
+    if form.validate_on_submit():
+        if form.did_you_award_a_contract.data == 'still-assessing':
```
samuelhwilliams: I might consider a patch like the one below to avoid the magic strings, but I suspect we probably do it as you have in other places.

```diff
diff --git a/app/main/forms/direct_award_forms.py b/app/main/forms/direct_award_forms.py
index 454f7284..b7eb1885 100644
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -25,12 +25,16 @@ class DMRadioField(RadioField):


 class DidYouAwardAContractForm(FlaskForm):
+    YES = 'yes'
+    NO = 'no'
+    STILL_ASSESSING = 'still-assessing'
+
     did_you_award_a_contract = DMRadioField(
         "Did you award a contract?",
         validators=[InputRequired(message="You need to answer this question.")],
         choices=[
-            ('yes', 'Yes'),
-            ('no', 'No'),
-            ('still-assessing', 'We are still assessing services')
+            (YES, 'Yes'),
+            (NO, 'No'),
+            (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
diff --git a/app/main/views/g_cloud.py b/app/main/views/g_cloud.py
index b499fe89..1b35ffa9 100644
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -564,15 +564,15 @@ def did_you_award_contract(framework_framework, project_id):
     form = DidYouAwardAContractForm()

     if form.validate_on_submit():
-        if form.did_you_award_a_contract.data == 'still-assessing':
+        if form.did_you_award_a_contract.data == form.STILL_ASSESSING:
             return redirect(url_for('.view_project',
                                     framework_framework=framework_framework,
                                     project_id=project_id))
-        elif form.did_you_award_a_contract.data == 'yes':
+        elif form.did_you_award_a_contract.data == form.YES:
             return redirect(url_for('.which_service_won_contract',
                                     framework_framework=framework_framework,
                                     project_id=project_id))
-        elif form.did_you_award_a_contract.data == 'no':
+        elif form.did_you_award_a_contract.data == form.NO:
             return redirect(url_for('.why_didnt_you_award_contract',
                                     framework_framework=framework_framework,
                                     project_id=project_id))
```

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,65 @@ def end_search(framework_framework, project_id):
...
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+
+    if not is_direct_award_project_accessible(project, current_user.id):
+        abort(404)
+
+    if project['lockedAt']:
+        abort(400)
```
samuelhwilliams: As discussed

app/main/views/g_cloud.py
```diff
@@ -544,6 +544,65 @@ def end_search(framework_framework, project_id):
...
+        framework=framework
+    ), 200 if not form.errors else 400
+
+
+@direct_award.route(
+    '/<string:framework_framework>/projects/<int:project_id>/which-service-won-contract',
+    methods=['GET']
+)
+def which_service_won_contract(framework_framework, project_id):
+    pass
```
samuelhwilliams: Can we maybe `abort(404)` rather than `pass` here?

lfdebrux:
@samuelhwilliams all good suggestions, adding now

app/main/forms/direct_award_forms.py
```diff
@@ -14,3 +15,26 @@ class CreateProjectForm(FlaskForm):
             Optional()
         ]
     )
+
+
+class DMRadioField(RadioField):
```
samuelhwilliams: It might be nice to add this to utils and stick a TODO in here to rip it out later when it's available through utils.

Not the end of the world and definitely don't hold off merging this to get it into utils and pull it in immediately.

app/templates/direct-award/did-you-award-contract.html
```diff
@@ -0,0 +1,70 @@
...
+{% endblock %}
+
+{% block main_content %}
+
+<div class="did-you-award-contract-page">
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <header class="page-heading">
+        <h1>
+          Did you award a contract for '{{project.name}}'?
```
samuelhwilliams: One of our old designers has got us all in the habit of only using u+2019 single right quotes (`&rsquo;` or `’`) everywhere in content. We're fairly consistent about this and it was a big thing a while ago, so we try to keep this alive for the time being.

tests/main/views/test_direct_award.py
```diff
@@ -467,6 +468,98 @@ def test_end_search_redirects_to_project_page(self):
...
+
+    def test_award_contract_page_renders(self):
+        self.login_as_buyer()
+
+        res = self.client.get('/buyers/direct-award/g-cloud/projects/1/did-you-award-contract')
+        assert res.status_code == 200
+
+        doc = html.fromstring(res.get_data(as_text=True))
+        assert len(doc.xpath(
+            '//h1[contains(normalize-space(), "Did you award a contract for \'My procurement project\'?")]')) == 1
```
samuelhwilliams: You can also use variables in the xpath strings if you want to avoid escaping strings. I think you can do something like the below. Just FYI, no need to change this.

```
doc.xpath('//h1[contains(normalize-space(), "$text")]',
          text="Did you award a contract for 'My procurement project'?")
```

tests/main/views/test_direct_award.py
```diff
@@ -467,6 +468,98 @@ def test_end_search_redirects_to_project_page(self):
...
+        assert res.location.endswith('/buyers/direct-award/g-cloud/projects/1/which-service-won-contract')
+
+    def test_award_contract_no_redirects_to_why_didnt_you_award_contract(self):
+        self.login_as_buyer()
+
+        res = self.client.post('/buyers/direct-award/g-cloud/projects/1/did-you-award-contract',
+                               data={'did_you_award_a_contract': 'no'})
+
+        assert res.status_code == 302
+        assert res.location.endswith('/buyers/direct-award/g-cloud/projects/1/why-didnt-you-award-contract')
```
samuelhwilliams: The above three tests (test_award_contract_we_are_still_assessing_redirects_to_project_page, test_award_contract_yes_redirects_to_which_service_won_contract, and test_award_contract_no_redirects_to_why_didnt_you_award_contract) feel like good candidates for test parametrization, which can help reduce repetition. In cases where there are just a few different variations, I think it's particularly good for aiding readability - but it is possible to create hard to parse parametrization so be careful not to overdo it. :)

You basically just add a decorater for `pytest.mark.parametrize`, then name the parameters and provide sample values. :)

Example:
```


    @pytest.mark.parametrize('form_choice, should_redirect_to_path',
                             (
                                 ('still-assessing', '/buyers/direct-award/g-cloud/projects/1'),
                                 ('yes', '/buyers/direct-award/g-cloud/projects/1/which-service-won-contract'),
                                 ('no', '/buyers/direct-award/g-cloud/projects/1/why-didnt-you-award-contract'),
                             ))
    def test_award_contract_redirects_based_on_answer(self, form_choice, should_redirect_to_path):
        self.login_as_buyer()

        res = self.client.post('/buyers/direct-award/g-cloud/projects/1/did-you-award-contract',
                               data={'did_you_award_a_contract': form_choice})

        assert res.status_code == 302
        assert res.location.endswith(should_redirect_to_path)
```

lfdebrux:
I've added the validation masthead

tests/main/views/test_direct_award.py
lfdebrux: This makes sense, I will make this change

app/templates/direct-award/did-you-award-contract.html
samuelhwilliams: We actually just use the right quotes, but I'm sure it's fine this time. 😛  🤷‍♂️

samuelhwilliams: Yeah Laurence scoured the code and corrected me.

I have _no idea_ why I had that obviously illogical decision in my head.

Ignore it!

PR #762, opened 2018-05-18
Reduce limit on frozen saved searches to 30 (from 100)

DilwoarH:
As a buyer I want to limit the number of results I can "save" so that I don't need to compare so many results.

AC:
Reduce the saved search limit from 100 to 30.

Note: Only affect new saved searches from now on.

Ticket: https://trello.com/c/kZm5L3ho/77-reduce-limit-on-frozen-saved-searches-to-30-from-100

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -445,7 +445,7 @@ class TestDirectAwardEndSearch(TestDirectAwardBase):
             "services": [],
             "meta": {
                 "query": {},
-                "total": 1000,
+                "total": 31,
                 "took": 3
             },
             "links": {}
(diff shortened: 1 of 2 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -445,7 +445,7 @@ def test_end_search_page_renders_error_when_results_more_than_limit(self):
             "services": [],
             "meta": {
                 "query": {},
-                "total": 1000,
+                "total": 31,
```
katstevens: How about we change this to use the setting instead? 

```
with self.app.app_context():
    self._search_api_client_helpers._get.return_value = {
        "services": [],
        "meta": {
            "query": {},
            "total": current_app.config['END_SEARCH_LIMIT'] + 1,
            "took": 3
        },
        "links": {}
    }
```

katstevens: Ah wait - it's not in the config is it 😹 Even easier then - just import from `app.views.g_cloud.END_SEARCH_LIMIT`.

DilwoarH: good idea 👍

PR #763, opened 2018-05-18
FE-4. Non award front end flow

DilwoarH:
As a buyer I want to indicate why I did not award a contract for this saved search.

Ticket: https://trello.com/c/M64gWnjq/75-fe-4-non-award-front-end-flow

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -58,3 +58,29 @@ class WhichServiceWonTheContractForm(FlaskForm):
             "value": service["id"],
             "description": service["supplier"]["name"]
         } for service in services['services']]
+
+
+class WhyDidYouNotAwardForm(FlaskForm):
+    why_did_you_not_award_the_contract = RadioField(
+        "Why didn't you award the contract?",
+        validators=[InputRequired(message="Please select a reason for why you didn't award the contract")]
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(WhyDidYouNotAwardForm, self).__init__(*args, **kwargs)
+
+        options = [
+            {
+                "label": "The work has been cancelled",
+                "value": "work_cancelled",
+                "description": "For example, because you no longer have the budget",
+            },
+            {
+                "label": "There were no suitable services",
+                "value": "no_suitable_services",
+                "description": "The services in your search results did not meet your requirements",
+            },
+        ]
+
+        self.why_did_you_not_award_the_contract.choices = [(option['value'], option['label']) for option in options]
+        self.why_did_you_not_award_the_contract.options = options
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -62,8 +62,8 @@ class WhichServiceWonTheContractForm(FlaskForm):
 
 class WhyDidYouNotAwardForm(FlaskForm):
     why_did_you_not_award_the_contract = RadioField(
-        "Why didn't you award the contract?",
-        validators=[InputRequired(message="Please select a reason for why you didn't award the contract")]
+        "Why didn't you award a contract?",
+        validators=[InputRequired(message="Please select a reason why you didn't award a contract")]
     )
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


galund:
Please also add as reviewer from outside the team if possible :-)

app/main/forms/direct_award_forms.py
```diff
@@ -14,3 +15,14 @@ class CreateProjectForm(FlaskForm):
             Optional()
         ]
     )
+
+
+class WhyDidntYouAwardForm(FlaskForm):
+    why_didnt_you_award_contract = RadioField(
```
galund: So if we must have URL's that are missing punctuation, then that's a Content thing, and I can live with it. But please can we not have functions or classes named that way!

app/main/forms/direct_award_forms.py
```diff
@@ -39,3 +38,13 @@ class DidYouAwardAContractForm(FlaskForm):
             (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
+
+class WhyDidYouNotAwardForm(FlaskForm):
+    why_did_you_not_award_the_contract = RadioField(
```
galund: If we just called this variable `reason` then in quite a few places I think the code would be more readable?

app/main/views/g_cloud.py
```diff
@@ -605,8 +605,38 @@ def which_service_won_contract(framework_framework, project_id):
...
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+
+    form = WhyDidYouNotAwardForm()
+
+    if request.method == "POST" and form.validate_on_submit():
+        flash('You’ve updated ‘' + project.name + '’')
+        return redirect(url_for('.view_project', framework_framework=framework_framework, project_id=project['id']))
+
+    form.why_did_you_not_award_the_contract.options = [
+        {
```
galund: I'm a bit confused here why we need to update `options`, presumably overriding the `choices` already set? in any even, it feels like these should be in the form constructor, shouldn't they?

DilwoarH: you're right i'll put it in as a form constructor. The reason why I needed it because the template needs the description value passed in

lfdebrux: @DilwoarH have you seen `DMRadioField`? It's a class added it to `direct_award_forms.py`, it creates the `options` in the correct format for you

DilwoarH: I put it inside the __init__ function now - can't use DMRadioField as it does not allow descriptions to be passed in and it uses the choices which only has label and value in it's array

lfdebrux: Ah, understood. I should move `DMRadioField` to `dmutils` anyway so I'll look at adding that functionality.

DilwoarH:
@samuelhwilliams @galund @lfdebrux - updated

app/main/forms/direct_award_forms.py
```diff
@@ -39,3 +39,29 @@ class DidYouAwardAContractForm(FlaskForm):
             (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
+
+
+class WhyDidYouNotAwardForm(FlaskForm):
+    why_did_you_not_award_the_contract = RadioField(
+        "Why didn't you award the contract?",
+        validators=[InputRequired(message="Please select a reason for why you didn't award the contract")]
```
galund: Can we double check with @karlchillmaid but "for why" doesn't seem like good grammar to me - "reason that you didn't" perhaps?

app/main/views/g_cloud.py
```diff
@@ -603,10 +603,39 @@ def which_service_won_contract(framework_framework, project_id):
...
+    form = WhyDidYouNotAwardForm()
+
+    if request.method == "POST" and form.validate_on_submit():
+        flash('You’ve updated ‘' + project['name'] + '’')
+        return redirect(url_for('.view_project', framework_framework=framework_framework, project_id=project['id']))
+
+    errors = [{
+        'input_name': input_name,
+        'question': getattr(form, input_name).label,
+    } for input_name in form.errors.keys()]
```
galund: I think @lfdebrux is planning to move this common thing into a function, so would be good if you can figure out how to do use the same code :-)

samuelhwilliams: It's now in dmutils 37.0.0 as `forms.get_errors_from_wtform`. Albeit pulling that utils update in does recommend some larger changes to the frontend, so it might be a pain pulling that in. I might suggest leaving it as-is here, until I can get around to pulling it in. I'm halfway through the work on the buyer-fe.

PR #764, opened 2018-05-18
Check project id when calling `get_direct_award_project` in tests

lfdebrux:
The implementation of `data_api_client.get_direct_award_project()`
raises a 404 if the project id is not valid.

This commit changes how the `data_api_client` Mock is patched so that
this behaviour is recreated in tests.

I noticed this difference in behaviour in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/761#discussion_r189235250.

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -10,6 +10,7 @@ import pytest
 from werkzeug.exceptions import BadRequest, NotFound
 
 from dmcontent.content_loader import ContentLoader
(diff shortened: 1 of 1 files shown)


lfdebrux:
I have created https://github.com/alphagov/digitalmarketplace-test-utils/pull/8 to turn this pattern into a generic function.

lfdebrux:
Closing this, we should just keep the current behaviour where if you expect a 404 in a test then you should patch the mock in that test

PR #765, opened 2018-05-21
Updates to flash message for too many results message

DilwoarH:
Ticket: https://trello.com/c/kZm5L3ho/77-reduce-limit-for-ending-a-search-to-30-from-100


DilwoarH:
@karlchillmaid:

This is the code `if int(search_count) > END_SEARCH_LIMIT` so effectively would be 30 or more

katstevens:
The unit tests will need fixing too, once the content has been agreed.

karlchillmaid:
`Refine your search terms until you have no more than 30 results.`

DilwoarH:
@katstevens - tests are now fixed

PR #770, opened 2018-05-24
FE-3. Tell us about your contract page

lfdebrux:
Ticket: [Trello](https://trello.com/c/kPG8B9ZJ/74-fe-3-tell-us-about-your-contract-page)

This page is for users to tell Digital Marketplace about the length and
value of the contract they are awarding through G-Cloud.

It uses WTForms for validation, and includes some new classes to
validate input from the `frontend-toolkit` date component.

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -1,7 +1,14 @@
 from flask_wtf import FlaskForm
-from wtforms.fields import RadioField
-from wtforms.validators import Length, Optional, InputRequired
+
+from wtforms.fields import DecimalField, RadioField
+from wtforms.validators import DataRequired, Length, NumberRange, Optional, InputRequired
+
+from .fields import DMRadioField, DateField
+from .validators import GreaterThan, Y2K
+
 from dmutils.forms import StripWhitespaceStringField
+
+from decimal import Decimal
 
 
 class CreateProjectForm(FlaskForm):
@@ -11,17 +18,8 @@ class CreateProjectForm(FlaskForm):
             Length(min=1,
                    max=100,
                    message="Names must be between 1 and 100 characters"),
-            Optional()
-        ]
-    )
-
-
-# TODO: move this into dmutils.forms
-class DMRadioField(RadioField):
-    @property
-    def options(self):
-        """The RadioField choices in a format suitable for the frontend toolkit"""
-        return [{"label": label, "value": value} for value, label in self.choices]
+            Optional(),
+        ])
 
 
 class DidYouAwardAContractForm(FlaskForm):
@@ -35,9 +33,8 @@ class DidYouAwardAContractForm(FlaskForm):
         choices=[
             (YES, 'Yes'),
             (NO, 'No'),
-            (STILL_ASSESSING, 'We are still assessing services')
-        ]
-    )
+            (STILL_ASSESSING, 'We are still assessing services'),
+        ])
 
 
 class WhichServiceWonTheContractForm(FlaskForm):
@@ -58,6 +55,41 @@ class WhichServiceWonTheContractForm(FlaskForm):
             "value": service["id"],
             "description": service["supplier"]["name"]
         } for service in services['services']]
+
+
+class TellUsAboutContractForm(FlaskForm):
+    INPUT_REQUIRED_MESSAGE = "You need to answer this question."
+    INVALID_DATE_MESSAGE = "Your answer must be a valid date."
+    INVALID_VALUE_MESSAGE = "Enter your value in pounds and pence using numbers and decimals only" \
+                            ", for example 9900.05 for 9900 pounds and 5 pence."
+
+    start_date = DateField(
+        "Start date",
+        validators=[
+            InputRequired(INPUT_REQUIRED_MESSAGE),
+            DataRequired(INVALID_DATE_MESSAGE),
+            Y2K(INVALID_DATE_MESSAGE),
+        ])
+    end_date = DateField(
+        "End date",
+        validators=[
+            InputRequired(INPUT_REQUIRED_MESSAGE),
+            DataRequired(INVALID_DATE_MESSAGE),
+            Y2K(INVALID_DATE_MESSAGE),
+            GreaterThan("start_date", "Your end date must be later than the start date."),
+        ])
+    value_in_pounds = DecimalField(
+        "Value",
+        validators=[
+            InputRequired(INPUT_REQUIRED_MESSAGE),
+            DataRequired(INVALID_VALUE_MESSAGE),
+            NumberRange(min=Decimal('0.01'), message=INVALID_VALUE_MESSAGE),
+        ])
+    buying_organisation = StripWhitespaceStringField(
+        "Organisation buying the service",
+        validators=[
+            InputRequired(INPUT_REQUIRED_MESSAGE)
+        ])
 
 
 class WhyDidYouNotAwardForm(FlaskForm):
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -70,6 +70,7 @@ class TellUsAboutContractForm(FlaskForm):
             DataRequired(INVALID_DATE_MESSAGE),
             Y2K(INVALID_DATE_MESSAGE),
         ])
+
     end_date = DateField(
         "End date",
         validators=[
@@ -78,6 +79,7 @@ class TellUsAboutContractForm(FlaskForm):
             Y2K(INVALID_DATE_MESSAGE),
             GreaterThan("start_date", "Your end date must be later than the start date."),
         ])
+
     value_in_pounds = DecimalField(
         "Value",
         validators=[
@@ -85,6 +87,7 @@ class TellUsAboutContractForm(FlaskForm):
             DataRequired(INVALID_VALUE_MESSAGE),
             NumberRange(min=Decimal('0.01'), message=INVALID_VALUE_MESSAGE),
         ])
+
     buying_organisation = StripWhitespaceStringField(
         "Organisation buying the service",
         validators=[
--- a/app/main/forms/fields.py
+++ b/app/main/forms/fields.py
@@ -21,20 +21,33 @@ class DMRadioField(RadioField):
 
 class DateField(Field):
     '''
-    A field enclosure for a date entry fieldset
+    A date field(set) that uses a day, month and year field.
+
+    It behaves like a WTForms.FieldForm, but it can be used
+    with validators like a normal WTForms.Field.
 
     >>> from wtforms import Form
+    >>> from wtforms.validators import DataRequired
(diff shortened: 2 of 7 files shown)


app/main/forms/direct_award_forms.py
```diff
@@ -39,3 +44,138 @@ class DidYouAwardAContractForm(FlaskForm):
             (STILL_ASSESSING, 'We are still assessing services')
         ]
     )
+
+
+class DMFormField(FormField):
+    @property
+    def errors(self):
+        return [e for errors in self.form.errors.values() for e in errors]
```
galund: Unless this needs to return a list, then generator-style (i.e. with regular not square brackets) would probably be better.

If this thing, and the below validators, feel like they aren't specifically tied to the Direct Award flow, then let's a least put them into a separate module, even if we aren't yet ready to move them into dm-utils.

app/main/views/g_cloud.py
```diff
@@ -609,6 +609,38 @@ def why_didnt_you_award_contract(framework_framework, project_id):
...
+    all_frameworks = data_api_client.find_frameworks().get('frameworks')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_framework)
+
+    # Get the requested Direct Award Project.
+    project = data_api_client.get_direct_award_project(project_id=project_id)['project']
+
+    form = TellUsAboutContractForm()
+
+    if form.validate_on_submit():
+        flash(Markup(f"""You've updated '{project['name']}'"""), 'success')
```
galund: I'd just go with double quotes unless you actually need the triple quotes, but I guess this is personal preference

app/main/views/g_cloud.py
```diff
@@ -609,6 +609,38 @@ def why_didnt_you_award_contract(framework_framework, project_id):
...
+        return redirect(url_for('.view_project', framework_framework=framework_framework, project_id=project_id))
+
+    errors = [
+        {
+            "input_name": input_name,
+            "question": getattr(form, input_name).label
+        } for input_name in form.errors.keys()]
```
galund: Building the list of errors in this way might be worth abstracting out into a separate function?

tests/main/views/test_direct_award.py
```diff
@@ -548,6 +548,101 @@ def test_award_contract_raises_400_if_project_is_not_locked(self):
...
+        }
+
+    def test_tell_us_about_contract_exists(self, client):
+        assert client.get(self.url).status_code == 200
+
+    def test_tell_us_about_contract_renders(self, x):
+        assert x('boolean(//h1[contains(normalize-space(), "Tell us about your contract")])')
+
+    def test_tell_us_about_contract_form_fields(self, x):
+        assert x('count(//input[@type="text"][substring-after(@name, "start_date")])') == 3
```
galund: If `x` is a regular instance function then surely it needs to be called via `self.x`, so I guess there's something that I'm not understanding here if this does actually work.

Either way, I think `x` could be better named, maybe fetch_and_inspect or something, I dunno. I get why you wanted a shorter name, tho, so eh.

app/main/views/g_cloud.py
lfdebrux: I went with triple quotes because all examples of flash messages I'd found had them, which is also why it is a Markup object. I agree it is unnecessary in this case but I wanted to be consistent.

app/main/views/g_cloud.py
lfdebrux: Definitely

tests/main/views/test_direct_award.py
lfdebrux: Instance methods are bound to `__self__` when accessed from an object instance. You can pass around a bound instance method much like any other callable. It's when accessing an instance method through the class that you get an unbound method.

`x` could definitely be better named. How about `xpath`? It's relatively short but also the name of the method, so relatively descriptive.

app/main/forms/direct_award_forms.py
lfdebrux: It needs to be a list as the property it is overloading returns a list that is usually accessed using `__getitem__`.

I will move them into a separate module.

tests/main/views/test_direct_award.py
galund: Hmm, but we're actually calling it here - we assert the result of the function, so we must be, no?

xpath is ok but kinda duplicative of the parameter name. It doesn't describe what we're doing, more how we're doing it.

app/main/views/g_cloud.py
lfdebrux: Looks like @samuelhwilliams beat me to this :) https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/763#discussion_r191420911

app/main/views/g_cloud.py
lfdebrux: Looks like @samuelhwilliams beat me to this 👍 https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/763#discussion_r191420911

app/main/views/g_cloud.py
galund: I think it was probably someone copy-pasting a multiline one, and then it became an unstoppable meme, rather than an intentional consistency, but whatevs :)

app/main/views/g_cloud.py
galund: If we're in a position to bring in that version of dm-utils then go for it, but if not then this can wait as long as there's a plan to do that.

app/main/forms/direct_award_forms.py
```diff
@@ -39,3 +37,40 @@ class DidYouAwardAContractForm(FlaskForm):
...
+    value_in_pounds = DecimalField(
+        "Value",
+        validators=[
+            InputRequired("You need to answer this question."),
+            DataRequired(
+                message="Enter your value in pounds and pence using numbers and decimals only"
+                        ", for example 9900.05 for 9900 pounds and 5 pence."),
+            NumberRange(
+                min=Decimal('0.01'),
+                message="Enter your value in pounds and pence using numbers and decimals only"
```
galund: If we need the same message for two validators then let's use a constant?

app/main/forms/fields.py
```diff
@@ -0,0 +1,78 @@
...
+    def process(self, formdata, data=unset_value):
+        self.process_errors = []
+
+        self.form_field.process(formdata, data)
+
+        raw_data = {field.name: field.raw_data[0] for field in self.form_field
+                    if field.raw_data}
+        if not any(raw_data.values()):
+            raw_data = {}
+        self.raw_data = [raw_data]
```
galund: I may have misunderstood, but I think you need to make sure that `raw_data` only gets set if `formdata is not None`.

Also you might be missing a check on whether `data is unset_value` - or maybe not, perhaps that case just falls out. But I was really just looking to see how similar this implementation is to the base class impl, and it seems slightly more different than I expected :)

I think your `if not any(raw_data.values())` test is unnecessary too - I can't see how a dict comprehension can ever not return a dict, even if it is empty?

tests/main/views/test_direct_award.py
```diff
@@ -548,6 +548,103 @@ def test_award_contract_raises_400_if_project_is_not_locked(self):
...
+        assert xpath(
+            'boolean('
+            '//a[@href="/buyers/direct-award/g-cloud/projects/1/which-service-won-contract"]'
+            '[contains(normalize-space(), "Previous page")])')
+
+    def test_tell_us_about_contract_form_action_url_is_tell_us_about_contract_url(self, xpath):
+        assert xpath('//form[@class="tell-us-about-contract-form"]/@action')[0] == self.url
+
+    def test_tell_us_about_contract_raises_400_if_project_is_not_accessible(self, client):
+        self.data_api_client.get_direct_award_project.side_effect = HTTPError(mock.Mock(status_code=400))
```
galund: Feels like there might be something wrong here - the other views use `is_direct_award_project_accessible`, and throw a 404 if it isn't. This view doesn't do that... but somehow the test passes. Hmm!

lfdebrux: This needs fixing

app/main/forms/fields.py
lfdebrux: Would it be worth adding some comments to this logic? It's been through several iterations of TDD to get it to a state where it works in most of the common scenarios I could think of (with different combinations of validators), which is why it has a couple of weird quirks.

galund: Sure.

BTW I see now that your `any` test is converting a dict consisting only of empty strings into an empty dict, so I guess I can definitely see where I was wrong there.

app/main/views/g_cloud.py
katstevens: Yes please - let's pull in the error formatter:

```
from dmutils.forms import get_errors_from_wtform

errors = get_errors_from_wtform(form)
```

lfdebrux: @samuelhwilliams is working on getting this done, apparently updating to `dmutils` with the helpers is not a straightforward thing 😢

app/main/forms/direct_award_forms.py
```diff
@@ -60,6 +57,41 @@ def __init__(self, services, *args, **kwargs):
...
+            GreaterThan("start_date", "Your end date must be later than the start date."),
+        ])
+    value_in_pounds = DecimalField(
+        "Value",
+        validators=[
+            InputRequired(INPUT_REQUIRED_MESSAGE),
+            DataRequired(INVALID_VALUE_MESSAGE),
+            NumberRange(min=Decimal('0.01'), message=INVALID_VALUE_MESSAGE),
+        ])
+    buying_organisation = StripWhitespaceStringField(
```
DilwoarH: maybe a single line space between the variables to make it readable

tests/main/views/test_direct_award.py
```diff
@@ -593,6 +593,114 @@ def test_which_service_did_you_award_page_should_not_render_if_not_locked_render
...
+    def test_tell_us_about_contract_raises_404_if_project_is_not_accessible(self, client):
+        self.data_api_client.get_direct_award_project.return_value['project']['id'] = 314159
+        self.data_api_client.get_direct_award_project.return_value['project']['users'][0]['id'] = 321
+
+        assert client.get('/buyers/direct-award/g-cloud/projects/31415/tell-us-about-contract').status_code == 404
+
+    def test_tell_us_about_contract_successful_post_redirects_to_project_overview(self, client, data):
+        res = client.post(self.url, data=data)
+        assert res.status_code == 302
+        assert res.location.endswith('/buyers/direct-award/g-cloud/projects/1')
```
katstevens: I'd include an assert that `self.data_api_client.update_contract_outcome` is called. If that's the right method. Er... we are saving these details to the API, right?

lfdebrux: Not yet

lfdebrux:
Closing in favour of a new PR

PR #771, opened 2018-05-25
Restrict opportunity pages

samuelhwilliams:
## Summary
We have some opportunity routes that accept a framework family in the
route. Where this is the case, we should check that the framework
actually supports opportunity-style views before rendering a response.

 ## Ticket
https://trello.com/c/WruHrcBf/456

### DIFF
--- a/app/main/helpers/framework_helpers.py
+++ b/app/main/helpers/framework_helpers.py
@@ -1,3 +1,5 @@
+from flask import abort
+
 from ... import content_loader
 
 
@@ -5,15 +7,24 @@ def get_frameworks_by_slug(data_api_client):
     return {framework['slug']: framework for framework in data_api_client.find_frameworks().get('frameworks')}
 
 
-def get_latest_live_framework(all_frameworks, framework_type):
+def get_latest_live_framework(all_frameworks, framework_family):
     try:
         latest = max(
-            (f for f in all_frameworks if f['status'] == 'live' and f['framework'] == framework_type),
+            (f for f in all_frameworks if f['status'] == 'live' and f['framework'] == framework_family),
             key=lambda f: f['id'],
         )
         return latest
     except ValueError:  # max of empty iterable
         return None
+
+
+def get_latest_live_framework_or_404(all_frameworks, framework_family):
+    latest_live_framework = get_latest_live_framework(all_frameworks, framework_family)
+
+    if not latest_live_framework:
+        abort(404, f"No latest framework found for `{framework_family}` family")
+
+    return latest_live_framework
 
 
 def get_lots_by_slug(framework_data):
@@ -29,3 +40,8 @@ def get_framework_description(data_api_client, framework_family):
     content_loader.load_messages(framework['slug'], ['descriptions'])
 
     return content_loader.get_message(framework['slug'], 'descriptions', 'framework')
+
+
+def abort_if_not_further_competition_framework(framework):
+    if not framework['hasFurtherCompetition']:
+        abort(404, f"Framework `{framework}` does not support further competition.")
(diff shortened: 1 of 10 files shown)


galund:
The more code we put in like this, the more we'll have to refactor when we eventually get this modelled correctly at lot level. So yeah I'd rather chat about this before we go down this route any further, preferably with Louise as well.

samuelhwilliams:
Sure, let's chat - although this PR was the reason for the rest of the changes so it's a shame we're blocking this now rather than sooner.

But better now than never if we think it's worth it.

app/main/views/marketplace.py
```diff
@@ -119,6 +124,13 @@ def external_404():
 
 @main.route('/<framework_framework>/opportunities/<brief_id>')
 def get_brief_by_id(framework_framework, brief_id):
+    frameworks = data_api_client.find_frameworks()['frameworks']
+    frameworks = [framework for framework in frameworks if framework['framework'] == framework_framework]
```
Wynndow: https://imgflip.com/i/2azn1o

galund: So I can insist we have a variable name change here: https://trello.com/c/iqFcVKpd/14-replace-framework-framework-with-framework-family

samuelhwilliams: Yeah I did start doing this, then it cascaded a bit with loads of broken views. I think it's easier to just open a separate PR making these changes, so I'll do that now.

samuelhwilliams: Have put up PRs to work on that ticket.

samuelhwilliams:
Spoke to @louzoid-gds and @galund  - we'll put this in, as the work's already done. Will add a comment to the API to formalise that we think this isn't an accurate representation of how direct_award/further_competition relate to procurement (specifically it should technically be at the lot-level not the framework-level).

app/main/helpers/framework_helpers.py
```diff
@@ -16,6 +18,15 @@ def get_latest_live_framework(all_frameworks, framework_type):
         return None
 
 
+def get_latest_live_framework_or_404(all_frameworks, framework_type):
+    latest_live_framework = get_latest_live_framework(all_frameworks, framework_type)
+
+    if not latest_live_framework:
+        abort(404, f"No latest framework found for `{framework_type}` family")
```
galund: I'd have thought that if this ever actually happens, something has gone terribly wrong, so surely dying loudly would be better than 404'ing? If this really can happen then of course `get..or_404` will avoid unnecessary noise in the logs and that's fine.

samuelhwilliams: Mm, it shouldn't happen under normal circumstances. But I think it would be annoying (and weirdly fragile) to generate a 500 if a user changed `...gov.ul/digital-outcomes-and-specialists/opportunities` to `...gov.uk/blah/opportunities` (or whatever).

galund: We should probably rename these to `framework_family` whilst we're here (or in the other PR).

samuelhwilliams: I'm going to do separate PRs for those in the rest of the apps soon.

samuelhwilliams: Yeah, done as part of the rebase.

galund: I _wasn't_ asking you to go around all the apps doing this everywhere, unless you happen to be looking at that code anyway :-D

samuelhwilliams: It's an easy enough change and I'd like to just get it over with. 👍 

Actually changing the API responses will be harder. I need to give it some thought.

PR #773, opened 2018-06-04
Iterate opt in flow for user research participants

DilwoarH:
**As a buyer I want to easily opt-in to user research so that I can help improve the buyer experience for G-Cloud**

The current opt-in journey has been unsuccessful, only 6 buyers have signed up. A higher number of suppliers, however, we suspect that this is due to G10 submission going on at the same time.

Buyer and suppliers cannot opt-in until they log in. Buyers and suppliers that don't log in don't have an opportunity to opt-in for user research.

***

**Changes**

Cookie banner to appear on all pages for logged in and logged out users until they click close or opt in. If clicked close, it reappears after 90 days.

***

**Notes**

Users will have to log in before they can opt-in.
We should use the existing opt-in journey as much as possible. This is non-mission work.

Ticket: https://trello.com/c/48pZNhaG/120-iterate-opt-in-flow-for-user-research-participants

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -488,10 +488,17 @@ def view_project(framework_family, project_id):
     else:
         current_project_stage = 'save_and_refine_search'
 
+    # custom dimension for google analytics
+    custom_dimensions = [{
+        "data_id": 8,
+        "data_value": current_project_stage
+    }]
+
     return render_template('direct-award/view-project.html',
                            framework=framework,
                            project=project,
                            current_project_stage=current_project_stage,
+                           custom_dimensions=custom_dimensions,
                            search=search,
                            buyer_search_page_url=buyer_search_page_url,
                            search_summary_sentence=search_summary_sentence,
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -20,6 +20,7 @@
 //= require ../../../node_modules/govuk_frontend_toolkit/javascripts/govuk/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/shim-links-with-button-role.js
 //= require ../../../node_modules/digitalmarketplace-frontend-toolkit/toolkit/javascripts/show-hide-content.js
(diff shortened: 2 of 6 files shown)


galund:
You might need to check the changes brought in by version 29 of the toolkit :/ (cc @katstevens )

app/main/views/g_cloud.py
```diff
@@ -487,11 +487,18 @@ def view_project(framework_family, project_id):
         current_project_stage = 'search_ended'
     else:
         current_project_stage = 'save_and_refine_search'
+    
+    # custom dimension for google analytics
+    custom_dimensions = [{
```
samuelhwilliams: If we're making this something that we can add on the Python side, it would be good to maybe have a `CustomDimension` class in the fe-toolkit we can use to have some sort of standard format for passing into the templates. Probably just a simple object with an ID and value is sufficient, but would be useful.

galund: Yep, kinda feel like we don't have anywhere obvious to put (or look for) this right now. I suppose it would have to go into dm-utils.

Perhaps this is only worth putting in if we come across another place where we actually need extra custom dimensions, it's maybe overkill for now given the single place we ever use it.

galund: FWIW I *do* think we should consider having a Python library that goes along with the frontend toolkit. I think there are probably quite a few things like the forms stuff that @lfdebrux and you @samuelhwilliams have been doing that might be better separated off from the rest of dm-utils.

But @risicle will hate me for ever for suggesting making another library, so let's not do that just yet!...

risicle: If you really want to set up another `scripts/` directory, another pytest setup, travis, jenkins job... and keep them all up to date with the way we do these things elsewhere, be my guest.

risicle: Can we not just have the base template(s) able to consume a `custom_dimensions` dict of key/values to turn into custom dimensions on rendering? a la flash messages?

galund: Well yes that's the current situation, and I think we'll leave it that way until this becomes more widespread and if people feel they a class would help ensure we get the right structure everywhere.

It's true that Flash messages in Flask don't use a class, but they do have a special function to call whose argument list determines what structure you can give them, so it's kinda equivalent.

PR #776, opened 2018-06-07
 Analytics for new award flow

DilwoarH:
Overall User story
=============

As a product manager I want to be able to get analytics for the new award flow so that I can track the behaviour of the user.

Resources
========
- https://drive.google.com/drive/u/0/folders/1fJM7CTuRzBEJ7xF_7nj0m4sQgjJQV9R5

See https://trello.com/c/2rthNi7A/71-new-states-for-stage-6-on-search-summary-task-list-page

https://trello.com/c/FMuTKjo4/114-analytics-for-new-award-flow

Ticket: https://trello.com/c/FMuTKjo4/114-analytics-for-new-award-flow

bump to 31.1.0

### DIFF
--- a/app/assets/javascripts/analytics/_init.js
+++ b/app/assets/javascripts/analytics/_init.js
@@ -5,6 +5,7 @@
     this.register();
     this.pageViews.init();
     this.events.init();
+    this.buyerSpecificEvents.init();
     this.virtualPageViews.init();
     this.scrollTracking.init();
     this.trackExternalLinks.init();
--- a/app/assets/javascripts/_analytics.js
+++ b/app/assets/javascripts/_analytics.js
@@ -2,7 +2,7 @@
 
 // TODO: move remaining shared analytics code into FE toolkit
 //= include analytics/_pageViews.js
-//= include analytics/_events.js
+//= include analytics/_buyerSpecificEvents.js
 //= include analytics/_virtualPageViews.js
 //= include analytics/_scrollTracking.js
 //= include analytics/_init.js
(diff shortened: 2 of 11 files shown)


app/assets/javascripts/analytics/_events.js
```diff
@@ -89,7 +89,7 @@
 
   GOVUK.GDM.analytics.LinkClick = LinkClick;
 
-  GOVUK.GDM.analytics.events = {
+  GOVUK.GDM.analytics.buyerSpecificEvents = {
```
galund: I think the name here is supposed to correspond to the module/file name, to make understanding/debugging more sane?

DilwoarH: @galund - updated

app/assets/javascripts/analytics/_init.js
```diff
@@ -5,6 +5,7 @@
     this.register();
     this.pageViews.init();
     this.events.init();
```
galund: So does this work / is it initialising the FE toolkit events.js, and if so shouldn't that be happening in `toolkit/javascripts/analytics/_init.js`?

DilwoarH: Hi @galund - yes this is needed as this is an override function. If the init is not done here then the function would not trigger.

galund: Comment here - and perhaps also in the effectively-overridden init function? (in the other repo I guess) - and then we're good. I'll approve now so you don't have to wait for me any more.

DilwoarH: Comment left here

https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/443#issuecomment-398343433

PR #777, opened 2018-06-07
New states for stage 6 on search summary (task list) page

DilwoarH:
As a user I want to see the state of my procurement / saved search.

States:

+ Contract awarded to x
+ No suitable services
+ The work has been canceled
+ Can't start yet (if the search isn't locked yet)
+ Tell us the outcome

https://docs.google.com/document/d/1uX__E-5tPWLWKQpX0w-2Dn3WzjLqeEg3JlBYnTFHhJU/edit#

Ticket: https://trello.com/c/2rthNi7A/71-new-states-for-stage-6-on-search-summary-task-list-page

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -480,6 +480,20 @@ def view_project(framework_family, project_id):
     content_loader.load_messages(framework['slug'], ['urls'])
     framework_urls = content_loader.get_message(framework['slug'], 'urls')
 
+    # Project outcome
+
+    project_outcome_label = None
+
+    if project['outcome'] and project['outcome']['result'] == "cancelled":
+        project_outcome_label = "The work has been cancelled"
+    elif project['outcome'] and project['outcome']['result'] == "none-suitable":
+        project_outcome_label = "No suitable services found"
+    elif project['outcome'] and project['outcome']['result'] == "awarded":
+        archived_service_id = project['outcome']['resultOfDirectAward']['archivedService']['id']
+        archived_service = data_api_client.get_archived_service(archived_service_id=archived_service_id)['services']
+        project_outcome_label = \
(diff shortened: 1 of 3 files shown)


app/templates/direct-award/view-project.html
```diff
@@ -104,7 +104,9 @@
                 data-analytics-action=\"External Link\"
                 >how to award a contract when you buy services</a>."])|safe},
               {"type": "text", "text": ''.join(["The buyer and supplier must both sign a copy of the contract before the service can be used."])|safe},
-              {"type": "box", "style": "inactive", "label": "Can't start yet"} if not project.downloadedAt else {}
+              {"type": "box", "style": "complete", "label": project.outcome} if project.outcome else {},
```
galund: So:

1) for non-award outcomes, we need to translate `project.outcome` into a human-readable string, as it's just a machine-readable identifier right now;

2) for award outcomes, you need to check `project.outcome.resultOfDirectAward.archivedServiceId` and look up the details of the service using the `/archived-services/<int:archived_service_id>` endpoint of the API. In this case we'll have to hard-code a reference to the service name question, and document that we've done that in the frameworks repo.

app/main/views/g_cloud.py
```diff
@@ -480,6 +480,19 @@ def view_project(framework_family, project_id):
     content_loader.load_messages(framework['slug'], ['urls'])
     framework_urls = content_loader.get_message(framework['slug'], 'urls')
 
+    # Project outcome
+
+    project_outcome_label = None
+
+    if project['outcome'] and project['outcome']['result'] == "cancelled":
+        project_outcome_label = "The work has been cancelled"
```
galund: This would be better as a dict at the top of the module?

galund: (Actually scratch that this is fine.)

PR #779, opened 2018-06-08
Tell us about your contract page

lfdebrux:
Ticket:
[Trello](https://trello.com/c/kPG8B9ZJ/74-fe-3-tell-us-about-your-contract-page)

This page is for users to tell Digital Marketplace about the length and
value of the contract they are awarding through G-Cloud.

It is reviving #770, but it does less to make the features easier to review.

The DateField class has been moved to dmutils (merged in https://github.com/alphagov/digitalmarketplace-utils/pull/406) and the modifications to WTForms have been removed. They will be added at a later date when another dmutils PR (https://github.com/alphagov/digitalmarketplace-utils/pull/400) is merged.

This PR also is consistent with the work @samuelhwilliams has done recently with validation mastheads and form errors.

tl;dr please approve! 👍 🙏 😁

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -640,7 +640,7 @@ def which_service_won_contract(framework_family, project_id):
 
     if request.method == "POST" and form.validate_on_submit():
         flash('Contract awarded.')
-        return redirect(url_for('.view_project', framework_family=framework_family, project_id=project['id']))
+        return redirect(url_for('.tell_us_about_contract', framework_family=framework_family, project_id=project['id']))
 
     errors = get_errors_from_wtform(form)
 
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -21,6 +21,7 @@ from ..exceptions import AuthException
 from ..forms.direct_award_forms import (
     CreateProjectForm,
     DidYouAwardAContractForm,
+    TellUsAboutContractForm,
     WhichServiceWonTheContractForm,
     WhyDidYouNotAwardForm
 )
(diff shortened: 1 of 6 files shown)


app/main/views/g_cloud.py
```diff
@@ -653,6 +654,40 @@ def which_service_won_contract(framework_family, project_id):
...
+    if not is_direct_award_project_accessible(project, current_user.id):
+        abort(404)
+
+    if not project['lockedAt']:
+        abort(400)
+
+    form = TellUsAboutContractForm()
+
+    if form.validate_on_submit():
+        flash(Markup(f"""You've updated '{project['name']}'"""), 'success')
```
samuelhwilliams: @lfdebrux We need to fix this before it gets out to production - wrapping this in Markup means we need to escape the project name, or this can happen:
screen shot 2018-06-08 at 19 19 17

samuelhwilliams: Probably all we need to do in this case is not wrap it in `Markup`.

lfdebrux: Just saw this.

Wow. That's a real howler. I am incredibly embarrassed.

That will teach me to copy paste without thinking :/

samuelhwilliams: No need to be embarrassed. Easy thing to miss and it's unlikely to happen again.

Been trying to think how we could have a test make sure this doesn't happen on a broader scale than writing a custom test for each page, but not managed to come up with a good idea yet. Some 'dangerous' API stubs that are easy to inject into tests might be a good start. Maybe a process in the manual, as well, listing common vulnerabilities that should be tested against when implementing a new page. Something to think about.

PR #781, opened 2018-06-11
FE-6 Award front end flow - dashboard to award submission

DilwoarH:
As a buyer, I want to tell the Digital Marketplace / CCS about the G-Cloud contract that I have awarded.

This ticket covers the initial "Tell us the outcome" link on the "Your saved searches" page, and also functional tests covering the entire process through to the award form being submitted.

The other pages should already have been covered in FE-1 to FE-4.

Ticket: https://trello.com/c/ricTDv4f/83-fe-6-award-front-end-flow-dashboard-to-award-submission

### DIFF
--- a/app/templates/direct-award/index.html
+++ b/app/templates/direct-award/index.html
@@ -46,21 +46,29 @@
       {% endcall %}
     {% endcall %}
 
-    {{ summary.heading("Search ended", id="search_ended_table") }}
+    {{ summary.heading("Results exported", id="search_ended_table") }}
     {% call(item) summary.list_table(closed_projects,
       caption="Search ended",
       empty_message="You have no saved searches",
       field_headings=[
         "Name",
-        "Ended"  ],
+        "Exported",
+        "Status"
+      ],
       field_headings_visible=True
     ) %}
       {% call summary.row() %}
-        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_family=framework.framework, project_id=item.id)) }}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_family=framework.framework, project_id=item.id), wide=False) }}
         {{ summary.text(item.lockedAt | datetimeformat ) }}
+        {{ 
+          summary.text("Awarded") if item.outcome and item.outcome.result == "awarded"
+          else summary.text("The work has been cancelled") if item.outcome and item.outcome.result == "cancelled"
+          else summary.text("No suitable services found") if item.outcome and item.outcome.result == "none-suitable"
+          else summary.service_link("Download results", url_for('direct_award.search_results', framework_family=framework.framework, project_id=item.id), wide=False) if not item.downloadedAt
+          else summary.service_link("Tell us the outcome", url_for('direct_award.did_you_award_contract', framework_family=framework.framework, project_id=item.id), wide=False) }}
       {% endcall %}
     {% endcall %}
-
+    
   </div>
 
 </div>
--- a/tests/fixtures/direct_award_project_list_fixture.json
+++ b/tests/fixtures/direct_award_project_list_fixture.json
@@ -1,90 +1,455 @@
 {
-  "links": {
-      "self": "http://127.0.0.1:5000/direct-award/projects?latest-first=1"
-  },
-  "meta": {
-      "total": 10
-  },
-  "projects": [
-      {
-          "active": true,
-          "createdAt": "2017-10-18T11:22:04.840284Z",
-          "downloadedAt": null,
-          "id": 61,
-          "lockedAt": null,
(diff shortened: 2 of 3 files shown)


app/templates/direct-award/index.html
```diff
@@ -52,16 +52,23 @@ <h1>
       empty_message="You have no saved searches",
       field_headings=[
         "Name",
-        "Ended"  ],
+        "Ended",
```
galund: Thought occurs that we might need a new word here from @karlchillmaid, perhaps "Finished searching date" or something, if we aren't using the word "end" in relation to searches anywhere now.

DilwoarH: @karlchillmaid - if you leave your comments here about the table heading, I'll update it

tests/fixtures/direct_award_project_list_fixture.json
```diff
@@ -1,90 +1,455 @@
...
+            "active": true,
+            "createdAt": "2018-05-24T12:08:17.226976Z",
+            "downloadedAt": "2018-05-24T12:08:25.251541Z",
+            "id": 207198828251156,
+            "lockedAt": "2018-05-24T12:08:21.207534Z",
+            "name": "test",
+            "outcome": null
+        }
+    ]
+}
```
samuelhwilliams: This file is insane and needs to die 😨 

We should be using (and where necessary, building out) the api stubs in our utils repo to stop us needing files like this.

lfdebrux: I think we should create a tech debt ticket where we can talk about how we create test fixtures, as there seem to be a couple of different ways we use (and I didn't know about the api stubs until you mentioned it).

app/templates/direct-award/index.html
kubabartwicki: Hey @DilwoarH, see below for the changes we discussed earlier that @karlchillmaid and I are suggesting. 

1) "Search ended" table name changes to "Results exported"
2) "Ended" column name changes to "Exported"
3) "Outcome" column name changes to "Status"
4) new state "Download results" for when search has been exported but not yet downloaded (i.e. user shouldn't be able to tell us the outcome just yet). This should link to the *Download your results* page

screen shot 2018-06-27 at 15 53 01

galund: Re (4), I think this is very much an edge case, especially if we do combine those steps as originally planned.

However, it would make sense for the "Download results" link to be there right up until they've awarded the contract, regardless of whether they downloaded it already.

tests/fixtures/direct_award_project_list_fixture.json
galund: Yeah I didn't know / had forgotten about the API stubs too. They also ought to move into dm-test-utils, I think?

samuelhwilliams: Oh, it uses some of the formatters defined in utils, so we'd end up with test-utils depending on utils to pull in the formatters. Not the biggest dependency, but... 🤷‍♂️

app/templates/direct-award/index.html
```diff
@@ -52,16 +52,23 @@ <h1>
...
     ) %}
       {% call summary.row() %}
-        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_family=framework.framework, project_id=item.id)) }}
+        {{ summary.service_link(item.name, url_for('direct_award.view_project', framework_family=framework.framework, project_id=item.id), wide=False) }}
         {{ summary.text(item.lockedAt | datetimeformat ) }}
+        {{ 
+          summary.text("Awarded") if item.outcome and item.outcome.result == "awarded"
+          else summary.text("The work has been cancelled") if item.outcome and item.outcome.result == "cancelled"
+          else summary.text("No suitable services found") if item.outcome and item.outcome.result == "none-suitable"
+          else summary.service_link("Tell us the outcome", url_for('direct_award.did_you_award_contract', framework_family=framework.framework, project_id=item.id), wide=False) }}
```
katstevens: Given this is an action, should the `Tell us the outcome` link be inside a `summary.field(action=True)` instead of a `summary.service_link`?

DilwoarH: @kubabartwicki ^^^

tests/fixtures/direct_award_project_list_fixture.json
samuelhwilliams: Ah that does make sense I suppose... Ok, if I get time during firebreak I'll shift them over to test utils.

app/templates/direct-award/index.html
DilwoarH: from @kubabartwicki  - "i’ve played with some variants but this worked best"

PR #782, opened 2018-06-11
Use the project outcome api

lfdebrux:
Ticket: [Trello](https://trello.com/c/BZbaASLO/143-link-frontend-of-did-you-award-flow-to-backend)

This PR uses the api calls introduced in https://trello.com/c/BZbaASLO/143-link-frontend-of-did-you-award-flow-to-backend to make the 'did you award the contract?' flow functional.

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -655,10 +655,10 @@ def which_service_won_contract(framework_family, project_id):
 
 
 @direct_award.route(
-    '/<string:framework_family>/projects/<int:project_id>/tell-us-about-contract',
+    '/<string:framework_family>/projects/<int:project_id>/outcomes/<int:outcome_id>/tell-us-about-contract',
     methods=['GET', 'POST']
 )
-def tell_us_about_contract(framework_family, project_id):
+def tell_us_about_contract(framework_family, project_id, outcome_id):
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
     framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_family)
 
@@ -681,10 +681,11 @@ def tell_us_about_contract(framework_family, project_id):
 
     return render_template(
         'direct-award/tell-us-about-contract.html',
-        project=project,
-        framework=framework,
-        form=form,
         errors=errors,
+        form=form,
+        framework=framework,
+        project=project,
+        outcome_id=outcome_id,
     ), 200 if not errors else 400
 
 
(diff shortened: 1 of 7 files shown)


app/main/views/g_cloud.py
```diff
@@ -638,9 +638,27 @@ def which_service_won_contract(framework_family, project_id):
...
+        except HTTPError:
+            flash(f"There was an error updating '{project['name']}', please try again later", 'error')
+            return render_template(
+                'direct-award/which-service-won-contract.html',
+                project=project,
+                errors={},
+                framework=framework,
+                services=services,
+                form=form,
+            ), 400
```
galund: So I think if we're doing this, we should proxy the error code from the API back to the client. It's not correct to give the client a 400 if the API gave us a 500!

However I think if we're aiming to fully validate the form before passing the data to the backend, then we don't need to catch errors here at all.

The specific error scenarios that the API can detect that the frontend cannot, like if you already set the outcome in another tab whilst leaving this form open in another, are often not helped by asking the user to try again later with the same form again!

If decide we must catch some `HTTPError`s like this then I think we should at least be re-raising for 5xx errors, so when diagnosing API errors we can see the full trace from the calling apps too.

risicle:
As you might have gathered from @galund and I'd conversation, you might have to hold back on merging these for a little bit as I'm changing the serialization format of the `resultOfDirectAward` section of the `Outcome` to be something like:

```
"resultOfDirectAward": {
    "search": {
        "id": self.direct_award_search_id,
    },
    "archivedService": {
        "id": self.direct_award_archived_service_id,
        "service": {
            "id": self.direct_award_archived_service.service_id,
        },
    },
},
```

app/main/views/g_cloud.py
```diff
@@ -708,6 +757,14 @@ def why_did_you_not_award_the_contract(framework_family, project_id):
...
     if form.validate_on_submit():
+        if form.why_did_you_not_award_the_contract.data == 'work_cancelled':
+            data_api_client.create_direct_award_project_outcome_cancelled(
+                project_id=project_id,
+                user_email=current_user.email_address)
+        elif form.why_did_you_not_award_the_contract.data == 'no_suitable_services':
+            data_api_client.create_direct_award_project_outcome_none_suitable(
+                project_id=project_id,
+                user_email=current_user.email_address)
         flash('You’ve updated ‘' + project['name'] + '’')
```
galund: Can we use format strings for these as it will make internationalization slightly easier if we ever do that

lfdebrux: Will do

app/main/views/g_cloud.py
lfdebrux: I will leave them out

lfdebrux:
I've squash-rebased in preparation for merging

PR #783, opened 2018-06-12
Make quotes curly

lfdebrux:
Some new flash messages used straight quotes and apostrophes instead of the curly ones.


andrewchong:
👍 Looks good to me.

PR #790, opened 2018-06-15
Change broken link to user research preferences

CrystalPea:
Trello ticket: https://trello.com/c/ZgQUKGI7/236-link-to-user-research-sign-up-page-instead-of-airtable

### DIFF
--- a/app/main/views/feedback.py
+++ b/app/main/views/feedback.py
@@ -1,6 +1,6 @@
 import requests
 
-from flask import current_app, request, redirect, flash, Markup
+from flask import current_app, request, redirect, flash, Markup, url_for
 from werkzeug.datastructures import MultiDict
 from werkzeug.urls import url_parse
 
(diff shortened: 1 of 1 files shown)


app/main/views/feedback.py
```diff
@@ -28,7 +28,8 @@ def send_feedback():
     flash(Markup(
         """Thank you for your message. If you have more detailed feedback, please email
         <a href="mailto:enquiries@digitalmarketplace.service.gov.uk">enquiries@digitalmarketplace.service.gov.uk</a> or
-        <a target="_blank" rel="external noopener noreferrer" href="https://airtable.com/shrkFM8L6Wfenzn5Q">take
+        <a target="_blank" rel="external noopener noreferrer"
+        href="https://www.digitalmarketplace.service.gov.uk/user/notifications/user-research">take
```
katstevens: We have an external route for this (`dmutils.external.user_research_consent`), so we should be able to avoid hardcoding the URL.

CrystalPea: Ahh, thank you 👍

CrystalPea:
@katstevens This url is now dynamically generated, thanks for the advice :)

PR #792, opened 2018-06-18
FE-2. Which service won the contract? page

DilwoarH:
Changes to fix the following comments:
- The list of radios is not using form-hint to display supplier names but instead uses question-description. This has a smaller font than form-hint and a different colour. Could we change please? Attached is a screen grab of how it should look like.
- The content is shifted left from the left so it's misaligned with headers and footers. Other pages are fine.

Ticket: https://trello.com/c/hHyDVwFZ/72-fe-2-which-service-won-the-contract-page

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -341,7 +341,7 @@ class TestDirectAwardProjectOverview(TestDirectAwardBase):
                                    '/buyers/direct-award/g-cloud/projects/1/results') is True
 
         assert self._task_has_link(tasklist, 6,
-                                   '/buyers/direct-award/g-cloud/projects/1/did-you-award-contract') is True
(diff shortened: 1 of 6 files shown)


tests/main/views/test_direct_award.py
```diff
@@ -769,12 +769,12 @@ def test_which_service_did_you_award_page_renders(self):
         assert len(doc.xpath(
             '//input[@type="radio"][contains(following-sibling::label, "The work has been cancelled")]')) == 1
         assert len(doc.xpath(
-            '//p[contains(normalize-space(text()), "For example, because you no longer have the budget")]\
+            '//span[contains(normalize-space(text()), "For example, because you no longer have the budget")]\
```
galund: hmmm - see comments on https://github.com/alphagov/digitalmarketplace-frontend-toolkit/pull/450/files

PR #798, opened 2018-06-26
"End search" becomes "before you export"

galund:
"End search" is now cast in terms "being ready to download".

* This is more a content change rather than new functionality, but there is a small form with a tick box that must be ticked.

https://trello.com/c/Z1a3weIe/131-replace-end-search-with-before-you-export-your-results

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -1,6 +1,6 @@
 from flask_wtf import FlaskForm
 
-from wtforms.fields import DecimalField, RadioField
+from wtforms.fields import DecimalField, RadioField, BooleanField
 from wtforms.validators import DataRequired, Length, NumberRange, Optional, InputRequired
 
 from dmutils.forms.fields import DateField
@@ -28,6 +28,13 @@ class DMRadioField(RadioField):
     def options(self):
         """The RadioField choices in a format suitable for the frontend toolkit"""
         return [{"label": label, "value": value} for value, label in self.choices]
+
+
+class DMBooleanField(BooleanField):
+    @property
+    def options(self):
+        # Even single boolean fields are expected to be in an 'options' list
+        return [{"label": self.label.text, "value": self.data}]
 
 
 class DidYouAwardAContractForm(FlaskForm):
@@ -125,3 +132,10 @@ class WhyDidYouNotAwardForm(FlaskForm):
 
         self.why_did_you_not_award_the_contract.choices = [(option['value'], option['label']) for option in options]
         self.why_did_you_not_award_the_contract.options = options
+
+
+class BeforeYouDownloadForm(FlaskForm):
+    user_understands = DMBooleanField(
+        "I understand that I cannot edit my search again after I export my results",
+        validators=[InputRequired(message="Please confirm that you understand before you continue.")]
+    )
--- a/app/templates/direct-award/end-search.html
+++ b/app/templates/direct-award/end-search.html
@@ -1,6 +1,6 @@
 {% extends "toolkit/layouts/_base_page.html" %}
 
-{% block page_title %}End your search - Digital Marketplace{% endblock %}
+{% block page_title %}Export your results - Digital Marketplace{% endblock %}
 
 {% block breadcrumb %}
   {%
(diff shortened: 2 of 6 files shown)


app/templates/direct-award/end-search.html
```diff
@@ -29,49 +29,61 @@
...
           data-analytics-category="Direct Award"
           data-analytics-action="End search"
           data-analytics-label="{{ search_count }}"
           {% if disable_end_search_btn %}disabled="disabled"{% endif %}
-          >End search and continue</button>
+          >Export results and continue</button>
       </form>
       
-      <p><a href="{{ url_for('direct_award.view_project', framework_family=framework.framework, project_id=project.id) }}">Return to overview</a></p>
+      <p><a href="{{ url_for('direct_award.view_project', framework_family=framework.framework, project_id=project.id) }}">Return to your task list</a></p>
```
lfdebrux: Can we change this to `framework.family`?

app/templates/direct-award/end-search.html
```diff
@@ -29,49 +29,61 @@
...
+      export your results.
+    </p>
+  </div>
+{% endif %}
+
 
 <div class="end-search-page">
   <div class="grid-row">
-    <div class="column-two-thirds">
+    <div class="column-two-thirds dmspeak">
```
lfdebrux: What does `dmspeak` do here?

app/templates/direct-award/end-search.html
```diff
@@ -29,49 +29,61 @@
...
       <header class="page-heading">
         <h1>
-          End your search
+          Before you export your results
         </h1>
       </header>
```
lfdebrux: I think there is a template for headings now, you can use
```
{% with heading="Before you export your results" %}
  {% include "toolkit/page-heading.html %}
{% endwith %}
```

app/main/forms/direct_award_forms.py
```diff
@@ -30,6 +30,13 @@ def options(self):
         return [{"label": label, "value": value} for value, label in self.choices]
 
 
+class DMBooleanField(BooleanField):
```
samuelhwilliams: Is this needed? It feels like we have this in utils now

app/templates/direct-award/end-search.html
risicle: Ahhhh dmspeak. Long story behind it I think but I think all you need to know is it's a dumping ground of all kinds of more "conversational" styles we use...

app/main/forms/direct_award_forms.py
galund: That PR didn't land yet :( https://github.com/alphagov/digitalmarketplace-utils/pull/400

app/templates/direct-award/end-search.html
galund: But ewww yuck no :-)

app/main/forms/direct_award_forms.py
samuelhwilliams: Aww, ok

app/templates/direct-award/end-search.html
```diff
@@ -29,49 +29,61 @@
...
 
 {% block main_content %}
+{% if errors %}
+  <div class="validation-masthead" aria-labelledby="validation-masthead-heading" aria-role="group" tabindex="-1">
+    <p>
+      You need to <a href="#user_understands">confirm that you’ve finished editing your search</a> before you
+      export your results.
+    </p>
+  </div>
+{% endif %}
```
samuelhwilliams: This feels like it should be using the validation masthead template from the frontend toolkit?

galund: It's based on that, but this is an intentional change because the form only has one question. So we don't need the 'heading plus list of question/errors' format, but something slightly different i.e. just one para of text.

samuelhwilliams: We use the standard heading+question for all the other one-question forms.

galund: Well the login screen has something different. The content here seems right to me, and has been discussed a fair amount.

app/templates/direct-award/end-search.html
```diff
@@ -29,49 +29,61 @@
 {% endblock %}
 
 {% block main_content %}
+{% if errors %}
+  <div class="validation-masthead" aria-labelledby="validation-masthead-heading" aria-role="group" tabindex="-1">
+    <p>
+      You need to <a href="#user_understands">confirm that you’ve finished editing your search</a> before you
```
samuelhwilliams: It would be nice to do something more like the toolkit here with `a href="#{{ form.user_understands.id }}"`.

PR #800, opened 2018-06-26
Future proofing saved search

Wynndow:
For [this trello card.](https://trello.com/c/LjbF4cLc)

The tests for the PR won't go green until [this frameworks PR](https://github.com/alphagov/digitalmarketplace-frameworks/pull/525) is merged and tagged.

As we move from G9 to G10 the behaviour of saved searches could be confusing for buyers. This PR is to add some temporary messages to explain exactly what is happening. See the screenshots below.

This PR relies on the dates in the Frameworks table being correct, especially for frameworkLiveAtUTC and frameworkExpiresAtUTC. These can be updated manually, however there is a plan to do some work on the api to ensure these dates are set automatically as much as can be.

Originally the content for this PR was in the Frameworks repo and was being pulled in. The last commit is a refactor to move the content into templates in this repo.

### Saved search not yet locked, before G10 goes live:
screen shot 2018-06-26 at 17 30 46

### Saved search not yet locked, after G10 goes live:
screen shot 2018-06-26 at 17 31 20

### Saved search is locked, before G10 goes live:
screen shot 2018-06-26 at 17 32 04

### Saved search is locked, after G10 goes live and before G9 expires:
screen shot 2018-06-26 at 17 32 56

### Saved search is locked, G10 is live, G9 is expired:
screen shot 2018-06-26 at 17 33 24


tests/main/helpers/test_framework_helpers.py
```diff
@@ -12,12 +17,63 @@ def setup_method(self, method):
...
     def test_get_lots_by_slug(self):
         g_cloud_9_data = next((f for f in self.available_frameworks if f['slug'] == 'g-cloud-9'))
-        lots_by_slug = framework_helpers.get_lots_by_slug(g_cloud_9_data)
+        lots_by_slug = get_lots_by_slug(g_cloud_9_data)
         assert lots_by_slug['cloud-hosting'] == g_cloud_9_data['lots'][0]
         assert lots_by_slug['cloud-software'] == g_cloud_9_data['lots'][1]
         assert lots_by_slug['cloud-support'] == g_cloud_9_data['lots'][2]
+
+
+class CustomAbortException(Exception):
```
samuelhwilliams: This is definitely a cross-app thing too.

app/main/helpers/search_save_helpers.py
```diff
@@ -38,3 +38,27 @@ def __init__(self, search_api_url, frameworks_by_slug, include_markup=False):
             filters.values(),
             lots_by_slug
         )
+
+
+def get_saved_search_temporary_message_status(project, framework, following_framework):
+    if not project['lockedAt']:
+        if following_framework['status'] in ['coming', 'open', 'pending', 'standstill']:
+            return 'not_locked_pre_live'
```
samuelhwilliams: These return value strings should probably be declared once as constant globals so that we aren't just hand writing them in various places.

We'll need to add them to the Jinja templating environment but that shouldn't be too difficult: http://jinja.pocoo.org/docs/2.10/api/#global-namespace

app/main/views/g_cloud.py
```diff
@@ -513,16 +513,33 @@ def view_project(framework_family, project_id):
...
-                           search=search,
-                           buyer_search_page_url=buyer_search_page_url,
-                           search_summary_sentence=search_summary_sentence,
-                           framework_urls=framework_urls,
-                           call_off_contract_url=framework_urls['call_off_contract_url'],
-                           project_outcome_label=project_outcome_label)
+    try:
+        following_framework = framework_helpers.get_framework_or_500(
+            data_api_client,
+            content_loader.get_metadata(framework['slug'], 'following_framework', 'slug'),
```
samuelhwilliams: Do we need to have this as defined metadata or could we safely derive it as the next framework in the same family from the database?

Eh, we could do it on the framework id but it's not super safe. I guess leave this as-is. 🤔

app/templates/direct-award/_saved-search-locked-project-temp-message.html
```diff
@@ -0,0 +1,38 @@
...
+         ' service</li><li><a href="{}">start a new search for '.format(buyer_search_page_url)|safe +
+         '{}'.format(following_framework.name)|nbsp + ' services</a></li></ul>'|safe),
+        'Any existing contracts for ' + '{}'.format(framework.name)|nbsp + ' services are still valid.'
+      ]
+    %}
+  {% elif temporary_message_status == 'locked_post_live_post_interim' %}
+    {% set temp_heading = 'The ' + '{}'.format(framework.name)|nbsp + ' services you found have expired' %}
+    {% set temp_messages = [
+        ('If you did not award a contract by {}, you need to <a href="{}">start a new search for '.format(framework.frameworkExpiresAtUTC|dateformat, buyer_search_page_url)|safe +
+        '{}'.format(following_framework.name)|nbsp + ' services</a>. If you did award a contract, please tell us the outcome.'|safe)
```
samuelhwilliams: `'{}'.format(following_framework.name)|nbsp ` can just be `following_framework.name|nbsp`, can't it?

tests/main/views/test_direct_award.py
```diff
@@ -350,6 +363,122 @@ def test_overview_renders_specific_elements_for_search_downloaded_state(self):
...
+        body = res.get_data(as_text=True)
+        doc = html.fromstring(body)
+
+        if position == 'sidebar':
+            assert len(doc.xpath("//div[@class='temporary-message']")) == 1
+            assert not doc.xpath("//div[@class='temporary-message-banner']")
+        elif position == 'banner':
+            assert len(doc.xpath("//div[@class='temporary-message-banner']")) == 1
+            assert not doc.xpath("//div[@class='temporary-message']")
+
```
samuelhwilliams: Maybe an `else raise` to catch a misconfigured parameter?

tests/main/views/test_direct_award.py
```diff
@@ -210,6 +211,18 @@ def _cannot_start_from_task(self, tasklist, cannot_start_from):
...
 
+    def _set_project_locked_and_framework_states(self, locked_at, framework_status, following_framework_status):
+        project = self._get_direct_award_project_fixture()
+        project['project']['lockedAt'] = locked_at
+        self.data_api_client.get_direct_award_project.return_value = project
+
+        self.data_api_client.find_frameworks.return_value = {
+            'frameworks': [framework(slug='g-cloud-9', status=framework_status)['frameworks']]
+        }
+        self.data_api_client.get_framework.side_effect = [
```
samuelhwilliams: Why `side_effect` here rather than `return_value`?

Wynndow:
@karlchillmaid It's because of the way that Python does string formatting. If you want to insert a variable into the middle of a string, you can use `format()`. Like this: `"I'm going to insert {} here".format(something)`
The `{}` is the thing that's going to be replaced, and the `something` between the brackets after `format` is what's replacing it. In the example on line 8,  the string actually includes the html as it needs to be passed to the template we're using. So the format goes on the end of the string, after the `</li>`. Make sense?

app/main/helpers/framework_helpers.py
```diff
@@ -45,3 +47,19 @@ def get_framework_description(data_api_client, framework_family):
 def abort_if_not_further_competition_framework(framework):
     if not framework['hasFurtherCompetition']:
         abort(404, f"Framework `{framework}` does not support further competition.")
+
+
+def get_framework_or_500(client, framework_slug, logger=None):
```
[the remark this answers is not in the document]
Wynndow: There's no where immediately obvious to put in in utils. Could potentially make a new module, `framework_helpers`, but then there's just going to be one thing in it which could easily be missed. Feels like common helpers from across the apps could be centralised in once go?

tests/main/helpers/test_framework_helpers.py
```diff
@@ -12,12 +17,63 @@ def setup_method(self, method):
 
     def test_get_latest_live_framework(self):
         latest_framework_when_fixture_updated = 'g-cloud-9'  # fixture set in base class
-        latest_framework_data = framework_helpers.get_latest_live_framework(self.available_frameworks, 'g-cloud')
+        latest_framework_data = get_latest_live_framework(self.available_frameworks, 'g-cloud')
```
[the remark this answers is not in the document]
Wynndow: Will do

app/main/helpers/framework_helpers.py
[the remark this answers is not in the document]
samuelhwilliams: Yep reasonable enough. No immediate concern with this being a bit duplicated.

app/main/helpers/search_save_helpers.py
Wynndow: The more I think about this the odder I find it. Totally get making the strings in constants, but at that point the string itself loses any real meaning. I'm only using them as way of identifying something and not for their actual content.
And once I've added them to the Jinja environment I'm going to have a similar issue, that I'm going to be handwriting the environment variables name, instead of just the string I'm using, so the same risk but with more faff. Maybe I'm missing something. What do you think?

app/main/views/g_cloud.py
Wynndow: Yeah, I thought about this a little. It would _probably_ be fine to derive it. But maybe not always. In the past we allowed suppliers to only have to apply to every other G-Cloud and two G-Clouds were live at the same time. This scenario would break `just pick the next framework iteration` model.

app/templates/direct-award/_saved-search-locked-project-temp-message.html
Wynndow: yes indeed.

tests/main/views/test_direct_award.py
Wynndow: sure

tests/main/views/test_direct_award.py
Wynndow: Yep. This was a hangover from a previous (wrong) version where I had two different returns - hence the list. Will change to `return_value`

Wynndow:
@samuelhwilliams I think everything is updated if you fancy another look.

app/__init__.py
```diff
@@ -101,6 +109,16 @@ def refresh_session():
...
 
+    @application.context_processor
+    def inject_saved_search_temp_message_statuses():
+        return dict(
+            NOT_LOCKED_PRE_LIVE=NOT_LOCKED_PRE_LIVE,
+            NOT_LOCKED_POST_LIVE=NOT_LOCKED_POST_LIVE,
+            LOCKED_PRE_LIVE=LOCKED_PRE_LIVE,
+            LOCKED_POST_LIVE_DURING_INTERIM=LOCKED_POST_LIVE_DURING_INTERIM,
+            LOCKED_POST_LIVE_POST_INTERIM=LOCKED_POST_LIVE_POST_INTERIM,
+        )
```
samuelhwilliams: If you change the states in search_save_helpers into an enum, you can simplify this code to avoid the name duplication:


```python
class SavedSearchStateEnum(enum.Enum):
    NOT_LOCKED_PRE_LIVE = 1
    NOT_LOCKED_POST_LIVE = 2
    LOCKED_PRE_LIVE = 3
    LOCKED_POST_LIVE_DURING_INTERIM = 4
    LOCKED_POST_LIVE_POST_INTERIM = 5
```

```python
@application.context_processor
    def inject_saved_search_temp_message_statuses():
        return {state.name: state.value for state in SavedSearchStateEnum}
```

app/main/helpers/search_save_helpers.py
```diff
@@ -38,3 +44,27 @@ def __init__(self, search_api_url, frameworks_by_slug, include_markup=False):
             filters.values(),
             lots_by_slug
         )
+
+
+def get_saved_search_temporary_message_status(project, framework, following_framework):
+    if not project['lockedAt']:
+        if following_framework['status'] in ['coming', 'open', 'pending', 'standstill']:
```
samuelhwilliams: Showing this message when the framework is `coming` feels super premature - it could be up to 3 months away at that point? I'd even question if `open` is too soon.

Be aware that when we show the messages we need to have the dates fairly locked in, which isn't always the case this early on. We should probably drop the message when the following framework is just `coming`, and we should consider probably just showing the month+year rather than exact date if it's in `open`.

Thoughts?

(For example DOS3 is currently coming but we have _no_ dates for it set up in the database yet)

karlchillmaid: +1

Agree with Sam that showing messages like this too early is not helpful for users.

app/__init__.py
Wynndow: Oh nice. I'll change.

app/main/helpers/search_save_helpers.py
Wynndow: Ah yeah, that's a a good point. Thinking about it I'd be tempted to only start showing the message once we get to pending. That would give a reasonable amount of time between the messages appearing and the new framework coming in. I'll chat to Patrick too, see what he thinks.

Wynndow: Have had a chat with Patrick, he thinks that we shouldn't show the messages before we enter 'pending', and from that point on we should only show it if the date has been set. We could have something like `if status in ['pending', 'standstill'] and framwork.FrameworkLiveAtUTC:`

Although I've just noticed the dates can't be `null`. Do we have a rule on what we set dummy dates too? Are they wildly in the past/future? Could set a check on if the date is after/before a certain time?

Wynndow: Have chatted further with Patrick re: always having a date set. We'e decided to only start showing the messages at the beginning of standstill, which should give buyers 6-8 working days to see the initial messages.

Do people think this is long enough. Bear in mind where we currently are in standstill, that these messages won't apply for a DOS framework, and what is probably going to happen with G11.

samuelhwilliams: I think I've set the dates for DOS3 to unix epoch, but there's no real standard. They are set by the developer when the new framework is posted to the API, so they can be anything.

We might just need to document that the framework should be created with as-accurate-as-possible dates, and that these need to be checked as the framework progresses.

Wynndow: @karlchillmaid @samuelhwilliams ^^^

Wynndow: Should have said, that we'll also be adding something to the run book and the dm manual to document that this date needs to be set.

samuelhwilliams: I think I agreed with your initial decision that pending+standstill is probably good. That should be something like 5 weeks. 10 days is _probably_ too short, given how it can take some time to actually compare services and that the process is fairly labour intensive. 5 weeks is probably longer than ideal, but I'd probably err on that side that too little time, at which point the banner could come too late for users to finish their procurement.

@karlchillmaid might have some ideas on which way to fall and how long it takes to compare services - I can't recall a length of time off the top of my head now...

Wynndow: The issue with pending is that we may not have the date from CCS by then. We didn't for this iteration. (Pending + conditional check on the date being set) would be perfect, but comparing the date to epoch feels like it might break.

samuelhwilliams: Yeah you're right. That's tricky.

Let's go for this then. We can always iterate in the future if we get evidence that it's too late 👍

samuelhwilliams: Actually I was being dumb. Standstill is definitely the right time given that we keep G9 services eligible for purchase for ~4 weeks after the new iteration goes live.

As you were!

app/main/helpers/search_save_helpers.py
```diff
@@ -38,3 +48,30 @@ def __init__(self, search_api_url, frameworks_by_slug, include_markup=False):
...
         )
+
+
+def get_saved_search_temporary_message_status(project, framework, following_framework):
+    if following_framework['status'] in ['coming', 'open', 'pending']:
+        return None
+
+    if not project['lockedAt']:
+        if following_framework['status'] == 'standstill':
+            return SavedSearchStateEnum.NOT_LOCKED_PRE_LIVE.value
```
samuelhwilliams: The name here is maybe slightly off now. Could be `NOT_LOCKED_STANDSTILL`?

app/main/helpers/search_save_helpers.py
```diff
@@ -38,3 +48,30 @@ def __init__(self, search_api_url, frameworks_by_slug, include_markup=False):
...
+        elif following_framework['status'] in ['live', 'expired']:
+            return SavedSearchStateEnum.NOT_LOCKED_POST_LIVE.value
+    else:
+        if following_framework['status'] == 'standstill':
+            return SavedSearchStateEnum.LOCKED_PRE_LIVE.value
```
samuelhwilliams: Same here.

app/main/views/g_cloud.py
```diff
@@ -513,16 +513,33 @@ def view_project(framework_family, project_id):
...
+        'direct-award/view-project.html',
+        framework=framework,
+        following_framework=following_framework,
+        project=project,
+        custom_dimensions=custom_dimensions,
+        search=search,
+        buyer_search_page_url=buyer_search_page_url,
+        search_summary_sentence=search_summary_sentence,
+        framework_urls=framework_urls,
+        call_off_contract_url=framework_urls['call_off_contract_url'],
```
samuelhwilliams: Weird that we pass this through separately when we pass all the framework urls through in the line above.

Not your code though.

app/templates/direct-award/_saved-search-not-locked-project-temp-message.html
```diff
@@ -0,0 +1,22 @@
+{% if not project.lockedAt and following_framework %}
+  <div class="column-one-third">
+    {% if temporary_message_status == NOT_LOCKED_PRE_LIVE %}
+      {% set temp_heading =  '{}'.format(following_framework.name)|nbsp + ' services will replace existing services on {}'.format(following_framework.frameworkLiveAtUTC|dateformat) %}
```
samuelhwilliams: We have a dmutils.dates helper that converts framework dates into their most common formats, so we could use that Python side rather than format here. Just call `update_framework_with_formatted_dates(following_framework)` and then reference it as `following_framework.frameworkLiveAt` in the template.

Optional change.

Wynndow: I'm gonna leave as is, as the template use both `dateformat` and `shortdateformat`

PR #796, opened 2018-06-21
Update the direct award task list

lfdebrux:
This PR collects together a group of changes to the direct award task list that were put together by our UCD people.

Note that when these changes are merged, they will trigger a visual regression.

---

### Tickets in this PR
- [Change step 1 of the task list](https://trello.com/c/2SvOyqcv) @lfdebrux
- [Update content on Save search, Export search and existing tasklist](https://trello.com/c/hOL3asOK)
    - Update content on export search @lfdebrux
    - Update content on save search @galund

---

### Changes between this branch and master

master                           | epic-capture-spend-data-tasklist-changes
-----------------------|---------
![master_view-project] | ![e6b5c3b_view-project]
![master_results]          | ![740133e_results]

[master_view-project]: https://user-images.githubusercontent.com/503614/41777169-62d667c8-7622-11e8-8193-33b9f7e5f5af.png
[c75e8fa_view-project]: https://user-images.githubusercontent.com/503614/41777197-77a4a822-7622-11e8-8928-315ecf5368fd.png
[c74168d_view-project]: https://user-images.githubusercontent.com/503614/41908516-26bda450-793c-11e8-95d7-16cc66e2785b.png
[e6b5c3b_view-project]: https://user-images.githubusercontent.com/503614/41912442-b35c7890-7947-11e8-9861-ac484af1ce13.png
[master_results]: https://user-images.githubusercontent.com/503614/41905672-c74e9aae-7933-11e8-95bc-bfd82c07acc1.png
[740133e_results]: https://user-images.githubusercontent.com/503614/41905698-d59646f2-7933-11e8-9435-dbef66ce1dd9.png

PR #801, opened 2018-06-27
Fix content in Customer Benefits Record step of tasklist

lfdebrux:
A small bit of content was missed in #796

[ticket]: https://trello.com/c/hOL3asOK

---

## Side by side view

screen shot 2018-06-27 at 14 20 40 2

### DIFF
--- a/app/templates/direct-award/view-project.html
+++ b/app/templates/direct-award/view-project.html
@@ -116,8 +116,9 @@
                                                                    data-analytics=\"trackEvent\"
                                                                    data-analytics-category=\"Direct Award\"
                                                                    data-analytics-action=\"External Link\"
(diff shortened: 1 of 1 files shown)


galund:
@karlchillmaid  FWIW I wonder if “They [CCS] will contact you if they find any issues with the supplier” sounds potentially a little ominous, given that people will have _already awarded the contract_ by the time they get around to talking to CCS :-)

karlchillmaid:
@galund I left a reply in Slack just before it went down.

I totally agree with that possible interpretation of the line, although I did try to make it sound as friendly as possible.

In our discussion with CCS this was the (only) buyer benefit that was identified as an incentive for the CBR form.

I'm tempted to put it out there and see how it performs. FWIW no one objected to it in user testing, but the number of participants was too low to be sure either way.

:)

PR #795, opened 2018-06-20
Change the layout of the search page

lfdebrux:
Ticket: [trello]

Our designers @kubabartwicki and Karl Chillmaid have changed the layout
of the services search page to try and encourage users to notice and use
the 'Save your search' feature.

This PR implements those changes, which are described in detail [in the ticket][trello].
In summary, the search summary and save button are moved to the top of
the page, and the styling of the text for both is tweaked.

This commit also touches the briefs search page slightly, reducing the
margin between the page heading and subheading.

services search page changes
briefs search page changes

[trello]: https://trello.com/c/4ETdzfOe/

PR #802, opened 2018-06-28
Move search suggestions in search layout

lfdebrux:
@samuelhwilliams spotted that in the 0 results case the search suggestions look out of place since #795.

This commit fixes this by moving the search suggestions out of the search summary component to inside the search results wrapper.

The suggestions have to be inside the results wrapper so it is updated dynamically with the live search.

---

## Side by side

screen shot 2018-07-02 at 15 49 22 2

### DIFF
--- a/tests/main/views/test_marketplace.py
+++ b/tests/main/views/test_marketplace.py
@@ -1671,14 +1671,16 @@ class TestCatalogueOfBriefsPage(BaseApplicationTest):
 
         res = self.client.get('/digital-outcomes-and-specialists/opportunities')
         assert res.status_code == 200
-        suggestion = re.findall(r'<p>Suggestions:<\/p>', res.get_data(as_text=True))
-        assert len(suggestion) == 1
+
+        xpath = html.fromstring(res.get_data(as_text=True)).xpath
+        assert xpath('boolean(//div[contains(@class, "search-suggestion")])')
 
     def test_should_not_render_suggestions_when_results(self):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities')
         assert res.status_code == 200
-        suggestion = re.findall(r'<p>Suggestions:<\/p>', res.get_data(as_text=True))
-        assert len(suggestion) == 0
+
+        xpath = html.fromstring(res.get_data(as_text=True)).xpath
+        assert not xpath('boolean(//div[contains(@class, "search-suggestion")])')
 
     def test_should_ignore_unknown_arguments(self):
         res = self.client.get('/digital-outcomes-and-specialists/opportunities?location=my-lovely-horse')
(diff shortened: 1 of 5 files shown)


lfdebrux:
It would be worth having a functional test on the dynamic behaviour of the search suggestions (i.e. it appears and disappears correctly when using live search)

kubabartwicki:
Hey @samuelhwilliams @lfdebrux 

Thanks for this. While having an empty right column and an occasional page jump isn't ideal, I think it's important that the `Suggestions` bit and the search summary `0 results found [...]` sit close to each other. With some extra padding added above the `Save your search` button, I don't think it looks strange at all, so not sure if we need to go ahead with this PR?

samuelhwilliams:
I guess I disagree. I think it looks pretty weird and that the page jump is quite jarring.

I understand the idea of proximity with these two elements, but I don't feel like they're too far apart here. Happy to be overruled, as design is your domain, but it felt broken enough when I encountered it to bring it up - and I think the other people who've seen it so far agreed.

lfdebrux:
I talked to @kubabartwicki about this in a bit more detail, we agreed on two options:

 1. Add a margin to the bottom of the suggestions text
    - pros: quick and easy, improves spacing at top of page
    - cons: leaves a lot of empty space, doesn't fix 'jump'

  2. Move the suggestions to the right column and add/improve the intro text so it is clear that the suggestions are related to the fact they are zero searches
    - pros: fills the empty space, fixes jump
    - cons: more work, content needs to be common to both direct award and further competition searches  

I've asked @karlchillmaid to look at content

karlchillmaid:
@lfdebrux 

New content

<p>Improve your search results by:</p>
  <ul class="list list-bullet">
      <li>removing filters</li>
      <li>choosing a different category</li>
      {% if framework_family == 'g-cloud' %}
        <li>double-checking your spelling</li>
        <li>using fewer keywords</li>
        <li>searching for something less specific, you can refine your results later</li>

tests/main/views/test_marketplace.py
```diff
@@ -1671,13 +1671,13 @@ def test_should_render_suggestions_for_0_results(self):
 
         res = self.client.get('/digital-outcomes-and-specialists/opportunities')
         assert res.status_code == 200
-        suggestion = re.findall(r'<p>Suggestions:<\/p>', res.get_data(as_text=True))
+        suggestion = re.findall(r'<div class="search-suggestion">', res.get_data(as_text=True))
```
samuelhwilliams: This feels a bit fragile. If we add a new class to this div then the test will break.

Can we use xpath `contains(@class, 'search-suggestion')` maybe instead?

lfdebrux: Good point, will fix

PR #804, opened 2018-06-28
FE-8 'We are still assessing' flow

lfdebrux:
Record when user states they are still assessing

[Ticket](https://trello.com/c/5Lrp1lOK)

project 1234 _microservices_ - digital 1

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import inflection
 from operator import itemgetter
 
-from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
+from flask import abort, render_template, request, redirect, current_app, url_for, flash
 from flask_login import current_user
 from werkzeug.urls import Href, url_encode, url_decode
 
@@ -623,8 +623,8 @@ def did_you_award_contract(framework_family, project_id):
                 project_id=project_id,
                 user_email=current_user.email_address)
             flash(
-                Markup(f"Your response for ‘{project['name']}’ has been saved.<br/>"
-                       "You still need to tell us the outcome when you've finished assessing services."),
+                f"Your response for ‘{project['name']}’ has been saved. "
+                "You still need to tell us the outcome when you’ve finished assessing services.",
                 'success'
             )
             return redirect(url_for('.view_project',
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -3,7 +3,7 @@ from datetime import datetime
 import inflection
 from operator import itemgetter
 
-from flask import abort, render_template, request, redirect, current_app, url_for, flash
+from flask import abort, render_template, request, redirect, current_app, url_for, flash, Markup
 from flask_login import current_user
 from werkzeug.urls import Href, url_encode, url_decode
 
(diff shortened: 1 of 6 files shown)


app/main/views/g_cloud.py
```diff
@@ -596,6 +596,14 @@ def did_you_award_contract(framework_family, project_id):
     form = DidYouAwardAContractForm()
     if form.validate_on_submit():
         if form.did_you_award_a_contract.data == form.STILL_ASSESSING:
+            data_api_client.mark_direct_award_project_as_still_assessing(
+                project_id=project_id,
+                user_email=current_user.email_address)
+            flash(
+                Markup(f"Your response for ‘{project['name']}’ has been saved.<br/>"
```
galund: This looks like an injection vulnerability to me - if you're wrapping in `Markup` then you're claiming the content is safe. But you haven'y escaped project name?

I don't think you need the `Markup` at all - a plain string would be fine?

galund: Oh, except you have a `<br>`... I suppose you will need to escape the project name separately then.

galund: Have removed the `<br>` as it feels wrong to have a line break here style-wise.

tests/main/views/test_direct_award.py
```diff
@@ -672,6 +672,26 @@ def test_award_contract_we_are_still_assessing_redirects_on_post(self, choice, e
...
+        self.data_api_client.mark_direct_award_project_as_still_assessing.assert_called_once()
+
+    def test_still_assessing_flashes(self):
+        self.login_as_buyer()
+
+        self.client.post('/buyers/direct-award/g-cloud/projects/1/did-you-award-contract',
+                         data={'did_you_award_a_contract': 'still-assessing'})
+
+        self.assert_flashes(
+            "Your response for ‘My procurement project &lt;’ has been saved. "
```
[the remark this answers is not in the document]
galund: A project name - or any other user input - can contain characters that might be significant to HTML, and so have to be escaped. We don't expect the API to prevent users from using angle brackets (etc) in their names. We expect the frontends to handle strings safely, i.e. know which strings are safe and which ones aren't.

Jinja and friends tend to handle this right (magic is in the `escape` function) - but not if (as was the case here) if we are combining user input with custom markup (a `<br>` tag as it happened, see the diff for my last commit).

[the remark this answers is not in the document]
katstevens: Ah - so if we have got rid of the `<br>`, do we still need this?

app/main/views/g_cloud.py
```diff
@@ -619,6 +619,14 @@ def did_you_award_contract(framework_family, project_id):
     form = DidYouAwardAContractForm()
     if form.validate_on_submit():
         if form.did_you_award_a_contract.data == form.STILL_ASSESSING:
+            data_api_client.mark_direct_award_project_as_still_assessing(
+                project_id=project_id,
+                user_email=current_user.email_address)
+            flash(
+                f"Your response for ‘{project['name']}’ has been saved. "
```
DilwoarH: Would it be better to have something like this ?
```
"Your response for ‘{}’ has been saved. ".format(project['name'])
```

galund: As discussed, but for posterity: it's the same thing.

Definitely you could argue that using `f` strings makes things look like templating, so maybe makes things look safe that aren't. But in both cases here we just end up with an unsafe string, so jinja (well, markupsafe's `escape` function specifically) will do what needs to be done.

tests/main/views/test_direct_award.py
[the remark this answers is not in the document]
galund: As discussed - but for posterity: yeah it's good to be checking we're doing the right thing, as introducing new `Markup` instances with XSS vulns is *really* easy to do.

PR #805, opened 2018-06-29
Direct award pre-project task list

risicle:
Trello https://trello.com/c/YlBpmOkZ/136-gcloud-start-on-the-task-list

 task list: add new view, pre_project_task_list, and adapt views leading into direct-award buyer journey

pre project task list

We now show this, initial, version of the task list when entering the journey and the existing lot selection view is renamed to `choose_lot` and placed at a new url. the "Start a new search" action button here now leads to that page.

This also introduces views that are under the "direct award" url path but are not login-restricted, so a new `direct_award_public` blueprint is introduced for this purpose.

The template is the existing "view project" template, which ironically seems to render relatively well if called without a `project`.

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -38,7 +38,7 @@ from ..helpers import framework_helpers
 from ..helpers.direct_award_helpers import is_direct_award_project_accessible, get_direct_award_projects
 from ..helpers.search_save_helpers import get_saved_search_temporary_message_status, SearchMeta
 from ..helpers.shared_helpers import get_fields_from_manifest, get_questions_from_manifest_by_id
-from ...main import main, direct_award
+from ...main import main, direct_award, direct_award_public
 from ..presenters.search_presenters import (
     filters_for_lot,
     set_filter_states,
@@ -58,9 +58,32 @@ TOO_MANY_RESULTS_MESSAGE = f"You have too many services to review. Refine your s
 
 @main.route('/g-cloud')
 def index_g_cloud():
+    return redirect(url_for("direct_award_public.pre_project_task_list", framework_family="g-cloud"), 301)
+
+
+@direct_award_public.route("/<string:framework_family>/start", methods=("GET",))
+def pre_project_task_list(framework_family):
+    frameworks_by_slug = framework_helpers.get_frameworks_by_slug(data_api_client)
+    framework = framework_helpers.get_latest_live_framework(frameworks_by_slug.values(), "g-cloud")
+
+    content_loader.load_messages(framework['slug'], ['urls'])
+    framework_urls = content_loader.get_message(framework['slug'], 'urls')
+
+    return render_template(
+        'direct-award/view-project.html',
+        project=None,
+        search=None,
+        framework=framework,
+        framework_urls=framework_urls,
+        call_off_contract_url=framework_urls['call_off_contract_url'],
+    )
+
+
+@direct_award_public.route('/<string:framework_family>/choose-lot')
+def choose_lot(framework_family):
     # if there are multiple live g-cloud frameworks, assume they all have the same lots
     all_frameworks = data_api_client.find_frameworks().get('frameworks')
-    framework = framework_helpers.get_latest_live_framework(all_frameworks, 'g-cloud')
+    framework = framework_helpers.get_latest_live_framework(all_frameworks, framework_family)
 
     content_loader.load_messages(framework['slug'], ['advice', 'descriptions'])
     gcloud_page_title = content_loader.get_message(framework['slug'], 'descriptions', 'framework')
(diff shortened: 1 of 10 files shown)


app/templates/direct-award/view-project.html
```diff
@@ -29,12 +34,21 @@
...
       {% with
-        heading = project.name
+        heading = project.name|default("Find cloud hosting, software and support")
       %}
         {% include "toolkit/page-heading.html" %}
       {% endwith %}
     </div>
   </div>
+  {% if not project %}
+    <p class="lede">
```
galund: I don't even think this *is* a lede, so let's just choose a different word that isn't quite so obscure?

Style-wise: can't it just be inside a `column-two-thirds`? Does it really need a style of its own?

app/main/views/g_cloud.py
```diff
@@ -543,6 +548,24 @@ def view_project(framework_family, project_id):
     )
 
 
+@direct_award_public.route("/<string:framework_family>/start-project", methods=("GET",))
```
galund: Obviously it doesn't matter but could this go further up the file, it'd be nice if they somehow got more into the order in which we expect people to see them, in terms of aiding understanding of the code...

app/templates/direct-award/view-project.html
```diff
@@ -18,7 +19,11 @@
           "link": url_for('direct_award.saved_search_overview', framework_family=framework.framework),
           "label": "Your saved searches"
       }
-    ]
+    ] if project else [
+      {
+          "label": "Find cloud hosting, software and support"
```
galund: It feels weird that this template is sometimes used in one breadcrumbs style (i.e. with the last item being a link) and the other (i.e. with no link, like here when there's no project)!

Maybe we can just add the current page name here (with no link). And seeing as it's repeated in two other places (in the title and in the heading), then maybe it could then be a variable?

app/templates/direct-award/view-project.html
risicle: I haven't chosen the name - it comes from a style in the govuk toolkit which we don't have for some reason and the scss itself is copypasta'd from ... the supplier app?

app/main/views/g_cloud.py
risicle: Yeah mentally flipped a coin on this one and in the end preferred breaking as little `git blame` information as possible.

app/templates/direct-award/view-project.html
risicle: I don't *think* variables can span across jinja `{% block%}`s otherwise I would have done so.

risicle: re other stuff, doing it like this is just what kuba said in the ticket, so...

app/main/views/g_cloud.py
risicle: wait i'm thinking of a different view...

app/templates/direct-award/view-project.html
galund: Re variables, it definitely does work - see https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/796/files#diff-65331f8ea2a95fe2321e350a64964497R3 :-)

app/templates/direct-award/view-project.html
galund: Ouch well it ought to go into our frontend toolkit then, but whatevs.

app/templates/direct-award/view-project.html
risicle: Oh! Think I was assuming because of the django template lang. Wonder how it copes with the scope ambiguity.

app/templates/direct-award/view-project.html
risicle: Kinda, maybe, eventually, or maybe we get around to doing design system stuff first...

app/main/views/g_cloud.py
```diff
@@ -58,9 +58,32 @@
 
 @main.route('/g-cloud')
 def index_g_cloud():
+    return redirect(url_for("direct_award_public.pre_project_task_list", framework_family="g-cloud"), 301)
+
+
+@direct_award_public.route("/<string:framework_family>/start-project", methods=("GET",))
```
lfdebrux: @karlchillmaid any thoughts on this url?

lfdebrux:
If we're changing other urls to `/direct-award/<framework_family>/...` should we change the g-cloud search url to `/direct-award/g-cloud/search`?

app/main/views/g_cloud.py
karlchillmaid: Are we doing something similar for DOS framework start pages too? (if so, what?)
Does it need to say '/start-project' – as project is not a term we user with end users (at the moment)?
Would `/start` work on it's own? 🤔

risicle:
That was quicker than expected.

PR #807, opened 2018-07-03
start project page: flag external links as such

risicle:
Trello https://trello.com/c/YlBpmOkZ/136-gcloud-start-on-the-task-list

### DIFF
--- a/app/templates/direct-award/view-project.html
+++ b/app/templates/direct-award/view-project.html
@@ -45,9 +45,21 @@
   {% if not project %}
     <p class="lede">
       Before you start you should
-      <a href="https://www.gov.uk/guidance/g-cloud-buyers-guide#requirements">write a list of requirements</a>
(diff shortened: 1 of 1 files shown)


app/templates/direct-award/view-project.html
```diff
@@ -45,9 +45,21 @@
   {% if not project %}
     <p class="lede">
       Before you start you should
-      <a href="https://www.gov.uk/guidance/g-cloud-buyers-guide#requirements">write a list of requirements</a>
+      <a href="https://www.gov.uk/guidance/g-cloud-buyers-guide#requirements"
+         target="_blank" rel="external noopener noreferrer"
```
galund: is there a good reason why we have noreferrer on all these links? cc @DilwoarH 

(I'd have thought that for gov.uk links then we should be a bit more chilled about this stuff, but maybe I'm not thinking this through)

PR #808, opened 2018-07-03
Add "Completed" to step 1 in task list when saved search has fewer than 30 results

lfdebrux:
@kubabartwicki [noticed][ticket] that there was no completed button on step 1 of the task list when the search has been saved but not locked.

This PR fixes it so that the box is shown, but only when the saved search has fewer than 30 results.

[ticket]: https://trello.com/c/JDIYkGdG

---

## When the search has fewer than 30 results
screen shot 2018-07-03 at 13 07 40

## When the search has more than 30 results
screen shot 2018-07-03 at 13 06 22

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -492,6 +492,7 @@ def view_project(framework_family, project_id):
         search_meta = SearchMeta(search['searchUrl'], frameworks_by_slug)
 
         search_summary_sentence = search_meta.search_summary.markup()
+        search_results_count = int(search_meta.search_summary.count)
         framework = frameworks_by_slug[search_meta.framework_slug]
         buyer_search_page_url = search_meta.url
 
@@ -500,6 +501,7 @@ def view_project(framework_family, project_id):
 
         search = None
         buyer_search_page_url = None
+        search_results_count = None
         search_summary_sentence = None
 
     content_loader.load_messages(framework['slug'], ['urls'])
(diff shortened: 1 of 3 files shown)


app/main/views/g_cloud.py
```diff
@@ -556,6 +558,7 @@ def view_project(framework_family, project_id):
         following_framework=following_framework,
         project=project,
         custom_dimensions=custom_dimensions,
+        count=search_summary_count,
```
galund: It might be nicer to pass the entire `search_meta` into the template rather than the two pieces? I know that's a bit more refactoring but feels a bit more future-proof?

lfdebrux: I agree with what you're saying in principle, but in the interests of keeping this a quick fix I'm going to leave it as is.

I'm also not sure that search_meta is the best place to be getting the count from; the code around it and `SearchSummary` is quite convoluted and I think refactoring them should definitely be done, but not in this PR.

app/main/views/g_cloud.py
```diff
@@ -500,6 +501,7 @@ def view_project(framework_family, project_id):
 
         search = None
         buyer_search_page_url = None
+        search_summary_count = 0
```
galund: Default None might be better in case someone was tempted to do some math with that zero :-)

Can we call it `search_results_count` as we're not counting summaries..

lfdebrux: Done

PR #809, opened 2018-07-03
Direct Award flow: combine export/download steps; add Ready to Assess

galund:
We're combining the _export_ (=='end search') step and the _download_ step, and we're adding a new _start assessing_ step, which in terms of actions just requires a single button press on the task list.

Probably loads of stuff wrong with this, but feels like I might as well get the PR up there already...

There's also a little tidy-up / deduplication of the logic for determining whether a search can be "ended" (which for users we are now calling "exporting").

screen shot 2018-07-04 at 11 21 14
screen shot 2018-07-04 at 11 22 20
screen shot 2018-07-04 at 11 22 39

Depends on:
* https://github.com/alphagov/digitalmarketplace-api/pull/814
* https://github.com/alphagov/digitalmarketplace-apiclient/pull/150
* https://github.com/alphagov/digitalmarketplace-frameworks/pull/530

### DIFF
--- a/app/main/helpers/dm_google_analytics.py
+++ b/app/main/helpers/dm_google_analytics.py
@@ -1,0 +1,36 @@
+from enum import Enum
+
+
+class CurrentProjectStageEnum(Enum):
+    SAVED = 'save_and_refine_search'
+    ENDED = 'search_ended'
+    DOWNLOADED = 'download_results'
+    READY_TO_ASSESS = 'ready_to_assess'
+    # the following correspond with digitalmarketplace-api:app.models.outcomes.Outcome.RESULT_CHOICES
+    FAILED = 'none-suitable'
+    CANCELLED = 'cancelled'
+    AWARDED = 'awarded'
+
+
+CUSTOM_DIMENSION_IDENTIFIERS = {
+    CurrentProjectStageEnum: 8
+}
+
+
+def custom_dimension(custom_dimension_enum, value):
+    """
+    Create a custom dimension dictionary for Google Analytics.
+
+    Pass in the relevant enum from this module, and either an instance or an actual value from that enum.
+    """
+
+    if value in CurrentProjectStageEnum:
+        custom_dimension_instance = value
+    else:
+        # If we've been given a string, try to convert it to an instance of the enum - if that fails then it'll be loud.
+        custom_dimension_instance = custom_dimension_enum(value)
+
+    return {
+        "data_id": CUSTOM_DIMENSION_IDENTIFIERS[custom_dimension_enum],
+        "data_value": custom_dimension_instance.value,
+    }
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -608,7 +608,7 @@ def end_search(framework_family, project_id):
 
         flash(PROJECT_ENDED_MESSAGE, 'success')
 
-        return redirect(url_for('.view_project', framework_family=framework_family, project_id=project['id']))
+        return redirect(url_for('.search_results', framework_family=framework_family, project_id=project['id']))
 
     errors = get_errors_from_wtform(form)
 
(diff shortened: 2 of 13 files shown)


app/main/views/g_cloud.py
```diff
@@ -526,6 +531,8 @@ def view_project(framework_family, project_id):
 
     if project['outcome']:
         current_project_stage = project['outcome']['result']
+    elif project['readyToAssessAt']:
+        current_project_stage = 'ready_to_assess'
```
galund: Need to check this with @DilwoarH

samuelhwilliams: If these are strings that define some set of finite states, it would be nice to use a proper enum (e.g. https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/ac6532f32d4141fd830e92c89253251d8ca2e79b/app/main/helpers/search_save_helpers.py#L15)

app/main/views/g_cloud.py
```diff
@@ -622,6 +630,31 @@ def end_search(framework_family, project_id):
     ), 200 if not errors else 400
 
 
+@direct_award.route('/<string:framework_family>/projects/<int:project_id>', methods=['POST'])
+def update_project(framework_family, project_id):
+    """Generic view for bits of the task list that need to POST stuff."""
```
samuelhwilliams: This doesn't feel like a particularly generic view if all it can do is set `readyToAssess`?

galund: Well, the URI is generic and it's referenced from a `<form>` that that entirely wraps the page. Potentially the same view could be used for other interactive or ajaxy things going on on that page.

I was just hoping to encourage people not to make another view that did the same thing. Can reword if necessary.

app/main/helpers/dm_google_analytics.py
```diff
@@ -0,0 +1,36 @@
...
+
+class CurrentProjectStageEnum(Enum):
+    SAVED = 'save_and_refine_search'
+    ENDED = 'search_ended'
+    DOWNLOADED = 'download_results'
+    READY_TO_ASSESS = 'ready_to_assess'
+    # the following correspond with digitalmarketplace-api:app.models.outcomes.Outcome.RESULT_CHOICES
+    FAILED = 'none-suitable'
+    CANCELLED = 'cancelled'
+    AWARDED = 'awarded'
```
risicle: :crystal_ball: My crystal ball can see us needing to be able to refer to a generic project state from other apps, and when that happens we should probably move this and implement it in-api. Preferably as a `@hybrid_property`.

Also I'd personally call this idea "project current stage" because "current project stage" has the ambiguity over which of "(current project) stage" vs "current (project stage)" it means. Whereas "(project current)" clearly doesn't mean anything unless we're getting into the dried fruit business.

It's not super important though.

tests/main/views/test_direct_award.py
```diff
@@ -200,6 +201,13 @@ def _task_has_link(self, tasklist, task, link):
 
         return len(anchors) == 1
 
+    def _task_has_button(self, tasklist, task, name, value):
+        """Task here refers to the tasklist item number rather than the zero-indexed python array. This feels easier to
+        understand and amend in the context of the tasklist."""
+        anchors = tasklist[task - 1].xpath('.//button[@name="{}" and @value="{}"]'.format(name, value))
```
risicle: FWIW there's a trick you can do with lxml's `xpath()` here so that you don't have to worry about escaping:

```
.xpath(".//button[@name=$name and @value=$val]", name=name, val=value)
```

app/main/helpers/dm_google_analytics.py
lfdebrux: I agree the project state should be in the api. This is a perfectly cromulent solution for now though.

tests/main/views/test_direct_award.py
lfdebrux: not worrying about escaping?? 👿 blasphemy 👿

risicle: Well, y'know, in that it's no longer an issue - with `.format()` there's nothing stopping the inserted string containing `"` or some non-xpath-valid character...

app/main/helpers/dm_google_analytics.py
risicle: :jebediah_springfield:

PR #811, opened 2018-07-05
Refactoring helper functions that use the search API client

katstevens:
Helpers should take an instance of the `search_api_client` as an argument, rather than importing directly from the app level. 

In this case it's only two helpers: `SearchMeta` and `build_lots_and_categories_link_tree`.

Advantages:
- consistent with how our helpers use the `data_api_client`
- simplifies patching setup in tests
- (in my opinion) helper functions shouldn't care about app-level objects, and certainly shouldn't need to deal them separately from the view
- it'll play nicely with the test helper base class I'm working on 😸

### DIFF
--- a/app/main/helpers/search_save_helpers.py
+++ b/app/main/helpers/search_save_helpers.py
@@ -3,7 +3,7 @@ from enum import Enum
 from flask import abort, current_app, url_for
 from werkzeug.datastructures import MultiDict
 
-from app import search_api_client, content_loader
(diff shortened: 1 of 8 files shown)


risicle:
Sure, DI is one way of doing it, and this looks like a perfectly neat way of doing it - did we discard adding it to the flask request context?

katstevens:
I guess we could, though I think that might make the tests more messy rather than less 😿

PR #814, opened 2018-07-12
APIClientMixin test class for patching data_api_client and search_api_client

katstevens:
Trello: https://trello.com/c/1CrEVFLP/25-dataapiclient-patching-part-2-consolidate-into-base-test-class

How it works:
- For each test module, create a subclass of the `BaseApiClientMixin` and initialise it with the client patch path for the view you want to patch (e.g. `app.main.views.marketplace.data_api_client` in `test_marketplace.py`). 
- You can potentially patch the client from two different paths in the same test module, with two different mixins.
- Add the subclassed mixin to each test class as the first argument (i.e. before `BaseApplicationTest`).
- Boom! `self.data_api_client` and `self.search_api_client` are available for use throughout the test class. The patches will be reset after every test. 

We can also now setup 'global' defaults for the client methods - I've done `data_api_client.find_frameworks` as an example. 

Also, it's optional! One class in the Buyer FE app (`TestDirectAwardURLGeneration`) requires the `search_api_client` *not* to be patched entirely, as it tests one of the client's methods that doesn't actually hit the API. Happily we can still use the previous 'patch certain methods in the setup' technique instead.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -14,6 +14,41 @@ from app import create_app, data_api_client
 from tests import login_for_tests
 
 
+class BaseDataAPIClientMixin:
+    """
+    Mixin for patching the API client when imported for each view module.
+
+    Import this base class to the test module, and initialise with the path to the import:
+
+    :: test_marketplace.py
+    class DataAPIClientMixin(BaseDataAPIClientMixin):
+        data_api_client_patch_path = 'app.main.views.marketplace.data_api_client'
+
+    class TestMyView(DataAPIClientMixin, BaseApplicationTest):
+        ...
+
+    Multiple subclasses of the mixin can be created and passed into the test class if the API client
+    is patched in different places (hopefully this should be rare!).
+    """
+    data_api_client_patch_path = None
+    data_api_client_patch = None
+
+    @classmethod
+    def setup_class(cls):
+        if cls.data_api_client_patch_path:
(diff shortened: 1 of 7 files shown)


tests/helpers.py
```diff
@@ -14,12 +14,109 @@
 from tests import login_for_tests
 
 
+def get_frameworks_list_fixture_data():
+    old_gcloud_lots = [api_stubs.lot(lot_id=1, slug='saas', name='Software as a Service'),
```
samuelhwilliams: `old_gcloud_lots` and `new_gcloud_lots` should already be defined in the api_stubs file I think, so I reckon you could import those directly if you need them.

samuelhwilliams: Oh you just lift and shifted this. Nevermind then maybe 🤷‍♂️

katstevens: Yeah, I just c+p-ed this as an example. In theory this could go in `dm_test_utils` or be replaced by api stubs.

lfdebrux:
This is awesome 👏 👏 👏 

Only one concern; would it be better if `BaseAPIClientMixin` was in `dm_test_utils`?

katstevens:
Ideally `BaseApplicationTest` would be in there as well! (Probably not quite ready for that yet.)

We don't actually have `dmtestutils` as a dependency for the Buyer FE yet 🤔 I think I'd do any moving/reshuffling in a separate PR(s).

lfdebrux:
That makes sense

PR #815, opened 2018-07-13
QA fixes to #812

lfdebrux:
Thanks to @kubabartwicki for spotting these

- Fix breadcrumb in choose-lot.html
- Supress empty query string for all lots
- Ensure that font size is dynamic

### DIFF
--- a/tests/main/views/test_g_cloud.py
+++ b/tests/main/views/test_g_cloud.py
@@ -29,7 +29,7 @@ class TestGCloudIndexResults(BaseApplicationTest):
         )
 
         assert res.status_code == 302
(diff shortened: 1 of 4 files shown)


tests/main/views/test_g_cloud.py
```diff
@@ -29,7 +29,7 @@ def test_posting_lot_redirects_to_search_with_lot(self, lot):
         )
 
         assert res.status_code == 302
-        assert res.location.endswith(f"/g-cloud/search?lot={lot}")
+        assert res.location.endswith("/g-cloud/search" + f"?lot={lot}" if lot else "")
```
risicle: I usually wrap ternary statements in brackets because it's hard to remember what the order of precedence is, but not important.

PR #821, opened 2018-07-25
Format links in briefs clarification questions

lfdebrux:
@louzoid-gds noticed that clarification questions with URLs in them do 
not render well on mobile (see [this 2nd line ticket][trello]).

This commit adds the `format_links` filter to the text of clarification 
questions, which as well as turning URLs into links, gives them 
line-breaking CSS properties that alleviate this issue.

[trello]: https://trello.com/c/y55PPcJ8

---

### Before

A screenshot of a table rendered in a mobile viewport where a long URL has caused the content in the left cell to overflow and push the right cell out of the viewport

### After

A screenshot of the same content as above, except the long URL is now a link and is wrapped in several places allowing the content in the left cell to remain within its boundaries

### DIFF
--- a/app/templates/_brief_q_and_a.html
+++ b/app/templates/_brief_q_and_a.html
@@ -25,7 +25,7 @@
   {% call summary.row() %}
     {% call summary.field(first=True, wide=False) -%}
       <span aria-label="question">{{ question.number }}.</span>
-      {{ question.question | preserve_line_breaks }}
+      {{ question.question | format_links | preserve_line_breaks }}
     {%- endcall %}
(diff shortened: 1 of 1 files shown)


katstevens:
Looks good - do we have any other blobs of content that this might apply to? Might be worth grepping the templates for the `preserve_line_breaks` filter.

lfdebrux:
@katstevens good thinking, there seem to be a couple of places that look like good candidates.

lfdebrux:
Thinking about this, adding links to other areas seems like a product question, so I'll leave those for now.

PR #823, opened 2018-07-27
Refactor the save search template

lfdebrux:
This PR refactors the save search template, which was highlighted in a [tech debt ticket](https://trello.com/c/wlfBF5J0/51-buyer-fe-save-searchhtml-page-uses-blend-of-wtforms-raw-html-fe-toolkit-templates) from Sam as being a bit messy.

That assessment was I think borne out by how long this refactor took... I took this refactoring on because it was blocking my work to use WTForms more widely in the frontend apps (see alphagov/digitalmarketplace-frontend-toolkit#400).

I used WTForms to move most of the validation logic out of the view and into the `CreateProjectForm`, so overall the number of lines is fairly static; however with alphagov/digitalmarketplace-frontend-toolkit#400 the new emphasis on convenience methods for WTForms will mean a future reduction in lines.

When making changes I made sure that unit tests and functional tests remained green, and there should be as few visual regressions as possible, so some peculiarities remain in the design of this page:

  1. The validation banner that appears at the top of the page when there is a form error is different to almost every other banner on this site; instead of showing the question name it shows the validation error message.

  2. The search query is passed to the view both as a URL query string and (when POSTing) as a form field. As far as I could tell the `search_query` form field was never used in the view, but I did not remove it because doing so would mean that the tests would diverge from the code, and I didn't want to touch the tests.

  3. Previously the form validation did not check whether the POSTed project was present in the list presented to the user in the selection box. It turned out that all the unit tests had been exercising this "feature" because the global project fixture was not in the global project list fixture. This is the one area where I did touch the tests, because whilst this might seem like a feature it is also a clear bug.

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -1,7 +1,7 @@
 from flask_wtf import FlaskForm
 
 from wtforms.fields import DecimalField, RadioField, BooleanField
-from wtforms.validators import DataRequired, Length, NumberRange, Optional, InputRequired
+from wtforms.validators import DataRequired, Length, NumberRange, InputRequired, ValidationError
 
 from dmutils.forms.fields import DateField
 from dmutils.forms.validators import GreaterThan
@@ -12,14 +12,43 @@ from decimal import Decimal
 
 
 class CreateProjectForm(FlaskForm):
-    name = StripWhitespaceStringField(
-        'Name your search', id="project_name",
+    save_search_selection = RadioField(
         validators=[
-            Length(min=1,
-                   max=100,
-                   message="Names must be between 1 and 100 characters"),
-            Optional(),
-        ])
+            InputRequired("Please choose where to save your search")
+        ]
+    )
+    name = StripWhitespaceStringField(
+        "Name your search. A reference number or short description of what you want to buy makes a good name.",
+    )
+
+    def __init__(self, projects, **kwargs):
+        super().__init__(**kwargs)
+
+        self.save_search_selection.options = [{
+            "label": project["name"] or f"Untitled project {project['id']}",
+            "value": str(project["id"]),
+        } for project in projects]
+        self.save_search_selection.options.append({
+            "label": "Save a new search",
+            "value": "new_search",
+            "reveal": {
+                "question": self.name.label.text,
+                "hint": "100 characters maximum",
+                "name": self.name.name,
+            }
+        })
+
+        self.save_search_selection.choices = [
+            (option["value"], option["label"]) for option in self.save_search_selection.options
+        ]
+
+    def validate_name(form, field):
+        if form.save_search_selection.data == "new_search":
+            try:
+                Length(min=1, max=100, message="Names must be between 1 and 100 characters")(form, field)
+            except ValidationError as e:
+                form.save_search_selection.options[-1]["reveal"]["error"] = e.args[0]
+                raise
 
 
 # TODO: move this into dmutils.forms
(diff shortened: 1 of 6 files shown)


risicle:
Notionally from 30,000ft, a hundred thumbs up. Will look closer later.

app/templates/direct-award/save-search.html
```diff
@@ -30,21 +30,25 @@
 <div class="single-question-page save-search-page">
   <div class="grid-row">
     <div class="column-one-whole">
-        {% if form.name.errors or save_search_selection_error %}
+        {#
+          we cannot use the toolkit/forms/validation.html template here,
+          as this page was designed with a govuk style error summary banner
```
katstevens: As discussed, it's a shame that this version doesn't have the yellow focus border on the div for keyboard users, but given the difference in user experience (displaying message rather than question name) we should get some input from Content before we change it. But we should change it at some point 😸

app/main/forms/direct_award_forms.py
```diff
@@ -12,14 +12,43 @@
...
+            "label": project["name"] or f"Untitled project {project['id']}",
+            "value": str(project["id"]),
+        } for project in projects]
+        self.save_search_selection.options.append({
+            "label": "Save a new search",
+            "value": "new_search",
+            "reveal": {
+                "question": self.name.label.text,
+                "hint": "100 characters maximum",
+                "name": self.name.name,
```
katstevens: 🏆 🏆 🏆 for moving the hint text etc here

app/main/views/g_cloud.py
```diff
@@ -424,17 +424,48 @@ def save_search(framework_family):
...
+
+        except HTTPError as e:
+            abort(e.status_code)
 
+        flash(PROJECT_SAVED_MESSAGE, 'success')
+
+        return redirect(url_for('.view_project',
+                                framework_family=framework_family,
+                                project_id=project['id']
+                                ), code=303)
```
katstevens: I know this was here before, but why have a `303` status? 🤔

app/main/views/g_cloud.py
```diff
@@ -424,17 +424,48 @@ def save_search(framework_family):
...
+        try:
+            data_api_client.create_direct_award_project_search(user_id=current_user.id,
+                                                               user_email=current_user.email_address,
+                                                               project_id=project['id'],
+                                                               search_url=search_api_url)
+
+        except HTTPError as e:
+            abort(e.status_code)
 
+        flash(PROJECT_SAVED_MESSAGE, 'success')
```
katstevens: I don't think our test coverage currently includes asserting the success flash message 😿  Can we sanity check this is still being rendered at the right time/place?

app/main/views/g_cloud.py
lfdebrux: [git blame](https://github.com/alphagov/digitalmarketplace-buyer-frontend/commit/31fe0ed76093264e0ed4110cc301205440dcc75f) suggests that it is more appropriate for a POST redirect, it's a HTTP/1.1 feature.

I think it's fine either way, maybe something to look into more in future?

app/main/views/g_cloud.py
lfdebrux: Will do

app/main/forms/direct_award_forms.py
lfdebrux: Soon it will be on the `self.name` field object itself 😄

app/templates/direct-award/save-search.html
lfdebrux: Yes, the accessibility issues do concern me though...

app/main/views/g_cloud.py
lfdebrux: screen shot 2018-07-27 at 13 53 24

Flash message success!

app/main/views/g_cloud.py
katstevens: Fair enough. I was just curious as we have quite a few other POST redirects in that module that use the default (302, I think).

PR #838, opened 2018-10-03
159-sub-categories-something-seems-broken

benvand:
Do not strip `parentCategory` in `clean_request_args` unless args are to be used in aggregation
Use `for_aggregation` output of `clean_request_args` when building search query
Do not use `parentCategory` as a filter group.

When we call `clean_request_args` it strips arguments from the request args unless they are added to a list called `restore_keys`. This is a problem because we later rely on these to determine the `parentCategory` of the selected filter (below) and if `parentCategory` isn't in the args then default it to `category['value'] == category['value']` or `True`.
https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/main/presenters/search_presenters.py#L307

This PR 
* makes it possible to preserve the expected `parentCategory` arg by ensuring the `for_aggregation` parameter is `False` in the call in the view to `clean_request_args`.
* Adds `parentCategory` to a list of skipped arguments when building groups to filter results by (we only require the `serviceCategories`)
* Supplies the `for_aggregation` kwarg set to `True` when we are filtering the request arguments for `build_search_query` (ie for aggregation)

## Before

screen shot 2018-10-03 at 13 39 03

## After

screen shot 2018-10-03 at 13 39 23

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -103,6 +103,7 @@ def clean_request_args(request_args, lot_filters, lots_by_slug, for_aggregation=
 
     if not for_aggregation:
         restore_keys.append('page')
+        restore_keys.append('parentCategory')
 
     if request_args.get('lot') in lots_by_slug:
         restore_keys.append('lot')
--- a/tests/main/presenters/test_search_presenters.py
+++ b/tests/main/presenters/test_search_presenters.py
@@ -4,6 +4,7 @@ import os
 
 import flask
 import mock
+import pytest
 from werkzeug.datastructures import MultiDict
 from werkzeug.urls import Href
 
@@ -502,7 +503,7 @@ class TestLotsAndCategoriesSelection(BaseApplicationTest):
                                         {
                                             'id': 'checkboxTreeExample-option-2.2',
                                             'label': 'Option 2.2',
-                                            'link': '/g-cloud/search?parentCategory=option+2&lot=cloud-software&checkboxTreeExample=option+2.2',  # NOQA
+                                            'link': '/g-cloud/search?lot=cloud-software&checkboxTreeExample=option+2.2&parentCategory=option+2',  # NOQA
                                             'name': 'checkboxTreeExample',
                                             'selected': False,
                                             'service_count': 0,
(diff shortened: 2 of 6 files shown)


app/main/helpers/search_helpers.py
```diff
@@ -103,6 +103,7 @@ def clean_request_args(request_args, lot_filters, lots_by_slug, for_aggregation=
 
     if not for_aggregation:
         restore_keys.append('page')
+        restore_keys.append('parentCategory')
```
katstevens: Below this we validate whether the `lot` is a valid lot before restoring it. Should we be doing this for `parentCategory` as well? (How we'd do that, I have no idea 😹)

katstevens: (Also I'm sure you're looking at tests already, but we should definitely also update `test_clean_request_args` with `parentCategory`)

tests/main/presenters/test_search_presenters.py
```diff
@@ -680,3 +681,256 @@ def test_build_lots_and_categories_link_tree_with_lot_and_no_category_filters_ha
...
+                        'label': 'Option 7.1',
+                        'name': 'checkboxTreeExample',
+                        'id': 'checkboxTreeExample-option-7.1',
+                        'value': 'option 7.1'
+                    }
+                ]
+            }
+        ]
+        with self.app.test_request_context(url):
+                tree = build_lots_and_categories_link_tree(
```
katstevens: Curious that flake8 didn't pick up on this extra indenting 🤔

tests/main/presenters/test_search_presenters.py
```diff
@@ -680,3 +681,256 @@ def test_build_lots_and_categories_link_tree_with_lot_and_no_category_filters_ha
...
+                self.search_api_client
+            )
+
+        tree_child_value_selected_lists = [
+            [(child.get('id'), child['selected']) for child in node.get('children', [])] for node in tree
+        ]
+        tree_child_value_selected = [item for sublist in tree_child_value_selected_lists for item in sublist]
+
+        assert ('checkboxTreeExample-option-7.1', False) in tree_child_value_selected
+        assert tree_child_value_selected.count(('checkboxTreeExample-option-2.1', True)) == 2
```
katstevens: So if I understand this right, if there's no `parentCategory` and the child appears in two categories, we expand the categories for both of them?

benvand: Yeah that's right! It's weird but I think it's a) the intended behaviour and b) actually the right thing to do. Because without a `parentCategory` we can't find our way back up the tree.

We can't just put selected for a random choice, because then we might assign selected to something the user didn't select.

We shouldn't do it for none because they selected something.

The weird thing about this PR, and this bug, is that it just propagates the `parentCatgory` url argument to where it needs to be (and also makes sure it gets stripped where necessary). The code to 'select' the correct categories was already there.

Some validation of the `parentCategory` vs the selection could be used to find a way around this, but I'm not sure there's a user need for it 🤔

tests/main/presenters/test_search_presenters.py
benvand: Hmmmm, technically valid python I guess, but not something we want, changed 👍

tests/main/presenters/test_search_presenters.py
katstevens: Makes sense - thanks!

PR #839, opened 2018-10-09
Fix 159 sub categories something seems broken

benvand:
## The commits here are best to look at separately.

One backs out the original fix. The other adds a new version of the fix that works with the same tests.

The below shows the diff of master _before the previous breaking PR was merged_ up to here.

https://github.com/alphagov/digitalmarketplace-buyer-frontend/compare/d864e9260c9d837ac57fb39dbaae709c30b02907...be81d6ab2fc2e937799934cd55fa2362bd45c402

Looking at the above comparison we:

* Remove the `parentCategory` from the list of args used to generate the URL for the _root_ of the tree. The root has no parent...
* We pass `parentCategory` from the request args to `_annotate_categories_with_selection`.
* We use the passed `parentCategory` instead of the `cleaned_request_args` one. `cleaned_request_args` will never contain `parentCategory`. It's stripped in `clean_request_args`.
* In `_annotate_categories_with_selection` we only use `parentCategory` to generate the url of child branches of the tree

And all the tests still work.

### DIFF
--- a/app/main/helpers/search_helpers.py
+++ b/app/main/helpers/search_helpers.py
@@ -103,7 +103,6 @@ def clean_request_args(request_args, lot_filters, lots_by_slug, for_aggregation=
 
     if not for_aggregation:
         restore_keys.append('page')
-        restore_keys.append('parentCategory')
 
     if request_args.get('lot') in lots_by_slug:
         restore_keys.append('lot')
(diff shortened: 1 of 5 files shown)


benvand:
😀  ran it against func tests locally and I _think_ the only failures were expected

PR #852, opened 2019-01-24
direct award views: gracefully 404 when supplied with invalid framework families in path

risicle:
Spawned from https://trello.com/c/wT6iM4SQ this morning

Turns out there were a bunch of views that didn't consider this case. Also add tests as this was clearly untested.

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -70,6 +70,8 @@ def index_g_cloud():
 
 @direct_award_public.route("/<string:framework_family>/start", methods=("GET",))
 def pre_project_task_list(framework_family):
+    if framework_family != "g-cloud":
+        abort(404)
     frameworks_by_slug = framework_helpers.get_frameworks_by_slug(data_api_client)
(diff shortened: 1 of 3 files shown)


app/main/views/g_cloud.py
```diff
@@ -70,6 +70,8 @@ def index_g_cloud():
 
 @direct_award_public.route("/<string:framework_family>/start", methods=("GET",))
 def pre_project_task_list(framework_family):
+    if framework_family != "g-cloud":
+        abort(404)
```
katstevens: Can this view use `get_latest_live_framework_or_404` like the ones below?

risicle: I thought so initially, but realizing it was hard coded to `"g-cloud"` (like, 3 lines down) I thought it best to stick with that.

risicle:
Also squeezed a nix file fix in.

app/main/views/g_cloud.py
katstevens: Yeah, I guess we don't want DOS direct awards sneaking in at the start.

PR #853, opened 2019-01-25
Following framework content for Buyer FE

katstevens:
Trello: https://trello.com/c/JPSjopsY/337-create-g11-framework-on-preview-and-set-to-coming

screenshot-localhost-2019 01 25-17-48-08

Pulls in the new G11 content ready for use on the Buyer FE. I've tested the home page locally with G11 framework statuses of coming, open, pending, standstill and live.

Also includes a fix for the new metadata structure (used on DOS3 but not G10). This is needed in the Direct Award saved search projects to alert buyers that any searches containing G10 services they are evaluating will be suspended on a certain date, and that they might want to start a new search with G11 services instead.

There are two checks happening: whether the `following_framework` content is present, and whether there's an associated database object for that following framework. Flake8 decided this was 'too complex' and I agreed, so it's been moved out into a helper function.

### DIFF
--- a/app/main/helpers/framework_helpers.py
+++ b/app/main/helpers/framework_helpers.py
@@ -68,27 +68,14 @@ def get_framework_or_500(client, framework_slug, logger=None):
 
 def get_following_framework(client, current_framework, logger):
     """
-    Check for 1) a database entry for the following framework 2) the content metadata for the following framework
+    1) Fetch the 'following_framework' metadata for the current framework
+    2) Look up the 'following_framework' in the database
     Return None if either or both are missing.
     """
     try:
         following_framework_content = content_loader.get_metadata(
             current_framework['slug'], 'following_framework', 'framework'
         )
-        if following_framework_content:
-            try:
-                return client.get_framework(following_framework_content['slug'])['frameworks']
-            except HTTPError as e:
-                if e.status_code == 404:
-                    if logger:
-                        logger.info(
-                            "Framework not found: {error}, framework_slug: {framework_slug}",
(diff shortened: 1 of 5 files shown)


app/main/helpers/framework_helpers.py
```diff
@@ -63,3 +64,20 @@ def get_framework_or_500(client, framework_slug, logger=None):
...
+def get_following_framework_or_500(client, current_framework, logger):
+    try:
+        following_framework_content = content_loader.get_metadata(
+            current_framework['slug'], 'following_framework', 'framework'
+        )
+        if following_framework_content:
+            # We raise an error if the following_framework content refers to a framework not in the database
+            return get_framework_or_500(
+                client,
+                following_framework_content['slug'],
```
lfdebrux: I presume this means that putting G10 in a pre-live state will cause a 500? Do we care about that?

katstevens: Hmm that's a good point - we should fail silently instead in case production isn't ready to show the message.

lfdebrux: Hopefully we should never get to that state in production? Do we have functional tests that would catch a failure?

katstevens: Well, we only want to create the prod db entry for G11 when we're ready to show the 'coming' message. If we deploy the content update before then, the view will find the `following_framework` content and will be expecting the data to go with it.

lfdebrux: Ah I see it used to fail silently as well. Fine then 👍

PR #855, opened 2019-02-06
Bump itsdangerous && add pyup

jonodrew:


lfdebrux:
Could we have this in two commits please? Sorry to be a pain

PR #882, opened 2019-05-01
link to docker with harakiri and view testing it

agalamatis:
trello: https://trello.com/c/lEFJrblv/469-actual-flask-request-seems-to-have-unbounded-time-to-run

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,1 +1,1 @@
-FROM digitalmarketplace/base-frontend:4.4.3
+FROM digitalmarketplace/base-frontend:4.4.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,8 +15,8 @@ git+https://github.com/alphagov/digitalmarketplace-apiclient.git@19.15.1#egg=dig
 ## The following requirements were added by pip freeze:
 asn1crypto==0.24.0
 blinker==1.4
-boto3==1.9.134
-botocore==1.12.134
+boto3==1.9.140
+botocore==1.12.140
 certifi==2019.3.9
 cffi==1.12.3
 chardet==3.0.4


app/main/views/marketplace.py
```diff
@@ -83,6 +83,13 @@ def index():
     )
 
 
+@main.route('/_test_harakiri', methods=['GET'])
+def test_timeout():
+    from time import sleep
+    sleep(300)
+    abort(500)
```
katstevens: I'd probably put this in `app.status.views.py` rather than `marketplace.py`

app/main/views/marketplace.py
```diff
@@ -83,6 +83,13 @@ def index():
     )
 
 
+@main.route('/_test_harakiri', methods=['GET'])
+def test_timeout():
+    from time import sleep
```
katstevens: Can we put imports at the top? I can't remember if this explicitly fails `flake8` or not...

Dockerfile
```diff
@@ -1 +1 @@
-FROM digitalmarketplace/base-frontend:4.4.3
+FROM digitalmarketplace/base-frontend:4.4.5
```
katstevens: Missing line break?

katstevens:
Also I think the Snyk check is complaining about the `urllib3` dependency, I was looking at that on a 2nd line ticket but got sidetracked

app/main/views/marketplace.py
agalamatis: We can not unit test, maybe e2e test, but even then there is no way around having to wait more than 60 seconds (nginx default timeout and our harakiri value).

Dockerfile
agalamatis: it's not python, but I will add one

katstevens:
I can't remember - did you say we were going to remove the test view from this PR after all?

lfdebrux:
Please squash these commits before merging!

agalamatis:
I will squash the whole branch into one commit before merging.

lfdebrux:
> I will squash the whole branch into one commit before merging.

Great.

If you could have two commits; one for the docker base image bump, one for the requirements bumps, that would be even better ;)

PR #898, opened 2019-06-07
Include declaration documents on service page

katstevens:
https://trello.com/c/h3dtJytJ/882-display-modern-slavery-document-uploads-on-service-page

modern-slavery-doc-service-page

The Service presenter class already combines supplier data with service data. We're now adding in declaration data (where the Modern Slavery doc URL is kept).

The hardcoded document list is moved to the top of the `service_presenters` module for easier future editing.

The section of the manifest that previously displayed the documents at the bottom of the page has been removed, and a new `_service_documents.html` template has been included (with the same content as the top panel). 

Content changes of the document names need to be OK-ed with Stefan.

### DIFF
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -1,10 +1,23 @@
 import os
 import re
+from collections import OrderedDict
 from urllib.parse import unquote, urlparse
 
 from app import content_loader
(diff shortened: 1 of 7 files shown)


katstevens:
@benvand tests now added!

PR #901, opened 2019-06-10
Remove report a problem form

lfdebrux:
https://trello.com/c/z63XpU1l/548-remove-is-there-anything-wrong-with-this-page-form

### DIFF
--- a/tests/main/views/test_feedback.py
+++ b/tests/main/views/test_feedback.py
@@ -1,4 +1,3 @@
-from lxml import html
 import mock
 
 from ...helpers import BaseApplicationTest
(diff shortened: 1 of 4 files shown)


tests/main/views/test_feedback.py
```diff
@@ -30,9 +29,3 @@ def test_google_gone_logs_problem(self, current_app, external_requests_post):
     def test_feedback_submission(self, external_requests_post):
         external_requests_post.return_value.status_code = 200
         self._check_success(self._post())
-
-    def test_feedback_form_has_csrf(self):
```
katstevens: I wonder if we should return `404` at the `/feedback` view for now?

lfdebrux: What would be the best way to do that, taking out the view altogether?

katstevens: Strictly speaking it should be a `410 Gone`, as we're retiring the URL. We have a template for that already - let's keep the `send_feedback` view but replace all the code with `abort(410)`.

lfdebrux: 👍 great that sounds like a plan

lfdebrux: I've put together a commit to do this

PR #913, opened 2019-07-12
content loader: make thread-safe through use of context-local deep copies

risicle:
https://trello.com/c/RW2tzjbM

This implementation is a little more involved than the one in the supplier frontend, required because the buyer frontend needs information from the `application` object and the api before it can populate its content loader. Hence this is performed quite late in the startup process. This should all
be fine as long as no code needs to access the methods of the content loader before the `application` is initialized, though in such a case the attempt is made to fail obviously and loudly.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,6 +1,9 @@
+from copy import deepcopy
+
 from flask import Flask, request, redirect, session, abort
 from flask_login import LoginManager
 from flask_wtf.csrf import CSRFProtect, CSRFError
+from werkzeug.local import Local, LocalProxy
 
 import dmapiclient
 from dmutils import init_app
@@ -8,6 +11,7 @@ from dmcontent.content_loader import ContentLoader
 from dmcontent.utils import try_load_manifest, try_load_metadata, try_load_messages
 from dmutils.user import User
 from dmutils.external import external as external_blueprint
+from dmutils.timing import logged_duration
 
 from config import configs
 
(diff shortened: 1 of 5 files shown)


app/__init__.py
```diff
@@ -17,7 +21,46 @@
...
+            elif framework_data['framework'] == 'digital-outcomes-and-specialists':
+                master_cl.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
+                try_load_manifest(master_cl, application, framework_data, 'briefs', 'briefs_search_filters')
+
+    # seal master_cl in a closure by returning a function which will only ever return an independent copy of it.
+    # this is of course only guaranteed when the initial_instance argument wasn't used.
+    return lambda: deepcopy(master_cl)
+
+
+def _content_loader_factory():
```
lfdebrux: You can define this in get_content_loader which might make it clear that no one else should use it

app/__init__.py
```diff
@@ -33,18 +76,12 @@ def create_app(config_name):
...
-                # we need to be able to display old services, even on expired frameworks
-                content_loader.load_manifest(framework_data['slug'], 'services', 'display_service')
-                content_loader.load_manifest(framework_data['slug'], 'services', 'download_results')
-                try_load_metadata(content_loader, application, framework_data, ['following_framework'])
-            elif framework_data['framework'] == 'digital-outcomes-and-specialists':
-                content_loader.load_manifest(framework_data['slug'], 'briefs', 'display_brief')
-                try_load_manifest(content_loader, application, framework_data, 'briefs', 'briefs_search_filters')
+    # replace placeholder _content_loader_factory with properly initialized one
+    global _content_loader_factory
+    _content_loader_factory = _make_content_loader_factory(
```
lfdebrux: Monkey-patching?? :laughing: :see_no_evil:

risicle: Well I'm monkey-patching my own function so I don't know if that counts.

PR #935, opened 2019-10-16
43-create-a-base-page-for-the-apps

benvand:
https://trello.com/c/PpFz5qRc/43-create-a-base-page-for-the-apps

The work to move to govuk-frontend will be easier if we have a base page in each app in which to import macros.

GOVUK Frontend may solve this down the road
https://github.com/alphagov/govuk-frontend/issues/1278

iIt seems from research (Googling and GOV Pay) for now this is the accepted pattern

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -1,4 +1,4 @@
-{% extends "toolkit/layouts/_base_page.html" %}
+{% extends "_base_page.html" %}
 
 {% block page_title %}Cookies - Digital Marketplace{% endblock %}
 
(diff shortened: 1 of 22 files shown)


app/templates/content/cookies.html
```diff
@@ -86,6 +86,12 @@ <h3>Universal Analytics</h3>
                 </td>
                 <td>10 minutes</td>
             </tr>
+            <tr>
+                <td>_gat_govuk_shared</td>
+                <td>Used to manage the rate at which page view requests are made
+                </td>
+                <td>10 minutes</td>
```
katstevens: Was this meant to be included? The description copies the one from the row above...

benvand: Good catch, might have been a rebase thing, or a sneaky left over

PR #943, opened 2019-11-01
Change the default port for the search api

lfdebrux:
2nd line ticket: https://trello.com/c/cwjzIU8U/1212-stop-search-api-from-using-port-5001

See the PR on the search api that adds `DM_SEARCH_API_PORT` for more details: https://github.com/alphagov/digitalmarketplace-search-api/pull/241


.flaskenv
```diff
@@ -0,0 +1,8 @@
+DM_API_PORT=5000
+DM_SEARCH_API_PORT=5009
+DM_BUYER_PORT=5002
+
+FLASK_APP=application:application
+FLASK_ENV=development
+FLASK_RUN_EXTRA_FILES=app/content/frameworks/
+FLASK_RUN_PORT=${DM_BUYER_PORT}
```
katstevens: Should `.flaskenv` be included in `.dockerignore`?

lfdebrux: Good idea

lfdebrux: It shouldn't be a problem if it is included, but better safe than sorry

lfdebrux: Ah, I remember now, our `.dockerignore` files for our apps exclude everything then add only certain files; we don't want this file to be included in the Docker image, so we don't add it to the `.dockerignore` file.

katstevens: Ahh that's right. Well remembered!

PR #944, opened 2019-11-05
switch uses of old button toolkit template to digitalmarketplace-govuk-frontend

risicle:
https://trello.com/c/reSSXWv8/

Took a bit of tinkering to be able to see all of these uses, but everything I found looked ok.

### DIFF
--- a/app/templates/direct-award/end-search.html
+++ b/app/templates/direct-award/end-search.html
@@ -65,13 +65,18 @@
 
         {{ form.user_understands }}
 
-        <button class="button-save end-your-search-button"
-          data-analytics="trackEvent"
-          data-analytics-category="Direct Award"
-          data-analytics-action="End search"
-          data-analytics-label="{{ search_count }}"
-          {% if disable_end_search_btn %}disabled="disabled"{% endif %}
-          >Export results and continue</button>
+        {{ govukButton({
+          "text": "Export results and continue",
+          "value": "Export results and continue",
+          "classes": "end-your-search-button",
+          "disabled": disable_end_search_btn,
+          "attributes": {
+            "data-analytics": "trackEvent",
+            "data-analytics-category": "Direct Award",
+            "data-analytics-action": "End search",
+            "data-analytics-label": "{{ search_count }}",
+          },
+        }) }}
       </form>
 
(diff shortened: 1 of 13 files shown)


app/templates/direct-award/save-search.html
```diff
@@ -87,7 +87,11 @@ <h1>
 
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
         <input type="hidden" name="search_query" value="{{ search_query }}"/>
-        <input type="submit" class="button-save" name="return_to_overview" value="Save and continue">
```
aliuk2012: good catch

app/templates/search/_filters.html
```diff
@@ -13,5 +13,11 @@
...
     {% endfor %}
   </div>
-  <button class="button-save js-hidden js-dm-live-search" type="submit" aria-controls="search-summary-accessible-hint-wrapper">Filter</button>
+  {{ govukButton({
+    "text": "Filter",
+    "classes": "js-hidden js-dm-live-search",
+    "attributes": {
+      "aria-controls": "search-summary-accessible-hint-wrapper",
+    },
+  }) }}
```
katstevens: Functional tests seem happy with this 👍

risicle: Yeah I think for button clicking we mostly use a fairly universal selector.

app/templates/direct-award/end-search.html
```diff
@@ -65,13 +65,18 @@ <h1>
 
         {{ form.user_understands }}
 
-        <button class="button-save end-your-search-button"
```
aliuk2012: Another good catch

app/templates/search/_services_save_search.html
```diff
@@ -1,15 +1,21 @@
+{# this file can be used from contexts that don't inherit from _base_page.html, so we must ensure we have all imports we need #}
+{% from "components/button/macro.njk" import govukButton %}
```
aliuk2012: 😭, thanks for documenting this partial and explaining why an import was needed here.

risicle:
Looks like I missed at least one old-style button in this... https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/aba67473d3b41bd731c665cca39c194b80c684ae/app/templates/brief.html#L149

PR #945, opened 2019-11-06
help page: add information about the Public Procurement Review Service

risicle:
https://trello.com/c/xey48L9F

@pocketstefan what do you think about the positioning of this paragraph? Stuck it second to last...

Also I noticed that all the other paragraphs use the form "Email abc@example.com if..." - do you think we should try and fit in with that somehow? Should I make the email address visible instead of just a blind `mailto:` link?

image

### DIFF
--- a/app/templates/help/index.html
+++ b/app/templates/help/index.html
@@ -51,6 +51,10 @@
           </ul>
         </div>
 
+        <h2 class="heading-xmedium">Report a late payment, an opportunity or a procurement</h2>
+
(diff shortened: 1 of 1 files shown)


pocketstefan:
@risicle Looks good! yes I agree let's change it to 'Email publicprocurementreview@cabinetoffice.gov.uk if a buyer isn't...'

risicle:
Cool. Is it important that we then wouldn't be using the full name "Public Procurement Review Service" anywhere if we did that?

pocketstefan:
@risicle No, I think just the email is fine.

risicle:
Cool. Changed.

risicle:
Woohoo

PR #946, opened 2019-11-08
switch missed use of old button toolkit template to digitalmarketplace-govuk-frontend

risicle:
https://trello.com/c/reSSXWv8

Missed one. This one doesn't produce _wonderful_ layout when switched, but I'm certain this will get fixed when the content above it gets converted...

2019-11-08T10:15:01 880172Zc

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -141,13 +141,9 @@
   <div class="grid-row">
     <div class="column-two-thirds">
       <form action="{{ url_for('external.start_brief_response', brief_id=brief.id) }}" method="get">
-        {%
-          with
-          label="Apply for this opportunity",
-          type="save"
-        %}
-          {% include "toolkit/button.html" %}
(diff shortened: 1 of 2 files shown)


risicle:
I've added an extra `<br/>` for now which seems to fix it so it's more or less in a releasable state.

image

risicle:
Bah forgot the tests too...

app/templates/brief.html
```diff
@@ -138,16 +138,13 @@ <h1>{{ brief.title }}</h1>
     </div>
   </div>
 {% elif brief.status == 'live' %}
+  <br/><!-- <- remove once grid and body elements have been converted to govuk frontend -->
```
katstevens: The new styles will sort it out, yes. On my preview feature, adding `<p class="govuk-body">` for the secondary-action-link (above the button) did the trick.

PR #948, opened 2019-11-11
Adds user research signup to help page

aliuk2012:
Adding this as we will no longer have a dedicated user research banner.

Ticket: https://trello.com/c/X4xRKdwl

### DIFF
--- a/app/templates/help/index.html
+++ b/app/templates/help/index.html
@@ -52,12 +52,18 @@
         </div>
 
         <h2 class="heading-xmedium">Report a late payment, an opportunity or a procurement</h2>
-
+        
         <p>Email <a href="mailto:publicprocurementreview@cabinetoffice.gov.uk">publicprocurementreview@cabinetoffice.gov.uk</a> if a buyer isn’t following the correct processes, the rules of the framework, or hasn’t paid you.</p>
 
         <h2 class="heading-xmedium">Feedback about the Digital Marketplace</h2>
 
         <p>Email <a href="mailto:cloud_digital@crowncommercial.gov.uk">cloud_digital@crowncommercial.gov.uk</a> if you have feedback about your experience using the Digital Marketplace.</p>
+        
+        <p>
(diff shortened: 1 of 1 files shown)


app/templates/help/index.html
```diff
@@ -52,12 +52,18 @@ <h2 class="heading-xmedium">Guidance about buying and selling on the Digital Mar
...
-
+        
         <p>Email <a href="mailto:publicprocurementreview@cabinetoffice.gov.uk">publicprocurementreview@cabinetoffice.gov.uk</a> if a buyer isn’t following the correct processes, the rules of the framework, or hasn’t paid you.</p>
 
         <h2 class="heading-xmedium">Feedback about the Digital Marketplace</h2>
 
         <p>Email <a href="mailto:cloud_digital@crowncommercial.gov.uk">cloud_digital@crowncommercial.gov.uk</a> if you have feedback about your experience using the Digital Marketplace.</p>
+        
+        <p>
+          <a href="{{ url_for('external.user_research_consent') }}" data-analytics="trackEvent" data-analytics-category="user-research" data-analytics-action="View sign up to user research screen">
```
katstevens: The analytics JS is looking for the `.user-research-banner-close-btn` class on the link: https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/javascripts/user-research-consent-banner.js

katstevens: I think that's just for showing and hiding, but can we check if it's needed for the event tracking as well?

aliuk2012: I think the js is just setting cookies and show/hiding the banner. It will be removed when we roll out the phase banner and completely remove user research banner. This markup was taken from the existing user research banner and I included the tracking attributes to ensure we still tracking these clicks 
https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/user-research-consent-banner.html#L14-L20

PR #949, opened 2019-11-11
NB testing edits

NoraGDS:
Deleted extra spaces after the word 'expires'

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -27,7 +27,6 @@
         <p>
             The Digital Marketplace service puts small files (known as &lsquo;cookies&rsquo;) on to your computer.
         </p>
-
         <p>
             These cookies are used to:
         <ul class="list-bullet">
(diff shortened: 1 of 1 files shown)


katstevens:
I can't see much different on the layout locally - the columns don't align anyway 😹 

I'm happy for the trailing `&nbsp` to be removed, but I'd recommend including them in the actual values in the Expired columns. For example, changing to `10&nbsp;minutes` would mean '10' and 'minutes' are not on separate lines:

Before:
cookies-universal-analytics

After:
cookies-with-nbsp

aliuk2012:
Normally I'd suggest using css to make sure that content does not break over multiple lines but in this instance maybe we should just leave it in there and globally sort out `&nbsp;` 

Example of using css white-space attribute to stop the third column from wrapping its content like the other two columns do. https://codepen.io/aliuk12/pen/LYYBgoJ

PR #957, opened 2019-11-26
Use govuk-link, govuk-list and govuk-body styles from govuk-frontend

risicle:
https://trello.com/c/Oi60uTxu

Trying to tread the line between converting too much and not enough.

### DIFF
--- a/app/templates/brief.html
+++ b/app/templates/brief.html
@@ -104,8 +104,8 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <header class="page-heading-smaller">
-      <p class="context">{{ brief.organisation }}</p>
-      <h1>{{ brief.title }}</h1>
+      <span class="govuk-caption-l">{{ brief.organisation }}</span>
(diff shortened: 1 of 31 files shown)


app/templates/brief.html
```diff
@@ -104,8 +104,8 @@
 <div class="grid-row">
   <div class="column-two-thirds">
     <header class="page-heading-smaller">
-      <p class="context">{{ brief.organisation }}</p>
-      <h1>{{ brief.title }}</h1>
+      <p class="govuk-caption-l">{{ brief.organisation }}</p>
```
lfdebrux: This should be a span?

```suggestion
      <span class="govuk-caption-l">{{ brief.organisation }}</p>
```

risicle: Oh yes it should.

Thanks, fixed.

lfdebrux:
These changes require a small fix to the functional tests to pass, I'll do that now.

lfdebrux:
Functional tests are fixed, this is ready to merge

risicle:
Thanks

PR #959, opened 2019-12-03
Nora gds cookiechanges 1

NoraGDS:
Updated content. Added new content and table for probe cookie

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -84,7 +84,8 @@
             </tr>
             <tr>
                 <td>_gat_govuk_shared</td>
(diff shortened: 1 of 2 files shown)


katstevens:
Content looks good to me - there is a failing Python unit test though, which is looking for the old `<h1>Cookies</h1>` header. I'll add a commit to fix that. I can also 'squash' your first test commit at the same time if you like?

NoraGDS:
Thank you. Yes please.

katstevens:
Here's a screenshot of how the changes look on my machine:

cookies-page

I think the GOV.UK link needs to be an `<a href="...">` rather than a markdown link. The rest of it looks OK to me, but have a look and see what you think 😸

risicle:
(that's "Looks Good To Me")

PR #960, opened 2019-12-04
This replaces toolkit/page-headings with govuk frontend page titles

agalamatis:
trello: https://trello.com/c/zSFIq80i/

### DIFF
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -34,6 +34,7 @@ $path: "/static/images/";
 @import "toolkit/_grids.scss";
 @import "toolkit/_instruction-list.scss";
 @import "toolkit/_notification-banners.scss";
+@import "toolkit/_page-headings.scss";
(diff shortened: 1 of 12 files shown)


app/assets/scss/application.scss
```diff
@@ -34,7 +34,6 @@ $path: "/static/images/";
 @import "toolkit/_grids.scss";
 @import "toolkit/_instruction-list.scss";
 @import "toolkit/_notification-banners.scss";
-@import "toolkit/_page-headings.scss";
```
lfdebrux: Unfortunately there are a lot of places where `page-heading` class is used directly, so we probably can't remove this line in this PR (unless you feel like also tackling those bits!).

agalamatis: Not in the next 65 minutes :(

lfdebrux: :D no worries, another day then :D

PR #967, opened 2019-12-18
[WIP] Allow users to set analytics cookie preferences

katstevens:
https://trello.com/c/1aF2CmgI/186-5-align-cookie-settings-with-govuk

Aligns us with the current GOV.UK cookie policy, allowing users to opt out of Google Analytics cookies. 

- Adapts code from GOV.UK publishing components, GOVUK static and GOVUK frontend
- Display the GOV.UK style banner with two buttons, 'Accept cookies' and 'Cookie settings'
- Creates a new page with a form for users to set their preferences (using utils `dmutils.external` url when linking to it, ready for use in other repos)
- Sets new cookies `cookie_policy` and `cookie_preferences_set` to store the user's preferences when either clicking 'Accept Cookies' (to accept everything) or when submitting the preference form.
- Only initialise the Google Analytics code if the `cookie_policy` is set [TODO: fix failing tests]

TODO:
- figure out whether to put this in our FE toolkit code or DMP GOVUK Frontend (`cookie-settings.js` and `cookie-settings.scss` will stay in Buyer FE alongside the new page)


aliuk2012:
3ac93e8 fixes the issue we had with the home page overriding the previous set cookie_policy to the default policy.

PR #969, opened 2019-12-19
Bumps digitalmarketplace-govuk-frontend to 0.3.1

aliuk2012:
Ticket: https://trello.com/c/5aqSYf2P https://trello.com/c/9ofCV3e2/

PR #972, opened 2020-01-03
Replace main_content block with mainContent

lfdebrux:
Ticket: https://trello.com/c/d3sJhHq8/221-replace-pagetitle-maincontent-block-with-pagetitle-maincontent-in-buyer-frontend

Required for #969


risicle:
WFM

PR #975, opened 2020-01-10
Replacing GOV.UK Template with GOV.UK Frontend Template

aliuk2012:
Ticket: https://trello.com/c/as2lcdbr/263-3-remove-govuktemplate-from-buyer-frontend

#### Before

Digital Marketplace homepage, as on preview

#### After

Digital Marketplace homepage, as for this pull request


app/templates/_base_page.html
```diff
@@ -1,27 +1,65 @@
...
-    </main>
-  </div>
+  {% block errorSummary %}
+    {% if errors %}
+      {{ govukErrorSummary({
+      "titleText": "There is a problem",
+      "errorList": errors.values(),
+    }) }}
+    {% endif %}
+  {% endblock %}
```
lfdebrux: I don't think we've done error summaries for this app yet

lfdebrux:
I've picked up this PR so that @aliuk2012 can concentrate on cookie banner stuff.

This PR should be ready for review now.

---

@katstevens and I made the decision to keep the homepage layout as close to as-is for now, which means that I've copied the HTML from the GOV.UK Frontend header component so that we can have a phase tag in the same link as the service name. Unfortunately that is not possible using the macro in govuk-frontend.

risicle:
Soft :+1: - this _looks_ good on all pages I've looked at. Can't say anything about the analytics/cookie business...

lfdebrux:
> Soft 👍 - this _looks_ good on all pages I've looked at. Can't say anything about the analytics/cookie business...

Cool, thanks. There shouldn't be any analytics or cookies working in this PR.

katstevens:
This functional test scenario is failing for me: `features/smoke-tests/homepage.feature:4 # Scenario: User can see the main links on the homepage` (specifically the `And I see the 'G-Cloud supplier A–Z' link`)

lfdebrux:
> This functional test scenario is failing for me: `features/smoke-tests/homepage.feature:4 # Scenario: User can see the main links on the homepage` (specifically the `And I see the 'G-Cloud supplier A–Z' link`)

I think this might be a typo/bug in the footer; previously we were using an em-dash and now we are not...

lfdebrux:
Okay, I've updated the footer in alphagov/digitalmarketplace-govuk-frontend#55 and pulled that version into this app.

This means that the footer link that was causing the failing test now reads 'G-Cloud supplier A to Z'. The screenshot above has been updated.

This will still cause the functional tests to fail; if you want them to pass you either need to checkout alphagov/digitalmarketplace-functional-tests#727 locally, or wait until that PR has been merged.

PR #982, opened 2020-01-30
Fix styles for temporary message text

katstevens:
Ticket: https://trello.com/c/lbW5IQFx/

Similar fix to https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1163.

The text on sidebar temporary messages for G-Cloud saved searches (only visible when the current framework is about to expire) is black by default. I've added in the styling fix to the `_temporary-messages.scss` (next to the banner fix) for now until we upgrade the temporary messages component in Digital Marketplace GOVUK Frontend.

Before:
g-cloud-temporary-message-banner

After:
g-cloud-temporary-message-fix

### DIFF
--- a/app/assets/scss/overrides/_temporary-messages.scss
+++ b/app/assets/scss/overrides/_temporary-messages.scss
@@ -15,3 +15,17 @@
     }
   }
 }
+
+.temporary-message {
+  p.temporary-message-message {
+    color: $white;
+
+    a, a:visited {
+      color: $white;
+    }
+
+    a:hover, a:active {
+      color: $light-blue-50;
+    }
+  }
+}


aliuk2012:
Could I make a suggestion? Instead of applying a temporary fix why not try to replace the component with `digitalmarket-govuk-frontend` Alert component? 

The master branch already includes the version of `digitalmarket-govuk-frontend` 
that includes the Alert component. It might not be completely removing temporary messages component but at least its starting to apply the newer component without introducing temporary hacks which may take some time before they are removed.

lfdebrux:
@katstevens Thanks for the PR, good spot on the issue. I had forgotten that we had two components/classes  with the same styles 🤦‍♂ 

@aliuk2012 I think with these temporary message "cards" we still have more design work to do (see the [section on 'Framework application call to action' in the Rationalising banners RFC](https://docs.google.com/document/d/1LKEFU7rXABukVeo-t0jhURT022zyvusIV9x8VJlDnpQ/edit#heading=h.y9018ytdrlq3), and I'm not sure replacing with an alert is a good long term solution. I think the temporary fix is acceptable until we've resolved the design issue, and lets us concentrate on getting our changes to remove govuk_template out to production.

katstevens:
@lfdebrux @aliuk2012 Yes - we are probably going to be reviewing the content (and potentially placing) for this particular banner before G12 anyway.

lfdebrux:
Travis is being weirdly slow on this PR, going to restart the tests

lfdebrux:
> ~Travis is being weirdly slow on this PR, going to restart the tests~ Well that didn't work...
> 
> image

Yay it worked eventually

PR #985, opened 2020-02-05
changed grid and column classes to govuk equivalents

agalamatis:
trello: https://trello.com/c/7TceUAc5/222-2-update-grids-to-use-govuk-frontend-styles-in-buyer-frontend

### DIFF
--- a/app/templates/search/briefs.html
+++ b/app/templates/search/briefs.html
@@ -31,7 +31,7 @@
 
 {% block page_heading %}
 <div class="govuk-grid-row">
-  <div class="govuk-grid-olumn-two-thirds">
+  <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">{{ framework_family_name }} opportunities</h1>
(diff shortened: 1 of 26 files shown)


app/templates/search/briefs.html
```diff
@@ -30,8 +30,8 @@
 {% endblock breadcrumb %}
 
 {% block page_heading %}
-<div class="grid-row">
-  <div class="column-two-thirds">
+<div class="govuk-grid-row">
+  <div class="govuk-grid-olumn-two-thirds">
```
lfdebrux:
```suggestion
  <div class="govuk-grid-column-two-thirds">
```

lfdebrux:
This PR has the interesting side-effect of making the number of services found in a G-Cloud search larger

#### Before
image

#### After
image

I don't think this is a problem (it actually looks more consistent with DOS searches, which look this way but haven't changed), I just wanted to have it recorded somewhere!

PR #1005, opened 2020-03-12
query_args_for_pagination: handle multi-valued dicts by expanding values to lists

risicle:
Originating from https://trello.com/c/KaL0fCtR

This is the format which `url_for()` likes when it's being asked to construct a querystring which might have multi-valued keys. This will allow the paging links to retain their full filter selection between pages.


risicle:
Still TODO: make FTs cover this.

PR #1021, opened 2020-04-27
Review G-Cloud Services Journey Error Messages

domoscargin:
https://trello.com/c/4SyHm8wo/14-2-review-error-messages-for-search-for-g-cloud-services-journey-in-buyer-frontend

Updates the error messages in the G-Cloud services journey to bring them in line with the [Design System guidance](https://design-system.service.gov.uk/components/text-input/#error-messages). 

See the equivalent changes for the G12 supplier application error messages: https://github.com/alphagov/digitalmarketplace-frameworks/pull/606

And for DOS Opportunities:
https://github.com/alphagov/digitalmarketplace-frameworks/pull/617

### Notes
- Commits are broken down by page, with a further commit for changes decided with content design.
- In-place message fixes only - this PR doesn't address things like needing to feed back exactly what's missing from an incomplete date, or providing a dynamic character count on fields with character limits.

### DIFF
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -57,7 +57,7 @@ class DidYouAwardAContractForm(FlaskForm):
 
     did_you_award_a_contract = DMRadioField(
         "Did you award a contract?",
-        validators=[InputRequired(message="You need to answer this question.")],
+        validators=[InputRequired(message="Select yes if you awarded a contract")],
         options=[
             {'value': YES, 'label': 'Yes'},
             {'value': NO, 'label': 'No'},
--- a/app/main/forms/direct_award_forms.py
+++ b/app/main/forms/direct_award_forms.py
@@ -82,24 +82,21 @@ class WhichServiceWonTheContractForm(FlaskForm):
 
 
 class TellUsAboutContractForm(FlaskForm):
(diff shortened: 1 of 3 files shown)


domoscargin:
Added a new commit with content design suggestions

lfdebrux:
Small thing, but we should probably remove the "Please" from the `WhyDidYouNotAwardForm` message to keep it consistent

https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1021/files#diff-9c9505cd2e36e479f9c6a0d2bcca4c90R137

```suggestion
        validators=[InputRequired(message="Select a reason why you didn't award a contract")]
```

app/main/forms/direct_award_forms.py
```diff
@@ -57,7 +57,7 @@ class DidYouAwardAContractForm(FlaskForm):
 
     did_you_award_a_contract = DMRadioField(
         "Did you award a contract?",
-        validators=[InputRequired(message="You need to answer this question.")],
+        validators=[InputRequired(message="Select if you have awarded your contract")],
```
lfdebrux, line 60: I like this solution 👍

domoscargin:
Updated `WhyDidYouNotAwardForm` - good catch, @lfdebrux

PR #1026, opened 2020-04-30
Move G-Cloud search summary and button

domoscargin:
https://trello.com/c/jyjVLG6K/43-1-move-search-results-summary-and-save-your-search-button-on-g-cloud-search-page

This brings the search page more in line with the DOS search page.

### DIFF
--- a/app/templates/search/_services_save_search.html
+++ b/app/templates/search/_services_save_search.html
@@ -3,8 +3,6 @@
 <div id="js-dm-live-save-search-form">
   <form action="{{ url_for('direct_award.save_search', framework_family=framework_family) }}">
     <input type="hidden" name="search_query" value="{{ search_query_url }}">
-    <div class="button-save-wrapper govuk-grid-row">
-      <div class="save-button govuk-grid-column-one-quarter">
         {{ govukButton({
           "text": "Save your search",
           "attributes": {
@@ -16,7 +14,5 @@
             "data-analytics-label": search_count|string(),
           },
         }) }}
-      </div>
-    </div>
   </form>
 </div>
--- a/app/assets/scss/_search-page.scss
+++ b/app/assets/scss/_search-page.scss
@@ -29,26 +29,6 @@
   }
 }
 
-.button-save-wrapper {
-  margin: 0px 0px 20px;
-
-  .save-summary-text {
-    @include core-16;
-    padding: 7px 15px;
-    @include media(mobile) {
-      margin-bottom: 10px;
-    }
-  }
-
-  .save-button {
-    padding: 0;
-
-    .button-save {
-      @include core-19;
-    }
-  }
-}
-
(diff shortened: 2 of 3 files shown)


app/templates/search/_services_save_search.html
```diff
@@ -4,7 +4,7 @@
   <form action="{{ url_for('direct_award.save_search', framework_family=framework_family) }}">
     <input type="hidden" name="search_query" value="{{ search_query_url }}">
     <div class="button-save-wrapper govuk-grid-row">
-      <div class="save-button govuk-grid-column-one-quarter">
+      <div class="save-button govuk-grid-column-full">
```
lfdebrux, line 7: Do we need both these wrapper divs and their classes? Can we replace them with one div and some govuk-frontend overrides?

domoscargin: Good thinking! Doing a little digging, there's quite a bit of unnecessary CSS we can jettison as well:

`.button-save` is the old button class, made redundant by using the new govuk button. 
`.save-summary-text` is no longer used anywhere

The only behaviour we need on the button is to align it to the right. I had a look, and there doesn't seem to be a built in govuk class for that, so I've kept the wrapper to apply `text-align: right;`.

But it'd be even neater to drop the wrapper as well. There are some examples on govuk frontend of using `position: absolute` to align right, but that's more CSS, and it needs a `position: relative` parent anyway, so I _think_ this is the neatest way.

lfdebrux: Ooh much nicer!

One other suggestion, I've heard that it's good practice to put `app-` in front of CSS classes that are specific to the app rather than a toolkit (I think it's GOV.UK style, not sure where it's documented though).

Anyway we've been trying to add `app-` to CCS styles that we want to keep and leaving it off old styles, so we know what we're happy with.

TL;DR I think we can rename this class to `app-button-save-wrapper`

lfdebrux: Oh and for extra pedantry we could try naming it using [BEM style](http://getbem.com/naming/), which is the methodology the Design System uses.

domoscargin: I've gone with `app-button--right` since I think BEM style modifiers should indicate what they do.

app/templates/search/_services_save_search.html
```diff
@@ -3,8 +3,7 @@
 <div id="js-dm-live-save-search-form">
   <form action="{{ url_for('direct_award.save_search', framework_family=framework_family) }}">
     <input type="hidden" name="search_query" value="{{ search_query_url }}">
-    <div class="button-save-wrapper govuk-grid-row">
-      <div class="save-button govuk-grid-column-one-quarter">
+    <div class=".app-button--right">
```
lfdebrux, line 6:
```suggestion
    <div class="app-button--right">
```

Issue #1014, opened 2020-04-14
"Questions asked by suppliers" wraps awkwardly on Safari when it contains long links


tvararu:
See https://www.digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities/12107

This is how it looks:

![Screenshot 2020-04-14 at 13 33 28](https://user-images.githubusercontent.com/1650875/79225456-a0e5f880-7e54-11ea-8aa3-02e93a942d12.png)

I think adding `word-break: break-all;` to anchor tags inside summary tables could be enough to fix this:

![Screenshot 2020-04-14 at 13 35 59](https://user-images.githubusercontent.com/1650875/79225593-e276a380-7e54-11ea-8ca2-e5966ea35a10.png)

Let me know if you'd like a PR with that!

katstevens:
Thanks for raising this issue @tvararu. We'll add it to our backlog.

PR #1032, opened 2020-05-11
Adds app-break-link class

domoscargin:
https://trello.com/c/bcU2s0Zs/33-2-links-in-summary-tables-eg-dos-opportunities-dont-wrap-in-safari-or-chrome

The `.break-link` class was previously deleted, resulting in long URLs not getting wrapped.

* Adds `.app-break-link` class styles
* Updates dm-utils version, which adds the class name


lfdebrux:
Can we rename this to `.app-break-link`?

domoscargin:
> Can we rename this to `.app-break-link`?

We can, but would need to release [this change](https://github.com/alphagov/digitalmarketplace-utils/pull/561) first.

lfdebrux:
This fixes #1014

lfdebrux:
I've tested this locally and it does the job!

PR #1036, opened 2020-05-19
Update service meta block template

domoscargin:
1. I've made the "service documents" h2 visible - I think it looks better and it's useful to sighted and non-sighted users.
2. I've adjusted the service ID section to try and make it work better with screenreaders. I've tested with ChromeVox and VoiceOver and both seem to read the service ID more like a series of digits now, rather than 4 sets of numbers.


lfdebrux:
These look good to me 👍

PR #1038, opened 2020-05-20
Replace validation banner with GOV.UK Frontend error summary component

agalamatis:
trello: https://trello.com/c/u5CoMrCP/19-1-replace-validation-banner-with-govuk-frontend-error-summary-component-in-the-buyer-frontend

There are two templates in `direct-award` (`which-service-won-contract` and `why-didnt-you-award-contract`) that don't have unit tests that verify that the errors context variable is rendered within a specific DOM container.

### DIFF
--- a/app/templates/direct-award/end-search.html
+++ b/app/templates/direct-award/end-search.html
@@ -31,16 +31,6 @@
 {% endblock breadcrumb %}
 
 {% block mainContent %}
-{% if errors %}
-  <div class="validation-masthead" aria-labelledby="validation-masthead-heading" aria-role="group" tabindex="-1">
-    <p>
-      You need to <a href="#{{ form.user_understands.id }}">confirm that you’ve finished editing your search</a> before you
-      export your results.
-    </p>
-  </div>
-{% endif %}
-
-
 <div class="end-search-page">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds dmspeak">
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -786,8 +786,9 @@ class TestDirectAwardAwardContract(TestDirectAwardBase):
 
         doc = html.fromstring(res.get_data(as_text=True))
         assert len(doc.xpath('//legend[contains(normalize-space(), "Select if you have awarded your contract")]')) == 1
-        assert doc.xpath('boolean(//div[@class="validation-masthead"])')
-        assert doc.xpath('count(//div[@class="validation-masthead"]/a[@class="validation-masthead-link"])') == 1
+        errors = doc.cssselect('div.govuk-error-summary a')
+        assert len(errors) == 1
+        assert errors[0].text_content() == 'Select if you have awarded your contract'
 
     @pytest.mark.parametrize('choice, expected_redirect',
                              (('still-assessing', '/buyers/direct-award/g-cloud/projects/1'),
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -158,10 +158,13 @@ class TestDirectAward(TestDirectAwardBase):
 
     def _asserts_for_create_project_failure(self, res):
         assert res.status_code == 400
-        html = res.get_data(as_text=True)
-        assert self.SEARCH_API_URL not in html  # it was once, so let's check
-        assert self.SIMPLE_SEARCH_PARAMS in html
-        assert "Search name must be between 1 and 100 characters" in html
+        doc = res.get_data(as_text=True)
+        assert self.SEARCH_API_URL not in doc  # it was once, so let's check
+        assert self.SIMPLE_SEARCH_PARAMS in doc
+        errors = html.fromstring(doc)
+        assert len(errors.cssselect("div.govuk-error-summary a")) == 1
+        assert errors.cssselect("div.govuk-error-summary a")[0].text_content() == \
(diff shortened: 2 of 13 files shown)


lfdebrux:
On the page `/buyers/direct-award/g-cloud/projects/<>/end-search` I got two error summaries

screencapture-localhost-buyers-direct-award-g-cloud-projects-587350007381662-end-search-2020-05-20-17_08_11

lfdebrux:
I also get two error summaries on the save search page. Again I think we should keep the top one.

screencapture-localhost-buyers-direct-award-g-cloud-save-search-2020-05-20-17_12_05

lfdebrux:
`which-service-won-contract` and `why-didnt-you-award-contract` look good 👍

lfdebrux:
Great, all the pages look good to me now.

If you want we can pair on adding new tests for the two pages that are missing them, or we can just merge this PR as is; up to you.

domoscargin:
On radio selects like /buyers/direct-award/g-cloud/choose-lot is there a reason for targeting the first option instead of the error message span like [GOV.UK](https://self-referral.test-for-coronavirus.service.gov.uk/test-type)?

Screenshot 2020-06-11 at 10 47 10

lfdebrux:
> On radio selects like /buyers/direct-award/g-cloud/choose-lot is there a reason for targeting the first option instead of the error message span like [GOV.UK](https://self-referral.test-for-coronavirus.service.gov.uk/test-type)?
> 

Interesting. I was going off this comment in govuk-frontend: `The first id should not have a number suffix so it's easy to link to from the error summary component`.

https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/radios/template.njk#L55

It also states in the [GOV.UK Design System page error summary guidance](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer) that

> For questions that require a user to select one or more options from a list using radios or checkboxes, link to the first radio or checkbox.

I know that GOV.UK sometimes don't always follow Design System guidance though, is there some dispute about this guidance?

lfdebrux:
> Similarly, the GOV.UK example wraps the "or" option with the error margin styling as well.

Yeah, our "or" is totally hacked in, when we replace that component it should fix itself.

domoscargin:
> > On radio selects like /buyers/direct-award/g-cloud/choose-lot is there a reason for targeting the first option instead of the error message span like [GOV.UK](https://self-referral.test-for-coronavirus.service.gov.uk/test-type)?
> 
> Interesting. I was going off this comment in govuk-frontend: `The first id should not have a number suffix so it's easy to link to from the error summary component`.
> 
> https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/radios/template.njk#L55
> 
> It also states in the [GOV.UK Design System page error summary guidance](https://design-system.service.gov.uk/components/error-summary/#linking-from-the-error-summary-to-each-answer) that
> 
> > For questions that require a user to select one or more options from a list using radios or checkboxes, link to the first radio or checkbox.
> 
> I know that GOV.UK sometimes don't always follow Design System guidance though, is there some dispute about this guidance?

Somehow I forgot that there was more content on the Design System page and went looking for practical examples! This implementation looks spot on in that regard, my bad!

lfdebrux:
> Somehow I forgot that there was more content on the Design System page and went looking for practical examples! This implementation looks spot on in that regard, my bad!

No worries 😁 It's good to see someone reviewing it properly! 😆

domoscargin:
Similarly, I guess this span spacing will have to wait until the component gets replaced?

Screenshot 2020-06-11 at 11 10 44

lfdebrux:
> Similarly, I guess this span spacing will have to wait until the component gets replaced?
> 

Yeesh that's messy HTML. Yes, sorry for prolonging your pain 😝

tests/main/views/test_direct_award.py
```diff
@@ -786,8 +786,9 @@ def test_award_contract_error_if_no_input(self):
 
         doc = html.fromstring(res.get_data(as_text=True))
         assert len(doc.xpath('//legend[contains(normalize-space(), "Select if you have awarded your contract")]')) == 1
-        assert doc.xpath('boolean(//div[@class="validation-masthead"])')
-        assert doc.xpath('count(//div[@class="validation-masthead"]/a[@class="validation-masthead-link"])') == 1
+        errors = doc.cssselect('div.govuk-error-summary a')
+        assert len(errors) == 1
+        assert errors[0].text_content() == 'Select if you have awarded your contract'
```
katstevens, line 794: Is this assertion looking at the same thing as line 788?

app/templates/direct-award/end-search.html
```diff
@@ -31,16 +31,6 @@
 {% endblock breadcrumb %}
 
 {% block mainContent %}
-{% if errors %}
-  <div class="validation-masthead" aria-labelledby="validation-masthead-heading" aria-role="group" tabindex="-1">
-    <p>
-      You need to <a href="#{{ form.user_understands.id }}">confirm that you’ve finished editing your search</a> before you
-      export your results.
```
katstevens, line 38: Good call on getting rid of this in favour of the shorter WTForm message.

tests/main/views/test_direct_award.py
lfdebrux: No, the `legend` element in line 788 is the label for the radio input element, which includes the error message next to the form inputs, whereas line 791 is looking at the error message in the error summary.

PR #1041, opened 2020-05-29
Replace summary table with summary list on G-Cloud Service and DOS Opportunity pages

domoscargin:
https://trello.com/c/CXGExCLO/55-3-replace-summary-table-with-summary-list-on-g-cloud-service-page-and-dos-opportunity-page

This update replaces `toolkit/summary-table` with the [GOV.UK Design System Summary List](https://design-system.service.gov.uk/components/summary-list/) on G-Cloud service pages (/g-cloud/services/<service_id>) and DOS Opportunity pages (/digital-outcomes-and-specialists/opportunities/<brief_id>).

### DIFF
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -4,6 +4,7 @@ from collections import OrderedDict
 from urllib.parse import unquote, urlparse
 
 from dmcontent.formats import format_service_price
+from dmcontent.html import to_summary_list_rows
 
 DECLARATION_DOCUMENT_KEYS = [
     ('modernSlaveryStatement', 'modernSlaveryStatementURL'),
@@ -29,6 +30,9 @@ class Service(object):
 
     def __init__(self, service_data, manifest, lots_by_slug, declaration=None):
         self.summary_manifest = manifest.summary(service_data)
+        # get attributes in format suitable for govukSummaryList
+        for section in self.summary_manifest:
+            section.summary_list = to_summary_list_rows(section.questions)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
         self.serviceSummary = service_data.get('serviceSummary', service_data.get('serviceDescription'))
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -32,7 +32,7 @@ class Service(object):
         self.summary_manifest = manifest.summary(service_data)
         # get attributes in format suitable for govukSummaryList
         for section in self.summary_manifest:
-            section.summary_list = to_summary_list_rows(section.questions)
+            section.summary_list = to_summary_list_rows(section.questions, preserve_line_breaks=True)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
         self.serviceSummary = service_data.get('serviceSummary', service_data.get('serviceDescription'))
--- a/app/main/presenters/service_presenters.py
+++ b/app/main/presenters/service_presenters.py
@@ -32,7 +32,7 @@ class Service(object):
         self.summary_manifest = manifest.summary(service_data)
         # get attributes in format suitable for govukSummaryList
         for section in self.summary_manifest:
-            section.summary_list = to_summary_list_rows(section.questions, preserve_line_breaks=True)
+            section.summary_list = to_summary_list_rows(section.questions, capitalize_first=True)
         # required attributes directly mapped to service_data values
         self.title = service_data['serviceName']
         self.serviceSummary = service_data.get('serviceSummary', service_data.get('serviceDescription'))
--- a/app/templates/_brief_attributes.html
+++ b/app/templates/_brief_attributes.html
@@ -1,44 +1,38 @@
-{% import "toolkit/summary-table.html" as summary %}
-{% call(item) summary.list_table(
-  [
-    {"label": "Published", "value": brief.publishedAt|dateformat},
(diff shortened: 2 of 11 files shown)


app/templates/_brief_q_and_a.html
```diff
@@ -11,25 +11,11 @@
...
-      {{ question.question | format_links | preserve_line_breaks }}
-    {%- endcall %}
-    {{ summary.text(question.answer | format_links | preserve_line_breaks) }}
-  {% endcall %}
-{% endcall %}
+<h2 id="clarification-questions" class="govuk-!-font-size-27 govuk-!-padding-top-6">Questions asked by suppliers</h2>
+{% if brief.clarificationQuestions is undefined or brief.clarificationQuestions|length == 0 %} 
+<p class="summary-item-no-content">No questions have been answered yet</p>
+{% endif %}
+{{ govukSummaryList({"rows": brief.clarificationQuestions}) }}
```
lfdebrux, line 18: Is there something we should be doing to make sure that the summary list and the heading are connected in HTML? Like a legend?

domoscargin: I don't think so. There's nothing like that in [this example](https://design-system.service.gov.uk/patterns/check-answers/), for instance. 

We could potentially slap an `aria-labelledby` or similar onto the summary list, but I think that would just be repeating the `<h2>` info.

lfdebrux: Okay no worries then

app/main/presenters/service_presenters.py
```diff
@@ -74,6 +81,18 @@ def get_service_id(self, service_data):
         else:
             return list(chunk_string(str(id), 4))
 
+    def get_external_framework_url(self, service_data):
```
katstevens, line 84: I thought we'd got rid of this?

katstevens: I think something's gone wrong with a rebase somewhere - the deleted stuff from https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1040 has crept back in

domoscargin: Whoops - should be sorted now

app/main/presenters/service_presenters.py
```diff
@@ -29,6 +30,9 @@ def _add_as_attribute_if_key_exists(self, key, service_data):
 
     def __init__(self, service_data, manifest, lots_by_slug, declaration=None):
         self.summary_manifest = manifest.summary(service_data)
+        # get attributes in format suitable for govukSummaryList
+        for section in self.summary_manifest:
+            section.summary_list = to_summary_list_rows(section.questions)
```
katstevens, line 35: I think we want to include `preserve_line_breaks` here (but not `format_links`)?

domoscargin: Good spot!

app/templates/_service_attributes.html
```diff
@@ -1,28 +1,18 @@
...
-        {% if not question.is_empty %}
-          {% call summary.row() %}
-            {{ summary.field_name(question.label) }}
-            {% with item_value = summary[question.type](question.filter_value|capitalize_first, question.assurance) %}
-              {{ (item_value | preserve_line_breaks) if question.type == 'textbox_large' else item_value }}
-            {% endwith %}
-          {% endcall %}
-        {% endif %}
-      {% endcall %}
+      <h2 id="{{ section.slug }}" class="govuk-heading-m govuk-!-font-size-27">{{ section.name }}</h2>
```
lfdebrux, line 10: Is the font size override here because that's what it used to look like, or because it looks better this way?

domoscargin: The former initially, but I found with all the bold questions the extra size helps the heading stand out a bit.

domoscargin:
I've pushed a change to de-bold the summary list keys - I think this helps with
1. Not overwhelming the user with emphasized copy (The Design System recommends to "[use bold sparingly](https://design-system.service.gov.uk/styles/typography/)"
2. Removing a font-size class from the headings which I included to help them stand out.

lfdebrux:
> I've pushed a change to de-bold the summary list keys - I think this helps with
> 
> 1. Not overwhelming the user with emphasized copy (The Design System recommends to "[use bold sparingly](https://design-system.service.gov.uk/styles/typography/)"
> 2. Removing a font-size class from the headings which I included to help them stand out.
> 
> I've left the text bold on mobile displays because then it has the opposite effect: it makes the text _easier_ to parse!

Is it worth suggesting to the Design System that they consider making this change standard?

app/templates/_brief_attributes.html
```diff
@@ -1,44 +1,38 @@
-{% import "toolkit/summary-table.html" as summary %}
-{% call(item) summary.list_table(
-  [
-    {"label": "Published", "value": brief.publishedAt|dateformat},
-    {"label": "Deadline for asking questions", "value": brief.clarificationQuestionsClosedAt|utcdatetimeformat},
-    {"label": "Closing date for applications", "value": brief.applicationsClosedAt|utcdatetimeformat}
+{% from "govuk-frontend/components/summary-list/macro.njk" import govukSummaryList %}
+
+<h2 class="govuk-visually-hidden">Important dates</h2>
```
lfdebrux, line 3: I'm tempted to suggest making this visible, what do you think?

lfdebrux:
Another fix needed for dmcontent, sorry, so that when the user selects only one option from a set of checkboxes that answer is displayed as text rather than as a list

image

https://github.com/alphagov/digitalmarketplace-content-loader/pull/82

PR #1044, opened 2020-06-02
Replace statistics component with GOV.UK FE styles

domoscargin:
https://trello.com/c/ECAfNadJ/94-1-remove-statistics-card-component-from-buyer-frontend

This replaces the [statistics component](https://github.com/alphagov/digitalmarketplace-frontend-toolkit/blob/master/toolkit/templates/statistics.html) with plain HTML and GOV.UK FE classes.

The Summary List's first column width is set to 30%, whereas standard GOV.UK spacing is 33.3333%. This means that when we [switch to the Summary List](https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1041) on this page, the alignment will be slightly out, hence the added SCSS to adjust.

However, if we do that, those same styles will be needed on the Briefs FE preview DOS opportunity page as well, so suggestions welcome.

### DIFF
--- a/app/assets/scss/_brief_page.scss
+++ b/app/assets/scss/_brief_page.scss
@@ -1,0 +1,5 @@
+.app-stat-block {
+    @include govuk-media-query($from: tablet) {
+        width: 30%;
+    }
+}
--- a/app/templates/_brief_applications_stats.html
+++ b/app/templates/_brief_applications_stats.html
@@ -1,22 +1,49 @@
-{% import "toolkit/statistics.html" as statistics %}
+<div class="govuk-grid-column-one-third app-stat-block" id="incomplete-applications">
+   <h2>
+      <span class="govuk-!-font-size-36 govuk-!-display-block">{{- brief_responses_stats.incomplete_responses_total }}</span>
+      <span>
+         {{- 
+            pluralize(
+               brief_responses_stats.incomplete_responses_total,
+               "Incomplete application",
+               "Incomplete applications"
+            ) 
+         -}}
+      </span>
+   </h2>
+   <p class="govuk-body govuk-!-font-size-16">
+      {%- if brief_responses_stats.incomplete_responses_total %}
+         {{- 
+            "%d SME, %d large" | format(
+               brief_responses_stats.incomplete_sme_responses,
+               brief_responses_stats.incomplete_large_responses
+            ) 
+         -}}
+      {% endif -%}
+   </p>
+</div>
+<div class="govuk-grid-column-one-third app-stat-block" id="completed-applications">
+   <h2>
+      <span class="govuk-!-font-size-36 govuk-!-display-block">{{- brief_responses_stats.completed_responses_total }}</span>
+      <span>
+         {{- 
+            pluralize(
+               brief_responses_stats.completed_responses_total,
+               "Completed application",
+               "Completed applications"
+            ) 
+         -}}
+      </span>
+   </h2>
+   <p class="govuk-body govuk-!-font-size-16">
+      {%- if brief_responses_stats.completed_responses_total %}
+         {{- 
+            "%d SME, %d large" | format(
+               brief_responses_stats.completed_sme_responses,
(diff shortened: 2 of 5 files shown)


katstevens:
To be honest, the alignment is already off for these on production, thanks to the padding 😹  I don't think it's a huge issue.

katstevens:
What do people think about putting new custom Sass that's intended for use across apps in digitalmarketplace-govuk-frontend? We have an `_all.scss` file to pull in the component styles - would it be awful to include an extra `_dm_app.scss` to keep our random bits and bobs in?

domoscargin:
> What do people think about putting new custom Sass that's intended for use across apps in digitalmarketplace-govuk-frontend? We have an `_all.scss` file to pull in the component styles - would it be awful to include an extra `_dm_app.scss` to keep our random bits and bobs in?

That was my instinct - it's just about the threshold of when things should be centralised, I guess - used in two apps?

domoscargin:
Though testing this change with the summary list changes, it seems to be very slightly out of whack anyway, and getting it to align exactly might be more fiddly than it's worth.

app/assets/scss/_brief_page.scss
```diff
@@ -0,0 +1,7 @@
+#incomplete-applications, #completed-applications {
+    width: 30%;
+
+    @include govuk-media-query($until: tablet) {
```
lfdebrux, line 4: Nice 👍 

This is a question more for my own curiosity than anything else, but what is the difference between using `$until: tablet` like you've done here and `$from: tablet` like the summary list does?

app/assets/scss/_brief_page.scss
```diff
@@ -0,0 +1,7 @@
+#incomplete-applications, #completed-applications {
```
lfdebrux, line 1: Is there reason why we can't use a class instead of ids?

lfdebrux:
My thoughts, following on from the discussion we had about this earlier:

1. I've left a [comment about the spacing of the summary list column in the GOV.UK Design System backlog](https://github.com/alphagov/govuk-design-system-backlog/issues/182#issuecomment-638182454). The fact that the column and the grid system don't allow feel like a bug to me.

2. I think that adding stuff to digitalmarketplace-govuk-frontend should be done as thoughtfully as possible, because it's a lot easier to add/remove stuff to an app that it is to a library. In this case, where we going to want to use this code in a maximum of two places, I think it might be more effort than it's worth.

3. I could see a strong argument for including this Sass in digitalmarketplace-govuk-frontend if we wanted to turn it into something like a [card component](https://github.com/alphagov/govuk-design-system-backlog/issues/113). We might have use for such a component in various places in our service, but that's a bit speculative.

app/assets/scss/_brief_page.scss
domoscargin: A class is preferable - the IDs were there as part of the stat block. It doesn't _look_ like those IDs are getting used by anything, but I kept them in just in case.

I've added a class - I could refactor and jettison the IDs if they're not necessary.

app/assets/scss/_brief_page.scss
domoscargin: My brain is mush today!

$from: tablet is better in this case, as the width is 100% by default anyway.

And generally $from would be better because that would usually indicate a mobile-first design.

app/templates/_brief_applications_stats.html
```diff
@@ -1,22 +1,49 @@
-{% import "toolkit/statistics.html" as statistics %}
+<div class="govuk-grid-column-one-third" class="app-stat-block" id="incomplete-applications">
```
lfdebrux, line 1: Chrome is ignoring the second class for me

```suggestion
<div class="govuk-grid-column-one-third app-stat-block" id="incomplete-applications">
```

domoscargin: Ugh, see earlier comment re: Brain being mush the day I worked on this. Thanks for the spot!

PR #1048, opened 2020-06-09
QA changes for summary list update

domoscargin:
* Adjust the margin on the Question and Answer session link on Brief pages.

BEFORE
Screenshot 2020-06-09 at 12 05 26

AFTER
Screenshot 2020-06-09 at 12 05 18

* Adds a "not empty" check to Brief pages to match Service pages and not display a heading for an empty section.


lfdebrux:
Looks like we need to put the question and answers in the HTML key of the summary list, I've found an example of a Q&A where the question includes HTML.

&#34;If we have more than 1 profile, which may be a good fit for the role, the:<br>1. Can we propose multiple profiles<br>2. If yes, then how?&#34;

http://localhost/digital-outcomes-and-specialists/opportunities/12024

PR #1050, opened 2020-06-16
display correct framework expiry date

gidsg:
https://trello.com/c/6mqoHNDb/125-1-expired-services-showing-wrong-framework-expiration-date-half-way-done

This fixes the unavailable banner to use the correct date from `frameworkExpiresAtUTC` field.
Before:
expired-service-date
After:
Screenshot 2020-06-16 at 16 25 04

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -206,6 +206,7 @@ def get_service_by_id(service_id):
             abort(e.status_code)
 
         service_unavailability_information = None
+        framework_expires_at_date = dateformat(framework['frameworkExpiresAtUTC'])
         status_code = 200
         if service['serviceMadeUnavailableAuditEvent'] is not None:
             service_unavailability_information = {
@@ -221,6 +222,7 @@ def get_service_by_id(service_id):
             service_unavailability_information=service_unavailability_information,
             lot=service_view_data.lot,
             gcloud_framework_description=framework_helpers.get_framework_description(data_api_client, 'g-cloud'),
+            framework_expires_at_date=framework_expires_at_date
         ), status_code
     except AuthException:
         abort(500, "Application error")
--- a/tests/main/views/test_services.py
+++ b/tests/main/views/test_services.py
@@ -273,7 +273,7 @@ class TestServicePage(DataAPIClientMixin, BaseApplicationTest):
         assert unavailable_banner.body_text() == 'The {} framework expired on {}. Any existing contracts with {} are' \
             ' still valid.'.format(
                 self.service['services']['frameworkName'],
-                'Tuesday 5 January 2016',
+                'Thursday 6 January 2000',
                 self.service['services']['supplierName'],
         )
 
--- a/tests/main/views/test_services.py
+++ b/tests/main/views/test_services.py
@@ -261,6 +261,9 @@ class TestServicePage(DataAPIClientMixin, BaseApplicationTest):
         self.data_api_client.get_service.return_value = self.service
 
         service_id = self.service['services']['id']
+        framework_copy_with_updated_expiry_date = self.data_api_client.get_framework.return_value.copy()
+        framework_copy_with_updated_expiry_date['frameworks']['frameworkExpiresAtUTC'] = '2020-01-05T17:01:07.649587Z'
+        self.data_api_client.get_framework.return_value = framework_copy_with_updated_expiry_date
         res = self.client.get('/g-cloud/services/{}'.format(service_id))
         assert res.status_code == 410
 
@@ -273,7 +276,7 @@ class TestServicePage(DataAPIClientMixin, BaseApplicationTest):
         assert unavailable_banner.body_text() == 'The {} framework expired on {}. Any existing contracts with {} are' \
             ' still valid.'.format(
                 self.service['services']['frameworkName'],
-                'Thursday 6 January 2000',
+                'Sunday 5 January 2020',
                 self.service['services']['supplierName'],
         )
 
--- a/app/templates/service.html
+++ b/app/templates/service.html
@@ -38,7 +38,7 @@
         heading = "This {} service is no longer available to buy.".format(service.frameworkName),
         message = " The {} framework expired on {}. Any existing contracts with {} are still valid.".format(
             service.frameworkName,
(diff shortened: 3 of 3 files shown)


app/main/views/g_cloud.py
```diff
@@ -209,7 +209,7 @@ def get_service_by_id(service_id):
         status_code = 200
         if service['serviceMadeUnavailableAuditEvent'] is not None:
             service_unavailability_information = {
-                'date': dateformat(service['serviceMadeUnavailableAuditEvent']['createdAt']),
+                'date': dateformat(framework.get('frameworkExpiresAtUTC')),
```
katstevens, line 212: I'm thinking we should either make this fail noisily if there's no `frameworkExpiresAtUTC` field on the framework object and use `framework['frameworkExpiresAtUTC']`, or provide an alternative bit of content in the template (perhaps just `'Supplier X stopped offering this service.'` without the date)? While all our current frameworks have that field, it's not required when we create it - this way we absolutely won't forget to add it 😸

tests/main/views/test_services.py
```diff
@@ -218,11 +218,12 @@ def test_enabled_service_has_unavailable_banner(self):
         assert unavailable_banner.exists
         assert unavailable_banner.heading_text() == '{} stopped offering this service on {}'.format(
             self.service['services']['supplierName'],
-            'Tuesday 5 January 2016.',
+            'Thursday 6 January 2000.',
         )
         assert unavailable_banner.body_text() == 'Any existing contracts for this service are still valid.'
 
     def test_disabled_service_has_unavailable_banner(self):
+        self.service.frameworkName
```
katstevens, line 226: I think this might have been c&p-ed from somewhere?

gidsg: 🤦  good spot! Have removed in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1050/commits/c52a9fce1342b30a24f7c5d5c96b596ce82110ea

app/main/views/g_cloud.py
gidsg: Hmm, wouldn't we only find this out when a framework is expired so maybe we would forget to add it initially still?

katstevens: 🤔  Normally we'd find out on Preview first (we switch the status there first to sanity check that everything's ok - including old service pages). However I don't think we have any automated coverage for it... Perhaps [an alternative date-less message in the template here](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/app/templates/service.html#L39) is the best bet then.

gidsg: @katstevens OK, I changed it now as per your first suggestion in https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1050/commits/baad6dd06a5ef1ce54c06cd5f03f2d3c8091b73c.

app/main/views/g_cloud.py
```diff
@@ -209,7 +209,7 @@ def get_service_by_id(service_id):
         status_code = 200
         if service['serviceMadeUnavailableAuditEvent'] is not None:
             service_unavailability_information = {
-                'date': dateformat(service['serviceMadeUnavailableAuditEvent']['createdAt']),
+                'date': dateformat(framework['frameworkExpiresAtUTC']),
```
katstevens, line 212: I've just realised we're using `service_unavailability_information.date` in two places on the template. The first case is the bug we're trying to solve, where the framework has expired. The second case is where the supplier has disabled the service themselves (which can happen at any time during the framework). So we do need to provide `(service['serviceMadeUnavailableAuditEvent']['createdAt']` to the template context.

tests/main/views/test_services.py
```diff
@@ -244,7 +244,7 @@ def test_disabled_service_has_unavailable_banner(self):
         assert unavailable_banner.exists
         assert unavailable_banner.heading_text() == '{} stopped offering this service on {}'.format(
             self.service['services']['supplierName'],
-            'Tuesday 5 January 2016.',
+            'Thursday 6 January 2000.',
```
katstevens, line 247: So this test should still show `Tuesday 5 January 2016`, as the date the supplier made the service unavailable.

tests/main/views/test_services.py
```diff
@@ -273,7 +273,7 @@ def test_expired_framework_causes_service_to_have_unavailable_banner(self):
         assert unavailable_banner.body_text() == 'The {} framework expired on {}. Any existing contracts with {} are' \
             ' still valid.'.format(
                 self.service['services']['frameworkName'],
-                'Tuesday 5 January 2016',
+                'Thursday 6 January 2000',
```
katstevens, line 276: I'm not 100% sure where this test date is coming from - where are we setting `frameworkExpiresAtUTC`?

app/main/views/g_cloud.py
gidsg: @katstevens Ahh, OK, makes sense.

May I ask in the case that the framework has expired do we also want to return 410 as we do if the service is unavailable? i.e. in: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/baad6dd06a5ef1ce54c06cd5f03f2d3c8091b73c/app/main/views/g_cloud.py#L215-L216

katstevens: In theory yes. However that might be a separate tech debt ticket 😸

tests/main/views/test_services.py
gidsg: @katstevens Good question! It was coming from this fixture: https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/be59e960522a02ff26e03e0a597400ca35d9c3b3/tests/main/views/test_services.py#L40 
I have updated the test now in be59e96 to update this value explicitly so it is clearer what value is being read.

tests/main/views/test_services.py
gidsg: @katstevens Good spot! Fixed in 47900828620dd5e2f8f03ee21e9364c4711828b1

tests/main/views/test_services.py
katstevens: Aha - from the `FrameworkStub`, gotcha. Good call on setting an explicit value 👍

PR #1052, opened 2020-06-19
Remove .app-break-link class

domoscargin:
The `app-break-link` class served to wrap long URLs in a summary table.

Since the upgrade to summary lists, we no longer need this class, as Summary Lists wrap text already.


lfdebrux:
We should also update dmutils in this PR when alphagov/digitalmarketplace-utils#566 is merged to make sure the links have the `govuk-link` class.

katstevens:
https://github.com/alphagov/digitalmarketplace-utils/pull/566 now merged!

PR #1054, opened 2020-06-24
Fix bug in save search view

lfdebrux:
The HTTP request `GET /buyers/direct-award/g-cloud/save-search` (note no query string) from a logged in user causes a 500.

This PR catches the case where the required query param is not in the URL and raises a HTTP 400 instead.

### DIFF
--- a/app/main/views/g_cloud.py
+++ b/app/main/views/g_cloud.py
@@ -420,7 +420,10 @@ def save_search(framework_family):
     framework = framework_helpers.get_latest_live_framework_or_404(all_frameworks, framework_family)
     lots_by_slug = framework_helpers.get_lots_by_slug(framework)
 
-    search_query = url_decode(request.values.get('search_query'))
+    if "search_query" not in request.values:
+        abort(400)
+
+    search_query = url_decode(request.values['search_query'])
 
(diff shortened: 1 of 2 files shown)


app/main/views/g_cloud.py
```diff
@@ -420,7 +420,10 @@ def save_search(framework_family):
     framework = framework_helpers.get_latest_live_framework_or_404(all_frameworks, framework_family)
     lots_by_slug = framework_helpers.get_lots_by_slug(framework)
 
-    search_query = url_decode(request.values.get('search_query'))
+    try:
+        search_query = url_decode(request.values['search_query'])
+    except KeyError:
+        abort(400)
```
katstevens, line 426: Do we need a `try...except`? How about

```
    if 'search_query' not in request.values:
        abort(400)
```

lfdebrux: Sure that would work. Out of curiosity, what is the reason to avoid the `try...except`?

lfdebrux: I've made the change.

katstevens: > Out of curiosity, what is the reason to avoid the try...except?

Good question! It's no less valid to do a `try...except`, it just feels easier to understand what's going on from the `if` statement?

PR #1055, opened 2020-06-25
Use govuk-frontend heading styles on Crown Hosting page

lfdebrux:
Small fix for headings on the Crown Hosting page that I noticed needed doing.

### Screenshot

Screenshot of the crown hosting page after PR changes

### DIFF
--- a/app/templates/content/index-crown-hosting.html
+++ b/app/templates/content/index-crown-hosting.html
@@ -39,7 +39,7 @@
           to talk about your datacentre hosting needs.
 
         </p>
-        <h2 class='summary-item-heading'>
+        <h2 class='govuk-heading-m'>
           Features
         </h2>
         <p class="govuk-body">
@@ -62,7 +62,7 @@
             leading environmental performance
           </li>
         </ul>
-        <h2 class='summary-item-heading'>
+        <h2 class='govuk-heading-m'>
           Benefits
         </h2>
         <p class="govuk-body">
@@ -96,7 +96,7 @@
         </ul>
       </div>
       <div class="govuk-grid-column-one-third">
-        <h2>
+        <h2 class="govuk-heading-m">
           Crown Hosting contact
         </h2>
         <p class="govuk-body">
--- a/app/templates/content/index-crown-hosting.html
+++ b/app/templates/content/index-crown-hosting.html
@@ -33,12 +33,7 @@
           The Crown Hosting Data Centres framework is available to all central
           government departments, arm’s length bodies and the wider public sector.
         </p>
-        <p class="govuk-body">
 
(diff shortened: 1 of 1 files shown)


app/templates/content/index-crown-hosting.html
```diff
@@ -39,7 +39,7 @@ <h1 class="govuk-heading-xl">
           to talk about your datacentre hosting needs.
 
         </p>
-        <h2 class='summary-item-heading'>
+        <h2 class='govuk-heading-l'>
```
katstevens, line 42: Any reason why we've gone for `govuk-heading-l` and not `-m`? I think we've used medium for pretty much all the other `h2`s so far, and it more closely matches the size on production at the moment?

app/templates/content/index-crown-hosting.html
```diff
@@ -96,7 +96,7 @@ <h2 class='summary-item-heading'>
         </ul>
       </div>
       <div class="govuk-grid-column-one-third">
-        <h2>
+        <h2 class="govuk-heading-m">
```
katstevens, line 99: While we're sorting out typography on this page, can we do something about the `<strong>` tag around `Crown Commercial Service` on line 103? I don't like the positioning to the right, either 😹  

My ideal solution for this would be to move that block into the main column (underneath the 'read more' section, which goes after the contact section on mobile view), and have something like this: 

```
<h2 class="govuk-heading-m">Crown Hosting contact</h2>
<p class="govuk-body">Email <a class="govuk-link" href="mailto:crownhosting@crowncommercial.gov.uk">crownhosting@crowncommercial.gov.uk</a> if you need support buying or selling Crown Hosting services.</p>
```
crown-hosting

However I realise this is getting a little out of scope!

@NoraGDS What do you think?

lfdebrux: Ah good idea. I'm happy to do that in this PR.

app/templates/content/index-crown-hosting.html
lfdebrux: Good catch! I will change it.

app/templates/content/index-crown-hosting.html
lfdebrux: Why do we need the email address on the page twice? 🤔

katstevens: Good question!

NoraGDS: The second paragraph from the top can be deleted. It duplicates the information in the contact section, which looks clearer in the new position as well.

lfdebrux: Thanks @NoraGDS !

PR #1056, opened 2020-06-26
Replace temporary notification banners with dmBanner

katstevens:
https://trello.com/c/7UUGcFmw/92-2-replace-notification-banners-with-digitalmarketplace-govuk-frontend-banner-component-in-g-cloud-service-page-and-dos-opportuni

Remarkably straightforward! Thank @agalamatis for doing the hard work up front on the `dmBanner` macro in digitalmarketplace-govuk-frontend.

Changes made in this PR:
- The new banners are now used on Brief pages, Service pages, and the Direct Award standstill messages for buyers (these have been moved from the right-hand side to the top, to be consistent with our other banners - see screenshot below).
- I've gone for heading-level `h2` on the banners as that's what we had before, and we don't want multiple `h1`s on the page of course. However the font size comes up much smaller than previously. Is this a good thing? 
- Renamed bits of Python code to no longer refer to `temporary_message`, hopefully this will make things less confusing for future Buyer FE coders
- Decoupled the home page framework messages from the `toolkit/temporary-message.html` macro and just hardcoded the `<div>`, as it's not really a proper temporary message or a banner

Not made in this PR:
- Removing every trace of `temporary-message` SCSS. It's too entangled and this PR is already too big

Screenshot of new Direct Award standstill message (see ticket for 'before'):
direct-award-standstill-message


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/main/helpers/search_save_helpers.py
```diff
@@ -54,7 +54,7 @@ def search_count(self):
         return int(self.search_summary.count)
 
 
-def get_saved_search_temporary_message_status(project, framework, following_framework):
+def get_saved_search_banner_message_status(project, framework, following_framework):
```
lfdebrux, line 57: :100:

This is really helpful, thanks! 😁

app/templates/brief.html
```diff
@@ -25,53 +25,41 @@
...
-      with
-      type = "temporary-message",
-      heading = "This opportunity is closed for applications",
-      message = "The deadline was {}".format(brief.applicationsClosedAt|dateformat)
-    %}
-      {% include "toolkit/notification-banner.html" %}
-    {% endwith %}
+    {{ dmBanner({
+      "title": "This opportunity is closed for applications",
+      "headingLevel": 2,
```
lfdebrux, line 30: The default heading level is 2, so we don't need to specify it.

lfdebrux:
I wonder if we should move the banners to below the page `h1` 🤔  It would be more semantically correct.

katstevens:
> I wonder if we should move the banners to below the page `h1` 🤔 It would be more semantically correct.

I think we'd need designer input for that...

app/templates/brief.html
katstevens: So it is! Will remove it

lfdebrux:
> I've gone for heading-level h2 on the banners as that's what we had before, and we don't want multiple h1s on the page of course. However the font size comes up much smaller than previously. Is this a good thing?

I think the font size is okay, if you want though we can look at modifying the component to have the ability to control the font size using the `govuk-heading--*` styles.

lfdebrux:
We'll need to update the functional tests to be able to recognise the new banners.

> Then I see a temporary-message banner message containing 'Awarded to'
> features/step_definitions/common_steps.rb:252
> Unable to find visible css ".banner-temporary-message-with-action" (Capybara::ElementNotFound)

katstevens:
> We'll need to update the functional tests to be able to recognise the new banners.
> 
> > Then I see a temporary-message banner message containing 'Awarded to'
> > features/step_definitions/common_steps.rb:252
> > Unable to find visible css ".banner-temporary-message-with-action" (Capybara::ElementNotFound)

Done! https://github.com/alphagov/digitalmarketplace-functional-tests/pull/774

PR #1058, opened 2020-07-09
Replace saved search summary tables with govukTable

domoscargin:
https://trello.com/c/EcbpN7V0/129-2-replace-summary-tables-on-saved-searches-page-with-tables

Replaces the summary table macros on the saved search page with the [Design System Table](https://design-system.service.gov.uk/components/table/).

### DIFF
--- a/app/templates/direct-award/index.html
+++ b/app/templates/direct-award/index.html
@@ -1,5 +1,5 @@
 {% extends "_base_page.html" %}
-{% import "toolkit/summary-table.html" as summary %}
+{% from "govuk-frontend/components/table/macro.njk" import govukTable %}
 {% block pageTitle %}
   Your saved searches - Digital Marketplace
 {% endblock %}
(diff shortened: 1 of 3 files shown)


app/templates/direct-award/index.html
```diff
@@ -30,47 +30,41 @@ <h1 class="govuk-heading-xl">
...
-      caption="Search ended",
-      empty_message="You have no saved searches",
-      field_headings=[
-        "Name",
-        "Exported",
-        "Status"
+  {% if open_projects|length %}
+    {{ govukTable({
+      "caption": "Searching",
+      "captionClasses": "govuk-heading-l",
```
katstevens, line 36: This probably wants to be a `govuk-heading-m` if it's an `h2`? Same for the other `h2`s on this template

tests/main/helpers/test_direct_award_helpers.py
```diff
@@ -0,0 +1,21 @@
...
+)
+def test_status_displays_correctly(result, expected):
+    framework = {'family': 'g-cloud'}
+    project = {'outcome': {'result': result}}
+    status = set_direct_award_display_status(project, framework)
+    assert status == expected
+
+# TODO test for link statuses - I get the following error:
+# Attempted to generate a URL without the application context being pushed.
+# This has to be executed when application context is available.
```
katstevens, line 21: The `url_for` function needs the Flask app context to resolve the URLs. To test, you can either provide the app context (there's an example of doing this in a test [here](https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/master/tests/main/views/test_marketplace.py#L1582) - that test needs access the Flask app config but it's the same principle), or you can patch out the `url_for` function to return whatever you like. I'd recommend the former 😸

app/main/helpers/direct_award_helpers.py
```diff
@@ -29,3 +31,29 @@ def get_direct_award_projects(data_api_client, user_id, return_type="all", sort_
         if sort_by_key:
             res[return_type].sort(key=itemgetter(sort_by_key))
         return res[return_type]
+
+
+def set_direct_award_display_status(project, framework):
```
katstevens, line 36: I'm a little hesitant about burying the content in a helper function, as it makes it harder to find/change later if we want to tweak the text. However I understand the need to split this bit of complexity out of the main view.

I'd suggest grouping the content together at least, something like:

```
project_outcome_content = {
  "awarded": "Awarded",
  "cancelled": "The work has been cancelled",
  "none-suitable": "No suitable services found",
  "download_results": "Download results",
  "awaiting_outcome": "Tell us the results"
}
```

Then we can use that dict to do something like:

```
if project['outcome']:
  output = project_outcome_content[project['outcome']['result']]
elif not project['downloadedAt']:
  output = link_to(download_url, project_outcome_content['download_results'])
else:
  output = link_to(tell_us_the_outcome_url, project_outcome_content['awaiting_outcome'])
```

Whether that's simple enough to put back into the main view, I will leave up to you... 😸

app/templates/direct-award/index.html
domoscargin: I agree this looks better - was trying to step down from the `<h1>` which is `xl`, but I think the visual hierarchy is still clear (as well as the semantic one).

app/main/helpers/direct_award_helpers.py
domoscargin: Agree with all of this - I should've taken a step back and thought about how I should structure things. Changing this has the added benefit of not requiring the `link_to` function from dmutils as well, so I'm going to close that.

katstevens:
Looking good! How do the functional tests fare against this branch?

domoscargin:
> Looking good! How do the functional tests fare against this branch?

All passes for me locally.

PR #1060, opened 2020-07-10
Replace text input on search for G-Cloud services journey

domoscargin:
https://trello.com/c/GRm5KiLE/111-2-replace-text-input-component-in-search-for-g-cloud-services-journey-except-keyword-search

Walked through this journey and there's only one form which directly uses text inputs.

* `/buyers/direct-award/g-cloud/save-search` uses `toolkit/forms/selection-buttons.html` which reveals an inner text input.
* `/buyers/direct-award/g-cloud/projects/<id>/did-you-award-contract` has two text inputs:
  * One is for currency. I've not converted this one because the [Design System Form input prefix/suffix variant](https://github.com/alphagov/govuk-design-system-backlog/issues/134#issuecomment-654968716) is due to be published soon, and we should use that to display the currency symbols.
  * The other field I've replaced directly with a `govukInput` macro using existing form data.

- [x] Have run local tests
- [x] Have run functional tests locally

### DIFF
--- a/tests/main/views/test_direct_award.py
+++ b/tests/main/views/test_direct_award.py
@@ -1049,13 +1049,18 @@ class TestDirectAwardTellUsAboutContract(TestDirectAwardBase):
         assert res.status_code == 400
 
         doc = html.fromstring(res.get_data(as_text=True))
-        assert doc.xpath('count(//span[@class="validation-message"])') == 1
+        assert (doc.xpath('count(//span[@class="validation-message"])') == 1 or
+                doc.xpath('count(//span[@class="govuk-error-message"])') == 1)
         errors = doc.cssselect('div.govuk-error-summary a')
         assert len(errors) == 1
         assert len(errors[0].text_content()) > 0
 
         for field, value in data.items():
-            assert doc.xpath(f'//input[@name="{field}"]/@value')[0] == value
+            input_element = doc.xpath(f'//input[@name="{field}"]')[0]
+            if not value:
+                assert('value' not in input_element.attrib)
+            else:
+                assert input_element.attrib['value'] == value
 
     def test_if_end_date_is_before_start_date_raise_400_and_show_validation_message(self, client, data):
         data.update({'end_date-year': str(int(data['start_date-year']) - 1)})
--- a/app/assets/scss/application.scss
+++ b/app/assets/scss/application.scss
@@ -79,6 +79,7 @@ $govuk-global-styles: true;
 @import "_help-page.scss";
 
 // Overrides
+@import "overrides/_text-input.scss";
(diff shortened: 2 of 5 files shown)


lfdebrux:
> One is for currency. I've not converted this one because the Design System Form input prefix/suffix variant is due to be published soon, and we should use that to display the currency symbols.

That's great news!

tests/main/views/test_direct_award.py
```diff
@@ -1049,13 +1049,18 @@ def test_invalid_data_raises_400_and_has_validation_messages_but_remains_in_form
...
+                doc.xpath('count(//span[@class="govuk-error-message"])') == 1)
         errors = doc.cssselect('div.govuk-error-summary a')
         assert len(errors) == 1
         assert len(errors[0].text_content()) > 0
 
         for field, value in data.items():
-            assert doc.xpath(f'//input[@name="{field}"]/@value')[0] == value
+            input_element = doc.xpath(f'//input[@name="{field}"]')[0]
+            if (value is None or value == ''):
+                assert(not input_element.xpath('/@value'))
```
gidsg, line 1061: I am a bit confused what are we asserting in this case?

domoscargin: You're right, this is confusing and hacky.

Previously the element being checked _always_ had a `value` attribute, even if the form input was empty.

With the GOVUK Frontend component, if the input is empty, it won't have a `value` attribute, so this is essentially checking that there's no element with a `value` attribute.

I've pushed a change to make it clearer.

It's hacky - previously the element being checked always had a `value` attribute, but now it's possible to not have that attribute if the form value is empty.

So in the case where the input is empty, this is asserting that the element does not have a `value` attribute.

I'm not v. familiar with xpath, so I dunno if there's a way

tests/main/views/test_direct_award.py
```diff
@@ -1049,13 +1049,18 @@ def test_invalid_data_raises_400_and_has_validation_messages_but_remains_in_form
...
+        assert (doc.xpath('count(//span[@class="validation-message"])') == 1 or
+                doc.xpath('count(//span[@class="govuk-error-message"])') == 1)
         errors = doc.cssselect('div.govuk-error-summary a')
         assert len(errors) == 1
         assert len(errors[0].text_content()) > 0
 
         for field, value in data.items():
-            assert doc.xpath(f'//input[@name="{field}"]/@value')[0] == value
+            input_element = doc.xpath(f'//input[@name="{field}"]')[0]
+            if (value is None or value == ''):
```
gidsg, line 1060: I think this is equivalent and more concise:
```suggestion
            if not value:
```

domoscargin: Thanks!

PR #1074, opened 2020-09-24
Replace filters with Option Select component

domoscargin:
https://trello.com/c/VCANsToa/233-3-replace-search-filters-with-govuk-checkbox-groups-option-select

Adding option select should resolve the accessibility issues attached to this ticket.

Uses the new option select component: https://github.com/alphagov/digitalmarketplace-govuk-frontend/pull/183


lfdebrux (approved):
Works for me 👍

PR #1075, opened 2020-09-25
Fix malformed HTML for guidance link

bjgill:
A missing '"' was causing problems. It didn't affect users - the link still worked fine. However, it looked really weird and probably messed up the analytics.

Now matches the form of the link just above it in the file


domoscargin (approved):
Good spot!

PR #1078, opened 2020-10-05
Use new pagination component on search pages

domoscargin:
https://trello.com/c/d1wSRrnT/179-2-use-govuk-style-for-previous-and-next-links-on-search-results-page


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/templates/search/_pagination.html
```diff
@@ -1,30 +1,24 @@
-<nav role="navigation">
-    <ul class="previous-next-navigation">
+{% if not dmPreviousNextPagination %}
+  {% from "digitalmarketplace/components/previous-next-pagination/macro.njk" import dmPreviousNextPagination %}
+{% endif %}
```
lfdebrux, line 3: I'm not sure the import guard is necessary

domoscargin: I thought so, but it ends up failing this test with a `'?live-results=true'` arg:

https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/37f0d7da87b41843f809c107b7cd376d6487308f/tests/main/views/test_marketplace.py#L1794-L1800

I think it's related to this bit here which seems to only be rendering the parts it needs, so it doesn't catch the base page?
https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/37f0d7da87b41843f809c107b7cd376d6487308f/app/main/views/marketplace.py#L349-L373

lfdebrux: Ah, alright then. Thanks for explaining that.

lfdebrux (approved):
Looks good to me 👍

PR #1082, opened 2020-10-08
hardcode expired banner text for G11

gidsg:
https://trello.com/c/26eeD0Pd/1949-g11-banner-content-change
Hardcoding this banner for G11 for now, ideally should be derived from the framework object
Screenshot 2020-10-08 at 16 53 19

### DIFF
--- a/app/templates/service.html
+++ b/app/templates/service.html
@@ -33,23 +33,23 @@
   <div class="govuk-grid-column-full">
     {% if service_unavailability_information.type == 'framework_update' %}
     {# TODO: Hardcoding this banner for G11 for now, should be derived from the framework object  #}
-       {% if service.frameworkName == 'G-Cloud 11' %}
-           {{ dmBanner({
+        {% if service.frameworkName == 'G-Cloud 11' %}
+            {{ dmBanner({
         "title": "This G-Cloud 11 service has been suspended and is no longer searchable on the Digital Marketplace.",
         "text": "The G-Cloud 11 framework expires on 18 December 2020 to allow in progress procurements to be concluded. Existing contracts with {} are still valid.".format(
             service.supplierName)
-      }) }}
-   {% else %}
-      {{ dmBanner({
+        }) }}
+        {% else %}
+            {{ dmBanner({
         "title": "This {} service is no longer available to buy.".format(service.frameworkName),
         "text": "The {} framework expired on {}. Any existing contracts with {} are still valid.".format(
             service.frameworkName,
             framework_expires_at_date,
             service.supplierName)
-      }) }}
-   {% endif %}
+        }) }}
+        {% endif %}
     {% else %}
-      {{ dmBanner({
+        {{ dmBanner({
         "title": "{} stopped offering this service on {}.".format(service.supplierName, service_unavailability_information.date),
         "text": "Any existing contracts for this service are still valid."
       }) }}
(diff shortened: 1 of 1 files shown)


app/templates/service.html
```diff
@@ -32,13 +32,22 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
     {% if service_unavailability_information.type == 'framework_update' %}
+    {# TODO: Hardcoding this banner for G11 for now, should be derived from the framework object  #}
+       {% if service.frameworkName == 'G-Cloud 11' %}
+           {{ dmBanner({
+        "title": "This G-Cloud 11 service has been suspended and is no longer searchable on the Digital Marketplace.",
+        "text": "The G-Cloud 11 framework expires on 18 December 2020 to allow in progress procurements to be concluded. Existing contracts with {} are still valid.".format(
```
bjgill, line 39:
```suggestion
        "text": "The G-Cloud 11 framework expires on 18 December 2020 to allow procurements in progress to be concluded. Existing contracts with {} are still valid.".format(
```

bjgill (commented):
Looks good. Just a wording suggestion

app/templates/service.html
bjgill: Or should we be more assertive? "Procurements in progress to be concluded before the G-Cloud 11 framework expires on 18 December 2020."?

gidsg: I feel like the wording is really one for @NoraGDS who's not in today. Can we use this for now and then can revise when @NoraGDS is back?

app/templates/service.html
```diff
@@ -32,13 +32,22 @@
...
+      }) }}
+   {% else %}
       {{ dmBanner({
         "title": "This {} service is no longer available to buy.".format(service.frameworkName),
         "text": "The {} framework expired on {}. Any existing contracts with {} are still valid.".format(
             service.frameworkName,
             framework_expires_at_date,
             service.supplierName)
       }) }}
+   {% endif %}
```
bjgill, line 50: Is this indentation correct? It doesn't seem to line up with the `if` to me...

app/templates/service.html
phantas: who is the source for this content? I agree with the change in any case. I'd still like Nora's eyes on this but I am happy to go ahead for now and change later if needed.

bjgill: Oh, I see this wording came from CCS. Never mind, then. Happy for this content to go in as-is and let a content professional to take a look tomorrow.

phantas (approved):

app/templates/service.html
phantas: it is correct

app/templates/service.html
```diff
@@ -32,13 +32,22 @@
...
     {% else %}
```
phantas, line 51: this however isn't.

app/templates/service.html
gidsg: OK I don't know, I got intellij to indent it and committed that in 49bc491

app/templates/service.html
gidsg: should be fixed in 49bc491

app/templates/service.html
gidsg: yes, source is CCS verbatim for now.

bjgill (approved):

PR #1088, opened 2020-10-19
Remove toolkit styles and components from search page

domoscargin:
https://trello.com/c/v8J4QAoc/238-2-review-search-page-changes

* Live search maintains its jQuery-ness, but with updated object refs for `DMGOVUKFrontend`. Removing jQuery can be handled separately from DS3.
* The clear filters script doesn't appear to be doing anything, as on the page, clicking the clear filters link just reloads the page. So I got rid of it.
* As far as I can tell, the toolkit option select was only used on the search page, so I've got rid of that import too.


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

lfdebrux (approved):
This is looking really good! ✨

PR #1090, opened 2020-10-20
Replace date input component with govukDateInput

domoscargin:
https://trello.com/c/iAataleT/209-2-replace-date-input-component-in-search-g-cloud-services-journey-2

This change does make the view template more verbose, but we could address this separately as a task to review how we generate presentation data for WTForm-based form questions.

### DIFF
--- a/app/templates/direct-award/tell-us-about-contract.html
+++ b/app/templates/direct-award/tell-us-about-contract.html
@@ -40,9 +40,69 @@
       <form method="POST" action="{{ url_for('direct_award.tell_us_about_contract', framework_family=framework.family, project_id=project.id, outcome_id=outcome_id) }}" class="tell-us-about-contract-form" novalidate>
         <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
 
-        {{ form.start_date}}
+        {{ govukDateInput ({
+          "id": "input-start_date",
+          "namePrefix": "start_date",
+          "fieldset": {
+            "legend": {
+              "text": form.start_date.question,
+              "classes": "govuk-fieldset__legend--m"
+            }
+          },
+          "hint": {
+            "text": form.start_date.hint
+          } if form.start_date.hint,
+          "errorMessage": errors.start_date.errorMessage if errors.start_date.errorMessage,
+          "items": [
+            {
+              "classes": "govuk-input--width-2",
+              "name": "day",
+              "value": form.start_date.value['start_date-day'] if form.start_date.value['start_date-day']
+            },
+            {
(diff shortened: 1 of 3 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/templates/direct-award/tell-us-about-contract.html
```diff
@@ -40,9 +40,69 @@ <h1 class="govuk-heading-xl">Tell us about your contract</h1>
...
+              "classes": "govuk-fieldset__legend--m"
+            }
+          },
+          "hint": {
+            "text": form.start_date.hint
+          } if form.start_date.hint,
+          "errorMessage": errors.start_date.errorMessage if errors.start_date.errorMessage,
+          "items": [
+            {
+              "classes": "govuk-input--width-2",
```
lfdebrux, line 58: We should add the class `govuk-input--error` if there is an error

https://design-system.service.gov.uk/components/date-input/#error-messages

domoscargin: I'm wondering whether I should look into adding date validators to dm-utils (I'm wary of messing around with the `DMDateField` class since it's used elsewhere), since users might be confused whether we highlight all fields with an error, or highlight no fields with an error (if they've only entered an incorrect day value, for example).

We've got a ticket to introduce more granular date errors nominally in sprint 3.

domoscargin: In the meantime, I've highlighted each field if there's any error.

app/templates/direct-award/tell-us-about-contract.html
```diff
@@ -40,9 +40,69 @@ <h1 class="govuk-heading-xl">Tell us about your contract</h1>
...
+              "classes": "govuk-fieldset__legend--m"
+            }
+          },
+          "hint": {
+            "text": form.start_date.hint
+          } if form.start_date.hint,
+          "errorMessage": errors.start_date.errorMessage if errors.start_date.errorMessage,
+          "items": [
+            {
+              "classes": "govuk-input--width-2" + (' govuk-input--error' if errors.start_date.errorMessage),
```
lfdebrux, line 58: Hmm, it doesn't like this, try
```suggestion
              "classes": "govuk-input--width-2" ~ (' govuk-input--error' if errors.start_date.errorMessage),
```

domoscargin: Ah yes, of course!

lfdebrux (approved):

PR #1093, opened 2020-10-26
Use govukInsetText for search summary on save search page

domoscargin:
https://trello.com/c/nGmcQJg6/206-1-replace-notice-panel-with-govuk-frontend-inset-text-component-in-buyer-frontend

Screenshot 2020-10-26 at 11 20 41


app/assets/scss/_search-summary.scss
```diff
@@ -32,3 +32,12 @@
...
 }
+
+/* On the Save search page, we display the search summary in a govukInsetText,
+so we need to adjust padding to match */
+.save-search-page {
+  .search-summary {
+      padding: 15px;
+      margin-top: 0;
+  }
+}
```
lfdebrux, line 43: Can we use a [spacing override class](https://design-system.service.gov.uk/styles/spacing/) instead?

app/assets/scss/application.scss
```diff
@@ -67,7 +67,6 @@ $govuk-global-styles: true;
 @import "_index-page.scss";
 @import "_legal.scss";
 @import "_notice.scss";
-@import "_save-search-page.scss";
```
lfdebrux, line 70: 👏

app/assets/scss/_search-summary.scss
domoscargin: I've just noticed this actually belongs to an existing ticket, so I'll park it for now and comment there.

PR #1096, opened 2020-10-27
Remove use of instruction list template

lfdebrux:
https://trello.com/c/XsNDGQLk/205-3-remove-instruction-list-from-search-g-cloud-services-journey

Rewrites `app/templates/direct-award/view_page.html` to use HTML and if statements, instead of the rather difficult to read instruction list structure.

We wrap the whole thing in an ordered list and give it the `app-task-list` class to make testing easier and to allow it to be made more of a task list in future, but we don't actually bring in any of the task list styles. Instead we add a few overrides to make it look as much like the previous design as possible.

### Screenshots

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top">
Screenshot before changes
</td>
<td valign="top">
Screenshot after changes
</td>
</tr>
</tbody>
</table>


app/assets/scss/application.scss
```diff
@@ -64,7 +63,6 @@ $govuk-global-styles: true;
 @import "_gcloud_browse_list.scss";
 @import "_index-page.scss";
 @import "_search-page.scss";
-@import "_search_overview.scss";
```
domoscargin, line 67: 🔥

app/assets/scss/overrides/_tag.scss
```diff
@@ -0,0 +1,17 @@
+.app-tag--box {
+  border: 3px solid $grey-1;
+  font-weight: bold;
+  color: $grey-1;
+  padding: 0.5em 0.75em 0.33em 0.5em;
```
domoscargin, line 5: I think this is fine as a temporary override until we rejig the page, but I've never been a fan of `em` since they can be quite volatile (govuk-frontend has tried to ditch 'em entirely I think). In this case we know the parent details, so not a particular issue.

app/templates/direct-award/view-project.html
```diff
@@ -69,141 +67,287 @@ <h1 class="govuk-heading-xl">{{ page_title }}</h1>
...
+          Export a list of the services you’ve found. Download suppliers’ service
+          descriptions and contact details to help you assess services. Keep a copy
+          for your records.
+        </p>
+
+        {% if project %}
+          {% if not can_end_search %}
+          <div class="govuk-form-group--error">
+            <span class="govuk-error-message">
+              <span class="govuk-visually-hidden">Error:</span>
```
domoscargin, line 151: Do we need to treat this like a normal form error (ie: display an error summary and reset the page title)?

It prevents any further actions on the page, but the action to fix is to follow the previous link. Perhaps when we have an error here we could display a button/link which redirects to the search page.

Or perhaps that's not necessary until we overhaul the page.

domoscargin (approved):
LGTM!

(a few comments, but I don't think anything is a showstopper)

app/assets/scss/overrides/_tag.scss
lfdebrux: Cool, thanks for flagging up, I'll bear that in mind in future!

app/templates/direct-award/view-project.html
lfdebrux: I think that's a good idea. We should be as consistent as we can and it's not a big change. I'll do that now.

Ideally we wouldn't let the user get to this state anyway (perhaps by greying out the save search button until they have less than 30 results, although I suppose it could be possible for the number of results to grow over time). But we'll have to leave that for later.

lfdebrux: Hmm, I've given it a try in 5a57dc8, what do you think? Currently the link just goes to the first item in the task list, which isn't that obvious, so maybe we should move the error message as well? Or we could just have the link in the error summary go to the search page, but that feels like surprising behaviour.

Screenshot of error summary

domoscargin: Hmm, I actually quite like moving the error message - we have a built in action, and the rest of the task list looks more consistent that way (All "cannot start yet").

domoscargin: Agree we shouldn't link directly from the Error Summary

lfdebrux: Et voila! Commit 14517b9

Screenshot of error summary and error message

PR #1097, opened 2020-10-28
Replace radios in search G-Cloud services journey

domoscargin:
https://trello.com/c/UQtUlH0n/207-2-replace-radios-with-govuk-frontend-radios-component-in-search-g-cloud-services-journey

For the most part, these were straightforward, but the save search form made use of conditional reveal. This has [outstanding accessibility issues](https://github.com/alphagov/govuk-frontend/issues/1991) so I've moved that field to a new page instead.

The code is quite repetitive - I could achieve the same effect by putting conditional logic on the template itself, but I thought it best to keep the urls discrete. Additionally, I've used WTForms as this seemed like the most efficient way to handle submission and validation - thinking about how we replace 'em is probably a future task.

Functional tests: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/813

### DIFF
--- a/app/templates/direct-award/did-you-award-contract.html
+++ b/app/templates/direct-award/did-you-award-contract.html
@@ -1,8 +1,6 @@
 {% extends "_base_page.html" %}
 
-{% block pageTitle %}
-  {% if errors %}Error: {% endif %}Did you award a contract? - Digital Marketplace
-{% endblock %}
+{% set page_name = 'Did you award a contract?' %}
 
 {% block breadcrumb %}
 {{ govukBreadcrumbs({
@@ -24,7 +22,7 @@
       "text": project.name
     },
     {
-      "text": "Did you award a contract?"
+      "text": page_name
     }
   ]
 }) }}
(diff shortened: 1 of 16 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/templates/direct-award/did-you-award-contract.html
```diff
@@ -1,7 +1,7 @@
 {% extends "_base_page.html" %}
 
 {% block pageTitle %}
-  Did you award a contract? - Digital Marketplace
+  {% if errors %}Error: {% endif %}Did you award a contract? - Digital Marketplace
```
lfdebrux, line 4: Rather than doing this for all pages, can we do this once in the base page template?

domoscargin: The way it's currently set up, I think we'd have to use:

```
{% set page_name = "Save your search" %}

{% block pageTitle %}
  {{ super() }}{{ page_name }} - Digital Marketplace
{% endblock %}```

Which is similar in terms of "update all the pages"-iness, but not quite the same as just setting an error message (in case there's something else in the parent block).

So I think our options look like:

1. Update each page with the super() call.
2. Generate page names in the presenters.
3. Use a global page title variable and a function to set it
4. Add the `if errors` to each page

Comment (lfdebrux) [reply app/templates/direct-award/did-you-award-contract.html:4]: What about

```
# base_page.html
{% block pageTitle %}
  {% if errors %}Error: {% endif %}{{ page_name}} - Digital Marketplace
{% endblock %}
```

Would we need the super then?

Comment (domoscargin) [reply app/templates/direct-award/did-you-award-contract.html:4]: Passing from the child to the parent through blocks doesn't seem to work.

https://stackoverflow.com/questions/24833370/pass-variables-from-child-template-to-parent-in-jinja2/24836460#24836460

Comment (domoscargin) [reply app/templates/direct-award/did-you-award-contract.html:4]: Oh wait, and not have the block in the child? I'll give that a go.

Comment (domoscargin) [reply app/templates/direct-award/did-you-award-contract.html:4]: It works!

Comment (lfdebrux) [reply app/templates/direct-award/did-you-award-contract.html:4]: It that doesn't work, using a block for pageName should.

```
# base_page.html
{% block pageTitle %}
  {% if errors %}Error: {% endif %}{% block pageName %}{% endblock %} - Digital Marketplace
{% endblock %}

# child
{% block pageName%}Save your search{% endblock %}
```

domoscargin: Changes made. Running functional tests locally now.

lfdebrux (approved):

PR #1098, opened 2020-10-29
Replace notice/panel with govukInsetText

domoscargin:
https://trello.com/c/nGmcQJg6/206-1-replace-notice-panel-with-govuk-frontend-inset-text-component-in-buyer-frontend

There were a few niggles with govukInsetText and our search summary styles. Additionally, our search summary wasn't being nicely wrapped by `<p>` tags or similar.

So I rejigged the search-summary to make more bits work by default (changing `<em>` to to `<strong>` tags, for example, still allows screenreaders to highlight those bits of text, but also removes the need to style those `<em>`'s to be bold.

The only outlier is the search summary count, which we style differently on the view project page. Depending on how we restyle that page, we might be able to drop the custom search summary styling altogether and just apply a `govuk-!-font-size` or similar to the count span.

Functional tests: https://github.com/alphagov/digitalmarketplace-functional-tests/pull/814


app/main/views/g_cloud.py
```diff
@@ -574,7 +574,9 @@ def save_search(framework_family):
                                errors=get_errors_from_wtform(form),
                                form=form,
                                save_search_options=save_search_options,
-                               search_summary_sentence=search_summary.markup(),
+                               search_summary_sentence=Markup(
+                                   f'<p class="app-search-summary govuk-body-s">{search_summary.markup()}</p>'
```
gidsg, line 578: since we share this wrapper tag in both usages could we move it into a shared helper or have `search_summary` include it?

gidsg (approved):
LGTM (added one minor comment for possible enhancement) 👍

tests/main/views/test_search.py
```diff
@@ -265,10 +271,10 @@ def test_should_render_summary_with_a_group_of_1_boolean_filter(self):
             '/g-cloud/search?q=email&lot=cloud-software&phoneSupport=true')
         assert res.status_code == 200
         summary = self.find_search_summary(res.get_data(as_text=True))[0]
-        assert '<span class="search-summary-count">1</span> result found' \
-            ' containing <em>email</em> in' \
-            ' <em>Cloud software</em>' \
-            ' where user support is available by <em>phone</em>' in summary
+        assert '<span class="app-search-summary__count">1</span> result found' \
```
gidsg, line 274: would triple quotes work for these rather than using `\` to line break?

domoscargin: Doing something like:

```
assert """<span class="app-search-summary__count">1</span> result found containing
     <strong>email</strong> in <strong>Cloud software</strong>""" in summary
```

Results in an error:

```
assert '<span class="app-search-summary__count">1</span> result found containing\n             <strong>email</strong> in <strong>Cloud software</strong>' in '<span class="app-search-summary__count">1</span> result found containing <strong>email</strong> in <strong>Cloud software</strong>'
```
Probably my lack of Python knowledge - is there a way to strip the newline from the string?

app/main/views/g_cloud.py
domoscargin: Done, by adding a `wrap` arg to `SearchSummary.markup()`

tests/main/views/test_search.py
domoscargin: Looks like either I need to `.replace` them, or add a `\` within triple quotes? 

Is there a best practice?

domoscargin: Done - wrapped 'em all in parentheses.

PR #1099, opened 2020-10-30
Replace Contact Details

domoscargin:
https://trello.com/c/TmFkHa5K/268-1-remove-contact-details-component-from-buyer-frontend

Replace uses of https://trello.com/c/CHscCdOh/81-contact-details with [a `<p>` tag with some breaks](https://trello.com/c/CHscCdOh/81-contact-details#comment-5eafd8133d38676a7c134366) as the Design System suggests.

The old component adds analytics tracking to the email link, and some structured data.

I've kept the analytics, but removed the structured data - it's the only instance of microdata across our site, so seems like an outlier.

### DIFF
--- a/app/assets/scss/_atoz_navigation.scss
+++ b/app/assets/scss/_atoz_navigation.scss
@@ -47,12 +47,3 @@
     }
   }
 }
-
-.supplier-profile {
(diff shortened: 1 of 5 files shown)


gidsg (approved):
👍

PR #1102, opened 2020-11-03
Remove DMspeak

domoscargin:
https://trello.com/c/gkBaTC87/266-2-remove-dmspeak-from-buyer-frontend-1

There wasn't actually much dmspeak left in this app, so this was straightforward.

There ARE plenty of custom styles that should be considered/dealt with, but since they weren't in scope for this ticket, I left them to be handled later.


lfdebrux (approved):
🔥

PR #1106, opened 2020-11-10
Migrate to govuk frontend v3

domoscargin:
https://trello.com/c/bFRxUx3x/231-5-migrate-buyer-frontend-to-govuk-frontend-v3

This PR removes all old frontend toolkit and govuk_template styles and components and fully replaces them with govuk-frontend v3 components.

## Todo:

* Rebase on master when #1111 is merged
* Run a code audit to catch any outstanding custom styles that we don't need
* Test!


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

lfdebrux (approved):
Amazing stuff. This was really easy to review because of all the detail in the commit messages, thanks for that.

As far as I can tell everything is working just dandy, I think we should 🚢 it to preview 👍

lfdebrux (changes requested):
Spotted a problem, the error summary ids need to be updated for radios in tell us about contract journey.

Edited to add: I think we need to copy these lines from briefs-frontend to get the error summary links to work

https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/e4436f247bcf0bb73f12d810733489312a0c6026/config.py#L85-L88

lfdebrux:
Another issue I've found, the conditional reveal doesn't seem to work for save a new search page.

Screenshot 2020-11-23 at 17 30 14

domoscargin:
> https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/e4436f247bcf0bb73f12d810733489312a0c6026/config.py#L85-L88

Hmm, I don't think this will work since it's all WTForm-based rather than content loader. But Briefs has a couple of instances where the `get_errors_from_wtforms` method is working with error summary, so I'll investigate that.

lfdebrux:
> > https://github.com/alphagov/digitalmarketplace-briefs-frontend/blob/e4436f247bcf0bb73f12d810733489312a0c6026/config.py#L85-L88
> 
> Hmm, I don't think this will work since it's all WTForm-based rather than content loader. But Briefs has a couple of instances where the `get_errors_from_wtforms` method is working with error summary, so I'll investigate that.

Ah right; that's why I couldn't get it working 😁 

In that case I think we just hardcode it.

app/main/forms/direct_award_forms.py
```diff
@@ -16,6 +16,7 @@
 
 class CreateProjectForm(FlaskForm):
     save_search_selection = DMRadioField(
+        id="input-save_search_selection",
```
domoscargin, line 19: I would've flailed a bit before finding this solution! 👏

domoscargin:
Thanks for these, @lfdebrux 

All looks good for me and passes locally.

lfdebrux (approved):

PR #1107, opened 2020-11-10
Get rid of unnecessary styles on service pages

domoscargin:
https://trello.com/c/lu1zDbsy/267-2-are-we-ready-to-migrate-buyer-frontend-to-govuk-frontend-v3

Trying to split out as much work as can be split out from the v3 migration branch (https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/1106).

We can replace all the classes in `_service_page.scss` with Design System classes, and this does that.


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/templates/service.html
```diff
@@ -86,7 +86,7 @@ <h2 class="govuk-heading-m app-summary-list-heading" id="service-documents">Serv
       </li>
     {% endfor %}
     </ul>
-    <a class="govuk-link" href="#main-content" class="return-to-top">Return to top ↑</a>
+    <a class="govuk-link govuk-body" href="#main-content">Return to top ↑</a>
```
lfdebrux, line 89: I wonder if we can get rid of this link altogether...

app/assets/scss/application.scss
```diff
@@ -64,7 +64,6 @@ $govuk-global-styles: true;
 @import "_index-page.scss";
 @import "_search-page.scss";
 @import "_search-summary.scss";
-@import "_service-page.scss";
```
lfdebrux, line 67: 🙌

app/templates/service.html
domoscargin: Considering we don't seem to use this anywhere else, and there's room for confusion with back to top links, I don't necessarily disagree...

lfdebrux (approved):
Great stuff 😁

PR #1108, opened 2020-11-10
Remove content-table styles

domoscargin:
https://trello.com/c/lu1zDbsy/267-2-are-we-ready-to-migrate-buyer-frontend-to-govuk-frontend-v3

Trying to split out as much work as can be split out from the v3 migration branch (#1106).

We use `.content-table` styles in one place: on the cookie page. And here they can be easily replaced by govukTables.


lfdebrux (approved):
🥳

PR #1113, opened 2020-11-13
Replace govukTable with govukSummaryList

domoscargin:
On mobile/small screens, table text gets pretty squashed. This solves that since summary lists are stacked.

However, the summary lists are also not particularly visually distinct when multiple are stacked.

It might be worth considering revisiting this page when the summary_card component is fully adopted: https://github.com/alphagov/govuk-design-system-backlog/issues/210

Screenshot 2020-11-13 at 16 10 31

Screenshot 2020-11-19 at 09 47 46

### DIFF
--- a/app/templates/content/cookies.html
+++ b/app/templates/content/cookies.html
@@ -40,19 +40,21 @@ Cookies - Digital Marketplace
     <p class="govuk-body">
       We store a session cookie on your device to help keep your information secure while you use the service.
     </p>
-
-    {{ govukTable({
-      "head": [
-        {"text": "Name", "classes": "govuk-body-s"},
-        {"text": "Purpose", "classes": "govuk-body-s"},
-        {"text": "Expires", "classes": "govuk-body-s"}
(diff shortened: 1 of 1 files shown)


lfdebrux:
Huh, I didn't know you could get the summary list to be laid out horizontally like that.

Does making the cookie name bold help with the visual separation at all?

domoscargin:
> Huh, I didn't know you could get the summary list to be laid out horizontally like that.
> 
> Does making the cookie name bold help with the visual separation at all?

Sorry, the first image is the govukTable on small screens, the second image is the govukSummaryList.

lfdebrux:
What do you think, better than tables or no?

domoscargin:
I've tweaked the spacing and bolded the cookie name.

I think the table experience on small screens is pretty rubbish, so I do think this is better. And should be easy to extend when the new spec comes in.

lfdebrux (approved):

PR #1116, opened 2020-11-27
Add override classes for briefs stat block font

domoscargin:
@fajerq caught this font mismatch on Brief pages. Rather than introducing css to cover it, I've used some override styles, because the hope is we can remove all custom styles for this component some time in the future.

### Before
Screenshot 2020-11-27 at 15 37 23

### After
Screenshot 2020-11-27 at 15 37 38


gidsg (approved):
👍 🖌️

PR #1120, opened 2020-12-02
Fix details list on briefs search page

domoscargin:
This list should have some classes on it to make it conform to how it appeared before.

Screenshot 2020-12-02 at 15 25 39


lfdebrux:
LGTM 👍

PR #1121, opened 2020-12-03
Add accessible format request text to service page

domoscargin:
https://trello.com/c/tUd7Ck2w/275-5-replace-document-links-with-attachment-component-in-frontend-apps

Adds details on how to request accessible versions of the documents on the service page.

## Before

Screenshot 2020-12-03 at 12 17 25

Screenshot 2020-12-03 at 12 17 37

## After
Screenshot 2020-12-03 at 12 17 45

Screenshot 2020-12-03 at 12 13 11

### DIFF
--- a/app/templates/service.html
+++ b/app/templates/service.html
@@ -76,6 +76,10 @@
 <div class="govuk-grid-row" id="service-attributes">
   <div class="govuk-grid-column-full">
     {% include '_service_attributes.html' %}
+  </div>
+</div>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
     <h2 class="govuk-heading-m app-summary-list-heading" id="service-documents">Service documents</h2>
     <ul class="govuk-list">
     {% for document in service.meta.documents %}
@@ -93,6 +97,11 @@
       </li>
     {% endfor %}
     </ul>
+    {{ govukDetails({
+      "classes": "govuk-!-font-size-16",
+      "summaryText": "Request an accessible format",
+      "html": 'If you use assistive technology (such as a screen reader) and need a version of these documents in a more accessible format, please email <a href="mailto:info@crowncommercial.gov.uk" target="_blank" class="govuk-link">info@crowncommercial.gov.uk</a>. Please tell us what format you need. It will help us if you say what assistive technology you use.'
+    })}}
   </div>
 </div>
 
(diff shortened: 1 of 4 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally

app/templates/service.html
```diff
@@ -93,6 +97,11 @@ <h2 class="govuk-heading-m app-summary-list-heading" id="service-documents">Serv
       </li>
     {% endfor %}
     </ul>
+    {{ govukDetails({
+      "classes": "govuk-!-font-size-16",
+      "summaryText": "Request an accessible format",
+      "html": 'If you use assistive technology (such as a screen reader) and need a version of these documents in a more accessible format, please email <a href="mailto:info@crowncommercial.gov.uk" target="_blank" class="govuk-link">info@crowncommercial.gov.uk</a>. Please tell us what format you need. It will help us if you say what assistive technology you use.'
```
lfdebrux, line 103: This line is missing the `span` from the other details component.

```suggestion
      "html": 'If you use assistive technology (such as a screen reader) and need a version of these documents in a more accessible format, please email <span class="govuk-!-display-block govuk-!-margin-top-1 govuk-!-margin-bottom-2"><a href="mailto:info@crowncommercial.gov.uk" target="_blank" class="govuk-link">info@crowncommercial.gov.uk</a>.</span> Please tell us what format you need. It will help us if you say what assistive technology you use.'
```

Is it possible to put the details component (or even the whole list of documents) in a template partial?

domoscargin: This was on purpose. The span'd version is for the small meta column, to make the email stand out without weird line breaks:
Screenshot 2020-12-08 at 12 28 32

The non-spanned version is for the more generous space at the bottom of the page:
Screenshot 2020-12-08 at 12 28 25

domoscargin: I could rejig them to look the same and move this out into a separate file though.

domoscargin: Done

lfdebrux: Ah, sorry, I didn't realise it was on purpose.

lfdebrux (approved):

app/templates/service.html
lfdebrux: I think both versions look good when testing them with a mobile form factor or when zoomed in to 400%, so I'd say it's up to you to go with which version you prefer.

PR #1134, opened 2021-02-19
Fix error summary link for choose a category page

lfdebrux:
govuk-frontend 3 does not have `-1` appended to the id for the first input element in a radios component.


bjgill (approved):
Nice. I assume we haven't made this mistake anywhere else?

lfdebrux:
> Nice. I assume we haven't made this mistake anywhere else?

I'm hoping not, because this form is a special case with it's hand-written error handling, but 🤷

PR #1150, opened 2021-03-08
Show 404s for deleted services

alex9smith:
In order to support a supplier's request to be removed completely from G-Cloud we are adding a new service status - `deleted`.

The URL for a service with `deleted` status should return a 404. This is different from the behaviour of `disabled` status, which still shows the service information at that URL but with a banner stating the service is unavailable.

https://trello.com/c/qqL3sqWV/2182-remove-supplier-92526-from-g-cloud

### DIFF
--- a/tests/main/views/test_services.py
+++ b/tests/main/views/test_services.py
@@ -321,3 +321,11 @@ class TestServicePage(DataAPIClientMixin, BaseApplicationTest):
             heading.strip() for heading in attribute_headings]
 
         assert 'Certifications' not in attribute_headings
+
+    def test_deleted_service_causes_404(self):
+        self.service["services"]["status"] = "deleted"
+        self.data_api_client.get_service.return_value = self.service
+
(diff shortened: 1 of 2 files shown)


tests/main/views/test_services.py
```diff
@@ -321,3 +321,11 @@ def test_certifications_section_not_displayed_if_service_has_none(self):
             heading.strip() for heading in attribute_headings]
 
         assert 'Certifications' not in attribute_headings
+
+    def test_deleted_service_causes_404(self):
+        self.service["services"]["status"] = "deleted"
+        self.data_api_client.get_service.return_value = self.service
+
+        service_id = self.service['services']['id']
+        res = self.client.get('/g-cloud/services/{}'.format(service_id))
```
gidsg, line 330: could use an `f` string instead?

gidsg (approved):
LGTM 👍 🗡️

PR #1154, opened 2021-03-18
Add message on how to request alternative formats of service documents

lfdebrux:
We want users to know who to send requests for accessible versions of service documents to, this commit adds a piece of content with guidance after all the documents. Users should direct their requests to suppliers, and suppliers should always have a public contact email address, so we use that.

We could have had this as part of the alternativeFormatHtml for each attachment component, but we thought that would have been too repetitive.


domoscargin (approved):
LGTM!

PR #1172, opened 2021-04-14
Simplify supplier prefix handling

bjgill:
https://trello.com/c/wW4w1XQZ/2198-p3-high-volume-scraping-every-two-hours

A few small things I found whilst looking at the code that displays the pages that are being repeatedly scraped.

### DIFF
--- a/app/main/views/suppliers.py
+++ b/app/main/views/suppliers.py
@@ -14,12 +14,9 @@ from ..helpers.shared_helpers import parse_link
 from ..helpers.framework_helpers import get_framework_description
 
 
-def process_prefix(prefix=None, format='view'):
+def process_prefix(prefix=None):
     if prefix == u"other":  # special case
-        if format == 'api':
-            return u"other"
-        else:
-            return prefix
+        return prefix
     if is_alpha(prefix):
         return prefix[:1].upper()
     return u"A"  # default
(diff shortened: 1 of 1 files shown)


app/main/views/suppliers.py
```diff
@@ -71,9 +63,9 @@ def suppliers_details(supplier_id):
     if any(supplier.get('service_counts', {}).get(framework_name, 0) > 0 for framework_name in live_framework_names):
         first_character_of_supplier_name = supplier["name"][:1]
         if is_alpha(first_character_of_supplier_name):
-            prefix = process_prefix(prefix=first_character_of_supplier_name, format='template')
+            prefix = first_character_of_supplier_name.upper()
         else:
-            prefix = u"other"
+            prefix = "other"
```
gidsg, line 68: Is this unnecessary use of formatter something a linter could detect?

gidsg (approved):
🥨 👍

app/main/views/suppliers.py
bjgill: A linter could detect this (e.g. https://github.com/5j9/flake8-no-u-prefixed-strings). However, I think the risk of people introducing new instances of this is very low, so it might not be worth it.

